PR #1, opened 2015-01-13
[JENKINS-26380] Potential deadlock in durable task when OnceRetentionStrategy

christ66:
Add ReentrantLock to the OnceRetentionStrategy. If taking a slave offline takes longer than the idleMinutes, or occurs at around the same time, then the plugin will deadlock.


stephenc:
Nope this is not the way to fix it. The removal part is wrong. It needs to be two step, both wrapped in code that holds the queue lock

PR #2, opened 2015-01-13
[JENKINS-26380] Terminate nodes correctly

stephenc:
[https://issues.jenkins-ci.org/browse/JENKINS-26380](JENKINS-26380)

Supersedes #1.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -24,12 +24,16 @@
 
 package org.jenkinsci.plugins.durabletask.executors;
 
+import hudson.model.Computer;
 import hudson.model.Executor;
 import hudson.model.ExecutorListener;
 import hudson.model.Queue;
 import hudson.slaves.AbstractCloudComputer;
+import hudson.slaves.AbstractCloudSlave;
 import hudson.slaves.CloudRetentionStrategy;
 import hudson.slaves.EphemeralNode;
+import jenkins.model.Jenkins;
+
 import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -76,15 +82,38 @@ private void done(Executor executor) {
...
-            c.getNode().terminate();
-        } catch (InterruptedException x) {
-            LOGGER.log(Level.WARNING, null, x);
-        } catch (IOException x) {
-            LOGGER.log(Level.WARNING, null, x);
+        synchronized (this) {
+            if (terminating) {
+                return;
+            }
+            terminating = true;
```
jglick: Seems like an `AtomicBoolean` would be more idiomatic, but perhaps you need the monitor on `OnceRetentionStrategy.this` anyway to be exclusive with the `threadPoolForRemoting` thread?

jglick:
Do not really follow the implications of the change, but seems OK to merge to me.

jglick:
Is this intended to be a final fix of JENKINS-26380, or do you expect other retention strategies are also in danger from the core locking change?

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
stephenc: correct we want the lock here to ensure exclusivity of starting the background task

stephenc:
@jglick other retention strategies are already broken if they didn't acquire the lock on Queue before checking idle and then in the background repeat after reacquiring the lock on Queue. The core change makes it easier to do that.

It does not make sense to add the lock to Jenkins.removeNode as there may be valid cases where you want to remove the node and jobs in progress be damned... hence it is up to the caller to ensure they have the correct lock based on whether the node being idle is important to them or not

PR #3, opened 2015-01-21
Check method should also contain a lock on the queue in the event it nee...

christ66:
...ds to terminate the slave machine.

(Continuation of #2 ) If the slave machine takes over 1 minute to terminate the check method will be executed which could encounter a deadlock. Instead we override the check method to hold the lock on the queue before terminating.
@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -32,6 +32,7 @@ import hudson.slaves.AbstractCloudComputer;
 import hudson.slaves.AbstractCloudSlave;
 import hudson.slaves.CloudRetentionStrategy;
 import hudson.slaves.EphemeralNode;
+import hudson.util.TimeUnit2;
 import jenkins.model.Jenkins;
 
 import java.io.IOException;
@@ -47,6 +48,8 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
     private static final Logger LOGGER = Logger.getLogger(OnceRetentionStrategy.class.getName());
 
     private transient boolean terminating;
+    
+    private int idleMinutes;
 
     /**
      * Creates the retention strategy.
@@ -54,14 +57,19 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
      */
     public OnceRetentionStrategy(int idleMinutes) {
         super(idleMinutes);
+        this.idleMinutes = idleMinutes;
     }
 
     @Override
     public long check(final AbstractCloudComputer c) {
         // When the slave is idle we should disable accepting tasks and check to see if it is already trying to
         // terminate. If it's not already trying to terminate then lets terminate manually.
-        if (c.isIdle()) {
-            done(c);
+        if (c.isIdle() && !disabled) {
+            final long idleMilliseconds = System.currentTimeMillis() - c.getIdleStartMilliseconds();
+            if (idleMilliseconds > TimeUnit2.MINUTES.toMillis(idleMinutes)) {
+                LOGGER.log(Level.FINE, "Disconnecting {0}", c.getName());
+                done(c);
+            }
         }
 
         // Return one because we want to check every minute if idle.
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -55,6 +55,53 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
     public OnceRetentionStrategy(int idleMinutes) {
         super(idleMinutes);
     }
+    
+    @Override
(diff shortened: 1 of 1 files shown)


stephenc:
:+1: from me... though as some of the changes are mine it would be better if you got someone else to take a peek before merging

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -56,6 +56,18 @@ public OnceRetentionStrategy(int idleMinutes) {
         super(idleMinutes);
     }
 
+    @Override
+    public long check(final AbstractCloudComputer c) {
+        // When the slave is idle we should disable accepting tasks and check to see if it is already trying to
+        // terminate. If it's not already trying to terminate then lets terminate manually.
+        if (c.isIdle()) {
+            done(c);
```
jglick: This is bypassing the `idleMinutes` condition previously done in the superclass, so whatever its other benefits, it cannot be correct. I suspect what you meant this method to be was something like

```java
        if (c.isIdle() && !disabled) {
            final long idleMilliseconds = System.currentTimeMillis() - c.getIdleStartMilliseconds();
            if (idleMilliseconds > MINUTES.toMillis(idleMinutes)) {
                LOGGER.log(LEVEL.FINE, "Disconnecting {0}", c.getName());
                done(c);
            }
        }
        return 1;
```

with a workaround for the fact that `CloudRetentionStrategy.idleMinutes` is `private`, either by shadowing the field here, or just extending `RetentionStrategy<AbstractCloudComputer>` directly (and inlining `super.start`).

stephenc: We need to bypass the super method as that does not lock correctly when terminating (at least in older versions of Jenkins)

In any case a Once retention strategy should probably ignore the idle minutes once it has been used... if it has not yet been used then it should consider the idle minutes

jglick:
I understand that this is fixing a problem which is probably difficult to reproduce and verify, but to what extent has this change been tested for possible regressions? An example would be a workflow job using the Mock Slave plugin (which depends on this strategy) to do something like (off the top of my head here!)

```groovy
def branches = [:]
def r = new Random()
for (int i = 0; i < 10; i++) {
  branches["b${i}"] = {
    for (int j = 0; j < r.nextInt(10); j++) {
      node('mock-cloud') {
        sh "sleep ${r.nextInt(300)}"
      }
    }
  }
}
parallel branches
```

and then verifying that no slave gets killed before its time, all slaves eventually are removed, etc.

(I never got to the stage of writing meaningful functional tests for this strategy, and probably it would require a scalability test harness to catch real problems.)

jglick:
The Docker plugin can also be used as an interactive testbed.

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
jglick: > a Once retention strategy should probably ignore the idle minutes once it has been used... if it has not yet been used then it should consider the idle minutes

Not sure I follow what “been used” means here. `done` called, or a task run on the slave? The purpose of having `idleMinutes` at all in this strategy is as a second line of defense in case the `ExecutorListener` event is never received for some reason: the slave will still be taken offline without human intervention, albeit tardily.

jglick:
Any status on this?

stephenc:
From my perspective this is good to go, not sure what Steven X thinks or
how he plans to address your comments on the idle delay

On Wednesday, January 28, 2015, Jesse Glick <notifications@github.com>
wrote:

> Any status on this?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/durable-task-plugin/pull/3#issuecomment-71758622>
> .
>


-- 
Sent from my phone

jglick:
So @christ66 are you planning to address the `idleMinutes` regression, or do I need to take this over? It sounds like this fix is a priority for some users.

christ66:
@jglick sorry about the delay. I fixed the idle minutes check.

stephenc:
+1 from me

On Saturday, January 31, 2015, Steven Christou <notifications@github.com>
wrote:

> @jglick <https://github.com/jglick> sorry about the delay. I fixed the
> idle minutes check.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jenkinsci/durable-task-plugin/pull/3#issuecomment-72299577>
> .
>


-- 
Sent from my phone

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -54,6 +57,23 @@
      */
     public OnceRetentionStrategy(int idleMinutes) {
         super(idleMinutes);
```
jglick: This patch basically makes it pointless to extend `CloudRetentionStrategy` at all, but that is a subject for a future cleanup refactoring. :+1:

jglick:
Whatever this was fixing, it does not seem to have fixed [JENKINS-27476](https://issues.jenkins-ci.org/browse/JENKINS-27476).

PR #5, opened 2015-04-09
Better support for stop

jglick:
- [X] handle `JENKINS_SERVER_COOKIE` better
- [X] accept a decorated `Launcher`
- [x] make `ProcessLiveness` more robust and/or pluggable

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
@@ -53,9 +53,21 @@ public abstract class Controller implements Serializable {
     /**
      * Checks whether the task has finished.
      * @param workspace the workspace in use
+     * @param launcher a way to start processes
      * @return an exit code (zero is successful), or null if the task appears to still be running
      */
-    public abstract @CheckForNull Integer exitStatus(FilePath workspace) throws IOException, InterruptedException;
+    public @CheckForNull Integer exitStatus(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
```diff
@@ -48,15 +53,43 @@
...
      * @param workspace the workspace in use
+     * @param launcher a way to start processes
      */
-    public abstract void stop(FilePath workspace) throws IOException, InterruptedException;
+    public void stop(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
+        if (Util.isOverridden(Controller.class, getClass(), "stop", FilePath.class)) {
+            stop(workspace);
+        } else {
+            throw new AbstractMethodError("implement stop(FilePath, Launcher)");
+        }
```
tfennelly: I get the light sweats when I see this `Util.isOverridden` funkery ;)

stephenc: sounds like you just need to culchie-up<sup>§</sup> and drink a stronger cuppa tae

<hr/>
<sup>§</sup> a gender neutral form of man-up

tfennelly:
:+1:

src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
jglick: Yeah makes me nervous too, especially since we lack a proper test framework for verifying binary compatibility.

PR #6, opened 2015-07-02
Provide UI configuration for strategy

KostyaSha:
There are different ways of using strategies with UI configuration:
- do wrappers in all plugins that using it, 
- subclass (private fields is pain) 
- just provide normal Describable

This PR:
- Provides UI configuration for strategy
- Suggest good idle timeout (example references in comments)

Every slave instance should have copy of this strategy because of 'idle' field, should this class be `Cloneable` to get configured copies?

[Review on Reviewable](https://reviewable.io/reviews/jenkinsci/durable-task-plugin/6)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -25,6 +25,7 @@
 package org.jenkinsci.plugins.durabletask.executors;
 
 import hudson.model.Computer;
+import hudson.model.Descriptor;
 import hudson.model.Executor;
 import hudson.model.ExecutorListener;
 import hudson.model.Queue;
@@ -32,8 +33,12 @@ import hudson.slaves.AbstractCloudComputer;
 import hudson.slaves.AbstractCloudSlave;
 import hudson.slaves.CloudRetentionStrategy;
 import hudson.slaves.EphemeralNode;
+import hudson.slaves.RetentionStrategy;
 import hudson.util.TimeUnit2;
 import jenkins.model.Jenkins;
+import org.kohsuke.accmod.Restricted;
+import org.kohsuke.accmod.restrictions.NoExternalUse;
+import org.kohsuke.stapler.DataBoundConstructor;
 
 import java.io.IOException;
 import java.util.logging.Level;
@@ -49,12 +54,13 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
 
     private transient boolean terminating;
     
-    private int idleMinutes;
+    protected int idleMinutes;
 
     /**
      * Creates the retention strategy.
      * @param idleMinutes number of minutes of idleness after which to kill the slave; serves a backup in case the strategy fails to detect the end of a task
      */
+    @DataBoundConstructor
     public OnceRetentionStrategy(int idleMinutes) {
         super(idleMinutes);
         this.idleMinutes = idleMinutes;
@@ -140,4 +146,18 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
         });
     }
 
+    @Override
+    public DescriptorImpl getDescriptor() {
+        return DESCRIPTOR;
+    }
+
+    @Restricted(NoExternalUse.class)
+    public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
+
+    public static final class DescriptorImpl extends Descriptor<RetentionStrategy<?>> {
+        @Override
+        public String getDisplayName() {
+            return "Once Retention Strategy";
+        }
+    }
 }
--- a/src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/config.groovy
+++ b/src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/config.groovy
@@ -1,0 +1,7 @@
+package org.jenkinsci.plugins.durabletask.executors.OnceRetentionStrategy
+
+def f = namespace(lib.FormTagLib);
+
+f.entry(title: "Idle timeout", field: "idleMinutes") {
+    f.number(default: 0)
+}
--- a/src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/help-idleMinutes.html
+++ b/src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/help-idleMinutes.html
@@ -1,0 +1,4 @@
+<div>
+    Number of minutes of idleness after which to kill the slave;
+    serves a backup in case the strategy fails to detect the end of a task
+</div>


KostyaSha:
Hm, but without descriptor i still will be unable to use it...

KostyaSha:
Updated PR

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -140,4 +144,18 @@ public void run() {
...
 
+    @Override
+    public DescriptorImpl getDescriptor() {
+        return DESCRIPTOR;
+    }
+
+    @Restricted(NoExternalUse.class)
+    public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
+
+    public static final class DescriptorImpl extends Descriptor<RetentionStrategy<?>> {
```
oleg-nenashev: There should be an <code>@Extension</code> somewhere

KostyaSha: https://github.com/jenkinsci/docker-plugin/pull/247

oleg-nenashev: It should be here. Otherwise this configuration feature is not complete. Any plugin will require additional stuff

KostyaSha: https://github.com/jenkinsci/docker-plugin/pull/247#issuecomment-111629010

oleg-nenashev: In such case I'm OK with it

oleg-nenashev:
:+1:

src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/config.groovy
```diff
@@ -0,0 +1,7 @@
+package org.jenkinsci.plugins.durabletask.executors.OnceRetentionStrategy
+
+def f = namespace(lib.FormTagLib);
+
+f.entry(title: "Idle timeout", field: "idleMinutes") {
+    f.number(default: 0)
```
KostyaSha: very bad default...

oleg-nenashev: AFAIK it was a default behavior, but I agree

KostyaSha: It caused many issues in docker plugin, as soon as strategy runs it kills slave. I reverted to 10 minutes according to users feedback.

FYI @stephenc

stephenc: Yeah I don't use this strategy in CJOC. I suspect it needs fine-tuning to be able to run with `0` as a default

KostyaSha: So CJOC doesn't support durable-tasks? ;)

stephenc: It does, but we use our own retention strategy rather than this one

stephenc: (and we use our own because we were using it since 2011... if it ain't broke don't fix it...)

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -49,12 +54,13 @@
 
     private transient boolean terminating;
     
-    private int idleMinutes;
+    protected int idleMinutes;
```
jglick: Why? There should rather be a `public` getter.

src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/config.groovy
jglick: Either `0` or `1` seems fine as a default.

> as soon as strategy runs it kills slave

This should only happen if there is no task running on it, obviously, so if you see differently then something else is wrong somewhere, but arbitrarily adjusting the idle timeout is not the fix. Any nonnegative value is legitimate here.

KostyaSha: @jglick i understood your point, i will think about removing durable-task support and not using this retention strategy in docker-plugin until jenkins and this plugin will provide normal API

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -140,4 +146,18 @@ public void run() {
...
+        return DESCRIPTOR;
+    }
+
+    @Restricted(NoExternalUse.class)
+    public static final DescriptorImpl DESCRIPTOR = new DescriptorImpl();
+
+    public static final class DescriptorImpl extends Descriptor<RetentionStrategy<?>> {
+        @Override
+        public String getDisplayName() {
+            return "Once Retention Strategy";
```
jglick: Probably meaningless to users. Might suggest _Use this slave for a single build only_.

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
KostyaSha: because i don't need getter and i need inherit from this class

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
jglick: BTW `DescriptorImpl` can probably be `private` in such a case (or even an anonymous inner class).

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
KostyaSha: either it will store 3d idleMinutes xml entry because every class hides it

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
KostyaSha: until there is no normal API i handle it externally in executors setting.

src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/config.groovy
KostyaSha: https://www.youtube.com/watch?v=aq_grSF64tM&feature=youtu.be&t=1219
good luck with 0 and tones of unimplement synchronisation locks.

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
jglick: I guess I am not understanding your use case. This class was not designed for subclassing.

My understanding is that you want to provide a user option in `docker`’s `Cloud` to select one of several retention strategies known to be meaningful for this cloud, right? So you could either

* Provide a fully custom configuration UI for that, in which case this PR is unnecessary because you would internally call the `OnceRetentionStrategy` constructor directly from `docker` plugin code when creating a cloud slave. `mock-slave` does this, for example.
* Or, provide a customized `List<Descriptor<RetentionStrategy<?>>>` to a `f:heteroProperty` or the like, in which case the standard `config` view of each strategy is displayed, and you would follow standard databinding idioms including a `public` getter (and `Descriptor.getDisplayName` would be used).

You seem to trying some other approach but I do not understand what that is, and I cannot really evaluate the PR without that background. (There was no description in the PR stating the use case.)

KostyaSha: > This class was not designed for subclassing.

and this is mistake, 

>you would internally call the OnceRetentionStrategy

Why should i implement and wrap this class? 
1) Provide usable from UI class, that

> follow standard data binding idioms including a public getter (and Descriptor.getDisplayName would be used).

2) Make class re-usable, for migration purposes (or any other) i should subclass it. 

CloudRetentionStrategy do also bad thing by hiding idleTimeout that you in the same way duplicating in your strategy and that after de-serialisation duplicates xml entries.

> You seem to trying some other approach but I do not understand what that is, and I cannot really evaluate the PR without that background. (There was no description in the PR stating the use case.)

Provide standard Describable/Descriptor strategy if you want other plugins to use this strategy, that's all.

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
KostyaSha: And how i will add it then into custom DescriptorList?

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
KostyaSha: Btw, RetentionStrategy may provide some 'isApplicableExecutor(int num)` for such checks

KostyaSha: > Probably meaningless to users. Might suggest Use this slave for a single build only.

Can be renamed as you wish, maybe better put description in html.

src/main/resources/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy/config.groovy
KostyaSha: @jglick https://github.com/jenkinsci/kubernetes-plugin/pull/26 
Let's make 5 minutes? One minute maybe not enough for long locked queue.

KostyaSha:
And now it unmergeable :(

KostyaSha:
Btw, after introducing `terminating` this class should support `Cloneable` because only copy should be passed to new slaves. Then it will be possible to configure object from UI and use clones for slaves.

KostyaSha:
One year achievement unlocked. RetentionStrategies still not reusable/UI configurable in core and here. Plugins still must copy-paste code.

oleg-nenashev:
@jenkinsci/code-reviewers @jglick

jglick:
I still think this is the wrong approach.

car-roll:

PR #8, opened 2015-07-31
Do not kill a one-shot slave merely because a flyweight task happened to run on it

jglick:
Complements https://github.com/jenkinsci/branch-api-plugin/pull/5. Either suffice to fix the problem, but better to have both.

@reviewbybees


oleg-nenashev:
:bee:

PR #7, opened 2015-07-10
Fixed FindBugs issues

rsandell:
And configured FindBugs for xml output so Jenkins can pick it up.

@reviewbybees

PR #9, opened 2015-08-03
Reverting an incorrect change from #7

jglick:
If left uncorrected, I suspect this would prevent `bat` steps from being used in flows when the Jenkins master is on Unix.

@reviewbybees esp. @rsandell


tfennelly:
:bee:

PR #10, opened 2015-10-28
Updated to 1.609.x

jglick:
Allows for some code simplifications, and use of new queue locking (@stephenc). Anyway I cannot find any plugins depending on this API which are not already on 1.609.x.

@reviewbybees


stephenc:
:bee:

tfennelly:
:bee:

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -120,27 +119,27 @@ private void done(final AbstractCloudComputer<?> c) {
         Computer.threadPoolForRemoting.submit(new Runnable() {
             @Override
             public void run() {
-                final Jenkins jenkins = Jenkins.getInstance();
-                // TODO once the baseline is 1.592+ switch to Queue.withLock
```
jglick, line 124: #2

PR #11, opened 2015-11-16
[JENKINS-27152] Ameliorate problem by using a predictable control directory name

jglick:
[JENKINS-27152](https://issues.jenkins-ci.org/browse/JENKINS-27152)

@reviewbybees and @cruftex


oleg-nenashev:
+1 for something with "jenkins" and "workflow"

jglick:
Well the API is not actually limited to Workflow; CloudBees also makes a plugin using it. Could perhaps have a prefix sensitive to the `FileMonitoringTask` subclass.

jglick:
Anyway this would only be a short-term workaround. I still intend to _fix_ JENKINS-27152, it just requires some study to avoid breaking `docker-workflow`.

PR #13, opened 2016-01-05
[JENKINS-32264] Use getpgid() to determine if process exists

rodrigc:
This is more portable than using "/proc/<pid>" which is
Linux-specific.


jglick:
More comprehensive fix in #14. Suggested ways to help test it:

* `mvn clean test` on FreeBSD.
* `mvn -DskipTests clean package` (any platform), install in Jenkins, run a FreeBSD slave, verify that `sh` steps normally work but return -1 if the slave machine is abruptly rebooted (and the slave agent reconnected), or the controller process is otherwise killed.

PR #14, opened 2016-01-05
[JENKINS-32264] Fix process liveness check for non-Linux platforms

jglick:
[JENKINS-32264](https://issues.jenkins-ci.org/browse/JENKINS-32264)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
@@ -29,6 +29,7 @@ import hudson.os.PosixAPI;
 import hudson.os.PosixException;
 import hudson.remoting.VirtualChannel;
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
@@ -36,6 +37,9 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.security.MasterToSlaveCallable;
 import jnr.posix.POSIX;
+import jnr.posix.POSIXFactory;
+import jnr.posix.POSIXHandler;
+import jnr.posix.util.DefaultPOSIXHandler;
 
 /**
  * Utility class to track whether a given process is still alive.
@@ -72,7 +76,14 @@ final class ProcessLiveness {
 
     private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launcher) throws IOException, InterruptedException {
         if (launcher instanceof Launcher.LocalLauncher || launcher instanceof Launcher.RemoteLauncher) {
-            return channel.call(new Liveness(pid));
+            try {
+                return channel.call(new Liveness(pid));
+            } catch (IllegalStateException x) {
+                LOGGER.log(Level.WARNING, "cannot determine liveness of " + pid, x);
+            } catch (PosixException x) {
+                LOGGER.log(Level.WARNING, "cannot determine liveness of " + pid, x);
+            }
+            return true;
         } else {
             // Using a special launcher; let it decide how to do this.
             // TODO perhaps this should be a method in Launcher, with the following fallback in DecoratedLauncher:
@@ -81,23 +92,26 @@ final class ProcessLiveness {
     }
 
     private static final class Liveness extends MasterToSlaveCallable<Boolean,RuntimeException> {
+        private static final POSIX jnr = PosixAPI.jnr();
         private final int pid;
         Liveness(int pid) {
             this.pid = pid;
         }
         @Override public Boolean call() throws RuntimeException {
-            POSIX jnr = PosixAPI.jnr();
-            try {
-                // jnr.getpgid() (on Linux at least) always returns -1. JNR bug? libc would not support overloads.
-                if (jnr.getpgid(0) == -1) {
-                    LOGGER.log(Level.WARNING, "getpgid does not seem to work on this platform, using {0}", jnr);
-                    return true;
-                }
-                return jnr.getpgid(pid) != -1;
-            } catch (PosixException x) {
-                LOGGER.log(Level.WARNING, "failed to determine liveness of " + pid + " using " + jnr, x);
-                return true;
+            try { // loadPOSIX sends problems to stderr, useless if we are on a slave
+                Method loadNativePOSIX = POSIXFactory.class.getDeclaredMethod("loadNativePOSIX", POSIXHandler.class);
+                loadNativePOSIX.setAccessible(true);
+                loadNativePOSIX.invoke(null, new DefaultPOSIXHandler());
+            } catch (NoSuchMethodException x) {
+                LOGGER.log(Level.WARNING, "loadNativePOSIX no longer exists?", x);
+            } catch (Exception x) {
+                throw new IllegalStateException(x);
             }
+            // jnr.getpgid() (on Linux at least) always returns -1. JNR bug? libc would not support overloads.
+            if (jnr.getpgid(0) == -1) {
+                throw new IllegalStateException("getpgid does not seem to work on this platform");
+            }
+            return jnr.getpgid(pid) != -1;
         }
     }
 
--- a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
@@ -75,7 +75,7 @@ final class ProcessLiveness {
         if (launcher instanceof Launcher.LocalLauncher || launcher instanceof Launcher.RemoteLauncher) {
             try {
                 boolean alive = channel.call(new Liveness(pid));
-                LOGGER.log(Level.FINE, "{0} is alive? {1}", new Object[] {pid, alive});
+                LOGGER.log(Level.FINER, "{0} is alive? {1}", new Object[] {pid, alive});
                 return alive;
             } catch (RuntimeException x) {
                 LOGGER.log(Level.WARNING, "cannot determine liveness of " + pid, x);
--- a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
@@ -25,15 +25,26 @@
 package org.jenkinsci.plugins.durabletask;
 
 import hudson.Launcher;
+import hudson.os.PosixAPI;
+import hudson.os.PosixException;
 import hudson.remoting.VirtualChannel;
-import java.io.File;
 import java.io.IOException;
+import java.util.Collections;
+import java.util.Map;
+import java.util.WeakHashMap;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import jenkins.security.MasterToSlaveCallable;
+import jnr.posix.POSIX;
 
 /**
  * Utility class to track whether a given process is still alive.
  */
 final class ProcessLiveness {
+
+    private static final Logger LOGGER = Logger.getLogger(ProcessLiveness.class.getName());
+
+    private static final Map<Launcher,Boolean> workingLaunchers = Collections.synchronizedMap(new WeakHashMap<Launcher,Boolean>());
 
     /**
      * Determines whether a process is believed to still be alive.
@@ -43,6 +54,23 @@ final class ProcessLiveness {
      * @return true if it is apparently still alive (or we cannot tell); false if it is believed to not be running
      */
     public static boolean isAlive(VirtualChannel channel, int pid, Launcher launcher) throws IOException, InterruptedException {
+        Boolean working = workingLaunchers.get(launcher);
+        if (working == null) {
+            // Check to see if our logic correctly reports that an unlikely PID is not running.
+            working = !_isAlive(channel, 9999, launcher);
+            workingLaunchers.put(launcher, working);
+            if (!working) {
+                LOGGER.log(Level.WARNING, "{0} on {1} does not seem able to determine whether processes are alive or not", new Object[] {launcher, channel});
+                // TODO Channel.toString should report slave name, but would be nice to also report OS
+            }
+        }
+        if (!working) {
+            return true;
+        }
+        return _isAlive(channel, pid, launcher);
+    }
+
+    private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launcher) throws IOException, InterruptedException {
         if (launcher instanceof Launcher.LocalLauncher || launcher instanceof Launcher.RemoteLauncher) {
             return channel.call(new Liveness(pid));
         } else {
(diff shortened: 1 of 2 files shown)


rodrigc:
Please change "unusual platforms" to POSIX platforms.  That comment is a bit offensive.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -43,6 +54,23 @@
      * @return true if it is apparently still alive (or we cannot tell); false if it is believed to not be running
      */
     public static boolean isAlive(VirtualChannel channel, int pid, Launcher launcher) throws IOException, InterruptedException {
+        Boolean working = workingLaunchers.get(launcher);
+        if (working == null) {
+            // Check to see if our logic correctly reports that an unlikely PID is not running.
+            working = !_isAlive(channel, 9999, launcher);
```
rodrigc: Why did you choose 9999?  That's a valid pid.

jglick:
Did not mean to offend, but Jenkins users running on anything other than Linux, Windows, or (mainly for slaves) Mac OS X are clearly a minority.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
jglick: Yes it is, but it _unlikely_ to be a running process. If it is, when the first process is run on this slave, then `ProcessLiveness` will simply be disabled—which only matters in case the controlling `sh` process is killed for some weird reason, or the computer is rebooted. This reasoning is explained in JIRA.

I could have chosen an even less-likely integer but I wanted to choose a _potentially_ valid PID. Linux normally caps PIDs at 32k, and I do not wish to experiment with the effect of calling `getpgid` on an argument outside the bounds of what the kernel considers valid PIDs. (From what I can tell it correctly returns -1 on Linux.) I am not sure if there is a portable way to tell what `pid_t` is.

rodrigc: The only thing that POSIX guarantees is that pid_t is a signed integral type.  The most portable way to find the size of pid_t, would be to write some C code to get the sizeof pid_t, such as one of the solutions mentioned here: http://stackoverflow.com/questions/12841488/how-can-i-determine-the-maximum-value-of-a-pid-t

Linux has /proc/sys/kernel/pid_max , which works, but is less portable.

JNR has a pid_t interface ( https://github.com/jnr/jnr-ffi/blob/master/src/main/java/jnr/ffi/types/pid_t.java ), but I am not sure if it is possible to get the pid_t size from that.

9999 is not that big, and these days it is possible to hit that.  Although as you correctly mention that the probability of hitting this number is low, it can definitely happen.  I don't like to see these kinds of things, because these kind of booby traps tend to manifest itself when scaling systems like these, and lead to WTF moments.

rodrigc:
I tried this patch on both FreeBSD and MacOS X, and keep seeing this in the log:

```
WARNING: failure in task not wrapped in SafeTimerTask
java.lang.IllegalStateException: getpgid is not implemented in jnr-posix
        at jnr.posix.util.DefaultPOSIXHandler.unimplementedError(DefaultPOSIXHandler.java:28)
        at jnr.posix.JavaPOSIX.unimplementedInt(JavaPOSIX.java:401)
        at jnr.posix.JavaPOSIX.getpgid(JavaPOSIX.java:314)
        at jnr.posix.LazyPOSIX.getpgid(LazyPOSIX.java:147)
        at org.jenkinsci.plugins.durabletask.ProcessLiveness$Liveness.call(ProcessLiveness.java:92)
        at org.jenkinsci.plugins.durabletask.ProcessLiveness$Liveness.call(ProcessLiveness.java:83)
        at hudson.remoting.LocalChannel.call(LocalChannel.java:45)
        at org.jenkinsci.plugins.durabletask.ProcessLiveness._isAlive(ProcessLiveness.java:75)
        at org.jenkinsci.plugins.durabletask.ProcessLiveness.isAlive(ProcessLiveness.java:60)
        at org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController.exitStatus(BourneShellScript.java:139)
        at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.check(DurableTaskStep.java:187)
        at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$Execution.run(DurableTaskStep.java:150)
        at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
        at java.util.concurrent.FutureTask.run(FutureTask.java:266)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.access$201(ScheduledThreadPoolExecutor.java:180)
        at java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:293)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
        at java.lang.Thread.run(Thread.java:745)
```

rodrigc:
I modified the code in this PR slightly, and then it worked:

```
diff --git a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
index dfdfa20..a4eb250 100644
--- a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
@@ -35,7 +35,6 @@ import java.util.WeakHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.security.MasterToSlaveCallable;
-import jnr.posix.POSIX;
 
 /**
  * Utility class to track whether a given process is still alive.
@@ -86,16 +85,15 @@ final class ProcessLiveness {
             this.pid = pid;
         }
         @Override public Boolean call() throws RuntimeException {
-            POSIX jnr = PosixAPI.jnr();
             try {
                 // jnr.getpgid() (on Linux at least) always returns -1. JNR bug? libc would not support overloads.
-                if (jnr.getpgid(0) == -1) {
-                    LOGGER.log(Level.WARNING, "getpgid does not seem to work on this platform, using {0}", jnr);
+                if (PosixAPI.jnr().getpgid(0) == -1) {
+                    LOGGER.log(Level.WARNING, "getpgid does not seem to work on this platform, using {0}", PosixAPI.jnr());
                     return true;
                 }
-                return jnr.getpgid(pid) != -1;
+                return hudson.os.PosixAPI.jnr().getpgid(pid) != -1;
             } catch (PosixException x) {
-                LOGGER.log(Level.WARNING, "failed to determine liveness of " + pid + " using " + jnr, x);
+                LOGGER.log(Level.WARNING, "failed to determine liveness of " + pid + " using " + PosixAPI.jnr(), x);
                 return true;
             }
         }
```

jglick:
Makes no sense, will have to dig deeper.

BTW did you ever ascertain why the current plugin failed?

Guess I need to install FreeBSD in VirtualBox to proceed.

rodrigc:
For the original unmodified plugin, I did not isolate the cause of the failures on FreeBSD.
With this patch, I don't understand why I got those exceptions, but I got them on both FreeBSD and MacOS X.

rodrigc:
If you manage to set up a VirtualBox VM with FreeBSD, you may wish to use vagrant.
You can do something like:

```
vagrant init freebsd/FreeBSD-11.0-CURRENT
vagrant up
vagrant ssh
```

Inside the VM, you can quickly bootstrap by installing some packages:

```
pkg install devel/jenkins
pkg install devel/git
pkg install devel/maven3
```

and make sure you have at least this in /etc/rc.conf:

```
ifconfig_em0="DHCP"
sshd_enable="YES"
jenkins_enable="YES"
```
After rebooting the VM, jenkins should be available at http://<IP of VM>/:8180/jenkins

jglick:
Just saw your comments, after spending time Googling for the magic incantation to get DHCP to work on a new VBox installation. :-/ Anyway the above is not quite what I want since I am interesting in a FreeBSD _slave_, with an arbitrary master.

I am confused about why the original `/proc` trick did not just work. `pkg install openjdk8` asked me to mount `procfs` on `/proc`, which I did. Do you not have such a mount?

jglick:
…and on this machine, Durable Task 1.7 seems to work fine. So I guess I am out of luck w.r.t. diagnosing the cause of your original issue, unless you specifically declined to mount `/proc` (in which case I still do not know how it would happen).

jglick:
Never mind, I was actually running with the patch, will retry.

jglick:
Still, it works for me with 1.5-alpha-1.

jglick:
At any rate, I can reproduce the ISE so I will try to track this down.

jglick:
I suspect your patch did not really make it work, it just changed how exceptions were handled so that the process was unconditionally assumed alive.

jglick:
With some richer debugging (hardcoded to FreeBSD) I get

```
java.lang.NoClassDefFoundError: Could not initialize class jnr.posix.POSIXFactory$DefaultLibCProvider$SingletonHolder
	at jnr.posix.POSIXFactory$DefaultLibCProvider.getLibC(POSIXFactory.java:223)
	at jnr.posix.BaseNativePOSIX.<init>(BaseNativePOSIX.java:37)
	at jnr.posix.FreeBSDPOSIX.<init>(FreeBSDPOSIX.java:40)
	at jnr.posix.POSIXFactory.loadFreeBSDPOSIX(POSIXFactory.java:107)
```

Unfortunately [it does not seem that I can make this richer logging automatic](https://github.com/jnr/jnr-posix/blob/master/src/main/java/jnr/posix/POSIXFactory.java#L92): any messages are printed in the slave JVM and apparently lost.

jglick:
Unfortunately my pet [JDK-8051847](https://bugs.openjdk.java.net/browse/JDK-8051847) makes it hard to track down the root cause here.

jglick:
Using the script console on the slave _before_ running any build, I can track it back to

```
java.lang.UnsatisfiedLinkError: /usr/lib/libc.so: invalid file format
	at jnr.ffi.provider.jffi.NativeLibrary.loadNativeLibraries(NativeLibrary.java:87)
	at jnr.ffi.provider.jffi.NativeLibrary.getNativeLibraries(NativeLibrary.java:70)
	at jnr.ffi.provider.jffi.NativeLibrary.getSymbolAddress(NativeLibrary.java:49)
	at jnr.ffi.provider.jffi.NativeLibrary.findSymbolAddress(NativeLibrary.java:59)
	at jnr.ffi.provider.jffi.AsmLibraryLoader.generateInterfaceImpl(AsmLibraryLoader.java:117)
	at jnr.ffi.provider.jffi.AsmLibraryLoader.loadLibrary(AsmLibraryLoader.java:59)
	at jnr.ffi.provider.jffi.NativeLibraryLoader.loadLibrary(NativeLibraryLoader.java:43)
	at jnr.ffi.LibraryLoader.load(LibraryLoader.java:228)
	at jnr.ffi.Library.loadLibrary(Library.java:123)
	at jnr.posix.POSIXFactory$DefaultLibCProvider$SingletonHolder.<clinit>(POSIXFactory.java:219)
	at jnr.posix.POSIXFactory$DefaultLibCProvider.getLibC(POSIXFactory.java:223)
	at jnr.posix.BaseNativePOSIX.<init>(BaseNativePOSIX.java:37)
	at jnr.posix.FreeBSDPOSIX.<init>(FreeBSDPOSIX.java:40)
	at jnr.posix.POSIXFactory.loadFreeBSDPOSIX(POSIXFactory.java:107)
	at jnr.posix.POSIXFactory.loadNativePOSIX(POSIXFactory.java:72)
```

It looks like the version of JNR we are using (3.0.1) just does not work on FreeBSD.

jglick:
3.0.23 does not seem to do any better.

`hudson.util.ProcessTree` would seem convenient…but has no FreeBSD implementation.

jglick:
`hudson.util.jna.GNUCLibrary` does not have seem to have any call allowing us to see if a PID is alive.

jglick:
Think I finally have something working with raw JNA on a FreeBSD slave using 1.609.2 and current Workflow code. If I do nothing, the `sh` step completes with exit code 0. If I `kill -9` the user process, it fails with status 137 (~ 128 + `SIGKILL`). If I `kill -9` the controller process, it fails with -1.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
jglick: Well it is not exactly a “trap”, it just means that in this case some robustness code will not kick in. (For reference, there is currently no liveness check on Windows at all.)

jglick:
Try out the revised patch if you get a chance.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
stephenc: OS-X regularly has PIDs above 32k on my machine...

stephenc: It would seem to me that you could use some of this trick: https://github.com/bytemanproject/byteman/blob/master/contrib/bmunit/src/org/jboss/byteman/contrib/bmunit/BMUnitConfigState.java#L433-L520 to get the PID of the current process... then isAlive on the current JVMs PID should return `true` and you could use fork...join on a JVM that outputs its own PID to get a PID that does not exist and verify that isAlive returns false for that PID

jglick: > to get the PID of the current process

Easily done via `getpgid(0)`.

> and you could use fork...join on a JVM that outputs its own PID to get a PID that does not exist

Oof, way too complicated.

Again, the impact of process 9999 existing is very minor. It simply means that some robustness code is disabled on this slave. If people see this unlikely event as a stumbling block, I would rather just disable this part of the check, and rely only on the check that `getpgid(0) != -1`.

rodrigc:
I tried the latest version of this patch which uses JNA on FreeBSD and MacOS X build slaves.
It worked on both.  Thanks for working on this.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
oleg-nenashev: It makes sense to mark this code as a HACK in any case. Should be an applicable temporary solution, but needs a better fix since there is a corner-case

oleg-nenashev:
@jglick 
It also applies to the original code, so no bugs.
I see the code does not explicitly check the platform before calling this liveness check? Do we presume that <code>BourneShellScript</code> always runs on Unix machine? If yes, this presumption is incorrect in general.

jglick:
> Do we presume that `BourneShellScript` always runs on Unix machine?

No, it might run on Cygwin. I have no test environment for this, so I have no idea if `sh` works in this environment currently, and cannot defend against regressions.

jglick:
> I have no idea if `sh` works in this environment currently

[JENKINS-32017](https://issues.jenkins-ci.org/browse/JENKINS-32017) claims it does not. @jtnord feel like owning this stuff by any chance?

jtnord:
@jglick - is this a case of running with this patch on Windows and cygwin and checking that it works/fails - or are you asking for something else?

jglick:
@jtnord well the linked report claims the plugin did not work on Cygwin to begin with. I just meant generally helping with Windows-specific tasks here.

rodrigc:
Unless you want to explicitly support Cygwin, I think that it would be reasonable
if you make a decision to booby trap the shell step so that it doesn't work at all on Windows, and only have the Batch step and Powershell work.

jglick:
That is how I originally had it but @kohsuke deleted it.

Anyway, we are off topic. Objections to merging this as is?

rodrigc:
I would say this PR should be merged as is.
It definitely solves some problems.

amuniz:
:bee:

jglick:
Released as 1.8-beta-1.

PR #15, opened 2016-01-13
This is my understanding of the javadocs for this method

stephenc:
- Also clarify that in a Jenkins restart, plugins need to re-add all the continued tasks from the previous session before InitMilestone.COMPLETED in order to be ableto properly guarantee the scheduling of those tasks ahead of any other regular tasks in the restored build queue.

@jglick Is this a correct understanding?

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
@@ -52,7 +52,33 @@ public interface ContinuedTask extends Queue.Task {
      * <strong>Implementations must not change the value after it has been initially accessed.</strong>
      * <p>
      * This method is used to identify those tasks that are a continuation of a task execution either currently running
-     * or in a previous Jenkins session. When a job starts a fresh task this method should return {@code false}.
+     * or in a previous Jenkins session.
+     * <p>
+     * There are two ways to start a chain of {@link ContinuableExecutable}s:
+     * <ul>
+     *     <li>
+     *         A regular {@link Queue.Task} that does not implement this interface return a
+     *         {@link ContinuableExecutable} from {@link Queue.Task#createExecutable()} (because if a {@link Queue.Task}
+     *         does not implement {@link ContinuedTask} then that is equivalent to {@link #isContinued()} ==
+     *         {@code false}).
+     *     </li>
+     *     <li>
+     *         A {@link ContinuedTask} with {@link #isContinued()} == {@code false} returns a
+     *         {@link ContinuableExecutable} from {@link #createExecutable()}.
+     *     </li>
+     * </ul>
+     * The chain of {@link ContinuableExecutable}s will continue for as long as either:
+     * <ul>
+     *     <li>
+     *         {@link ContinuableExecutable#run()} completes normally and
+     *         Each {@link ContinuableExecutable#willContinue()} is {@code true} on/after completion.
+     *     </li>
+     *     <li>
+     *         {@link ContinuableExecutable#run()} did not complete normally (i.e. JVM shutdown) and there is a
+     *         {@link ContinuedTask} instance in the {@link Queue} where {@link #isContinued()} ==
+     *         {@code true} that will continue the chain.
+     *     </li>
+     * </ul>
      * After a restart of Jenkins when the plugin implementing ContinuedTask identifies that there are some tasks that
      * may already be in progress it will need to resubmit those tasks to the Queue (before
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -60,6 +76,10 @@ private static boolean isContinued(Queue.Task task) {
             if (isContinued(item.task)) {
                 return null;
             }
+            Jenkins jenkins = Jenkins.getInstance();
+            if (jenkins == null || jenkins.getInitLevel().compareTo(InitMilestone.COMPLETED) < 0) {
```
jglick: Does Jenkins normally allow queue maintenance before completion? If so, seems like a bug in core.

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -45,7 +47,21 @@
 public interface ContinuedTask extends Queue.Task {
 
     /**
-     * True if the task should actually be consider continued now.
+     * Is this task a continuation of an earlier task or a new task.
+     * <p>
+     * <strong>Implementations must not change the value after it has been initially accessed.</strong>
+     * <p>
+     * This method is used to identify those tasks that are a continuation of a task execution either currently running
+     * or in a previous Jenkins session. When a job starts a fresh task this method should return {@code false}.
```
jglick: …or not implement this interface at all.

stephenc:
@jglick how are we doing now?

jglick:
Still a little concerned about `PleaseWait`. This is making a behavioral change for Jenkins as a whole, even when there are no continuable tasks in sight. If Jenkins is doing queue maintenance too early, that is something that ought to be fixed in core.

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
stephenc: So when the queue is [instantiated the MaintainTask is scheduled](https://github.com/jenkinsci/jenkins/blob/ec95aa5b21ab8d5bc0a74bf7553b57523e76766f/core/src/main/java/hudson/model/Queue.java#L308)... that's going to [run every 5 seconds](https://github.com/jenkinsci/jenkins/blob/ec95aa5b21ab8d5bc0a74bf7553b57523e76766f/core/src/main/java/hudson/model/Queue.java#L2706)

The Queue is instantiated [when Jenkins is being constructed](https://github.com/jenkinsci/jenkins/blob/ec95aa5b21ab8d5bc0a74bf7553b57523e76766f/core/src/main/java/jenkins/model/Jenkins.java#L774) after that [we execute the reactor](https://github.com/jenkinsci/jenkins/blob/ec95aa5b21ab8d5bc0a74bf7553b57523e76766f/core/src/main/java/jenkins/model/Jenkins.java#L816) and some time later we [update the computer list](https://github.com/jenkinsci/jenkins/blob/ec95aa5b21ab8d5bc0a74bf7553b57523e76766f/core/src/main/java/jenkins/model/Jenkins.java#L844) which will also [schedule a maintenance](https://github.com/jenkinsci/jenkins/blob/ec95aa5b21ab8d5bc0a74bf7553b57523e76766f/core/src/main/java/hudson/model/AbstractCIBase.java#L207)

The key point here is that the 5 second time starts counting down once the Queue is instantiated... so if the reactor takes more than 5 seconds to get to `COMPLETED` then the queue will start doing maintenance before the reactor is complete...

Sounds like a possible bug... for sure I think it would be better to have the maintenance skipped until we are completed... but until the baseline version of Jenkins has such a fix I believe this is necessary

stephenc:
@jglick hopefully my comments explain the requirement for `PleaseWait` at least until the bug in core is fixed in the baseline version of Jenkins that durable-task depends on

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
KostyaSha: @stephenc would be cool if you could add more comments into Queue code/methods/classes as it  seems only authors knows what is happening in queue.

jglick: I am not comfortable with this plugin modifying a subtle aspect of queue processing in Jenkins even when no existing tasks implement the interface. If there is something wrong in base Jenkins then a bug should be filed for it and it should be fixed in core.

stephenc:
I give up on this PR ever having a path forward

PR #17, opened 2016-02-19
[JENKINS-28400] Better handle failure to start wrapper sh script

jglick:
[JENKINS-28400](https://issues.jenkins-ci.org/browse/JENKINS-28400)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -36,7 +36,6 @@ import hudson.tasks.Shell;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.WeakHashMap;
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -56,6 +60,17 @@ public String getScript() {
         return script;
     }
 
+    /**
+     * Set of workspaces which we have already run a process in.
+     * Copying output from the controller process consumes a Java thread, so we want to avoid it generally.
+     * But we do it the first time we run a process in a new workspace, to assist in diagnosis.
+     * (For example, if we are unable to write to it due to permissions, we want to see that error message.)
+     * Ideally we would display output the first time a given {@link Launcher} was used in that workspace,
```
jglick: Note that `docker.image("IMAGE").inside {sh 'echo hello'}` does _not_ print this diagnostic when the workspace is not writable inside the container, since `docker inspect -f . IMAGE` had previously been run outside the container in the same workspace. Perhaps fixable by running the diagnostics for the first _several_ processes in this workspace, etc.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -46,6 +49,11 @@
  */
 public final class BourneShellScript extends FileMonitoringTask {
 
+    /** Number of times we will show launch diagnostics in a newly encountered workspace before going mute to save resources. */
+    private static /* not final */ int NOVEL_WORKSPACE_DIAGNOSTICS_COUNT = Integer.getInteger(BourneShellScript.class.getName() + ".NOVEL_WORKSPACE_DIAGNOSTICS_COUNT", 10);
+    /** Number of seconds we will wait for a controller script to be launched before assuming the launch failed. */
+    private static /* not final */ int LAUNCH_FAILURE_TIMEOUT = Integer.getInteger(BourneShellScript.class.getName() + ".LAUNCH_FAILURE_TIMEOUT", 15);
```
amuniz: Why not regular UI configuration? JVM startup options are difficult to find and use.

jglick: I do not expect these parameters to need to be configured; I am leaving the door open to experimentation without source code modification in case I receive bug reports or support requests that seem to be related.

amuniz:
:bee:

PR #4, opened 2015-03-05
[JENKINS-25678] Space-in-path bug

jglick:
[JENKINS-25678](https://issues.jenkins-ci.org/browse/JENKINS-25678)

@reviewbybees

PR #18, opened 2016-02-20
[JENKINS-25678] Space-in-path bug, revisited

jglick:
[JENKINS-25678](https://issues.jenkins-ci.org/browse/JENKINS-25678) Amends #4.

Subsumes #17; [incremental diff](https://github.com/jglick/durable-task-plugin/compare/launch-failure-JENKINS-28400...win-space-in-path-JENKINS-25678)

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -54,14 +54,14 @@ public final class WindowsBatchScript extends FileMonitoringTask {
         }
         BatchController c = new BatchController(ws);
 
-        c.getBatchFile1(ws).write(String.format("cmd /c \"%s\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+        c.getBatchFile1(ws).write(String.format("cmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
                 c.getBatchFile2(ws),
                 c.getLogFile(ws),
                 c.getResultFile(ws)
         ), "UTF-8");
         c.getBatchFile2(ws).write(script, "UTF-8");
 
-        Launcher.ProcStarter ps = launcher.launch().cmds("cmd", "/c", c.getBatchFile1(ws).getRemote()).envs(envVars).pwd(ws).quiet(true);
+        Launcher.ProcStarter ps = launcher.launch().cmds("cmd", "/c", "\"\"" + c.getBatchFile1(ws) + "\"\"").envs(envVars).pwd(ws).quiet(true);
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -92,10 +111,22 @@ public String getScript() {
         args.addAll(Arrays.asList("sh", "-c", cmd));
         Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(envVars).pwd(ws).quiet(true);
         listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+/", "") + "] Running shell script"); // -x will give details
-        /* Uncomment for diagnosis (and comment following line) in case wrapper script fails. Otherwise skip since it consumes a thread:
-        ps.stdout(listener);
-        */
-        ps.readStdout().readStderr(); // TODO RemoteLauncher.launch fails to check ps.stdout == NULL_OUTPUT_STREAM, so it creates a useless thread even if you never called stdout(…)
+        boolean novel;
+        synchronized (encounteredPaths) {
```
michaelneale: better to do it this way than a whole `synchronizedMap` around encounteredPaths? (curious)

michaelneale:
:bee:  saw previous comments about config options vs GUI but I think system props are fine for this type of thing (will they be shown in the list of available ones? or this is just for "inside baseball" debugging?)

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
jglick: `synchronizedMap` does not let you atomically increment a value. There may be some utility in Guava etc. that does this.

jglick:
> this is just for "inside baseball" debugging?

Not familiar with that metaphor, but I guess yes. If someone feels the need to tune one of these parameters, either I will have suggested it or I will want to know about it.

jglick:
@michaelneale BTW please review #17 first, which I think all of your comments applied to, and for this PR only review the incremental diff shown in the description. Confusing, I know, but without access to prompt reviews I am forced to create cascading PRs in order to make progress.

amuniz:
:bee:

amuniz:
BTW did you run the test in a Windows box? It is [skipped](https://jenkins.ci.cloudbees.com/job/plugins/job/durable-task-plugin/org.jenkins-ci.plugins$durable-task/73/testReport/org.jenkinsci.plugins.durabletask/WindowsBatchScriptTest) in the PR build (as it runs on Linux) so we don't have a confirmation of the test is working.

jglick:
> did you run the test in a Windows box?

Yes of course, I developed the test and the fix iteratively in Windows (XP in this case). I ought to double-check it in 2012 R2 before merging.

jglick:
Passes in 2012 R2.

PR #19, opened 2016-02-20
[JENKINS-27152] Store control files outside of the workspace

jglick:
[JENKINS-27152](https://issues.jenkins-ci.org/browse/JENKINS-27152)

Supersedes #11.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -140,11 +140,11 @@ public final class BourneShellScript extends FileMonitoringTask {
             super(ws);
         }
 
-        public FilePath getScriptFile(FilePath ws) {
+        public FilePath getScriptFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("script.sh");
         }
 
-        FilePath pidFile(FilePath ws) {
+        FilePath pidFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("pid");
         }
 
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -161,8 +161,20 @@ public abstract class FileMonitoringTask extends DurableTask {
         /**
          * Directory in which this controller can place files.
          */
-        public FilePath controlDir(FilePath ws) {
-            return tempDir(ws).child("durable-" + id);
+        public FilePath controlDir(FilePath ws) throws IOException, InterruptedException {
+            FilePath cd = tempDir(ws).child("durable-" + id);
+            if (cd.isDirectory()) { // normal case 1.8+ after creation
+                return cd;
+            }
+            FilePath oldCD = ws.child("." + id);
+            if (oldCD.isDirectory()) { // compatibility with 1.6
+                return oldCD;
+            }
+            oldCD = ws.child(".jenkins-" + id);
+            if (oldCD.isDirectory()) { // compatibility with 1.7
+                return oldCD;
+            }
+            return cd;
         }
 
         // TODO https://github.com/jenkinsci/jenkins/pull/2066
@@ -173,14 +185,14 @@ public abstract class FileMonitoringTask extends DurableTask {
         /**
          * File in which the exit code of the process should be reported.
          */
-        public FilePath getResultFile(FilePath workspace) {
+        public FilePath getResultFile(FilePath workspace) throws IOException, InterruptedException {
             return controlDir(workspace).child("jenkins-result.txt");
         }
 
         /**
          * File in which the stdout/stderr
          */
-        public FilePath getLogFile(FilePath workspace) {
+        public FilePath getLogFile(FilePath workspace) throws IOException, InterruptedException {
             return controlDir(workspace).child("jenkins-log.txt");
         }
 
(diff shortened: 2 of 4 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -89,6 +90,7 @@ protected FileMonitoringController doLaunch(FilePath workspace, Launcher launche
 
         protected FileMonitoringController(FilePath ws) throws IOException, InterruptedException {
             // can't keep ws reference because Controller is expected to be serializable
+            ws.mkdirs();
```
amuniz: `controlDir(ws)` - just below - is calling `ws.child("."+id);` so the subsequent call to `mkdirs()` would create the `ws` dirs too, so this line is perhaps not necessary (?)

amuniz: Ok, `controlDir` was modified too, never mind.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -107,4 +109,17 @@ public void smokeTest() throws Exception {
...
+    @Test public void cleanWorkspace() throws Exception {
+        Controller c = new BourneShellScript("touch stuff && echo ---`ls -1a`---").launch(new EnvVars(), ws, launcher, listener);
+        while (c.exitStatus(ws, launcher) == null) {
+            Thread.sleep(100);
+        }
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        c.writeLog(ws, baos);
+        assertEquals(0, c.exitStatus(ws, launcher).intValue());
+        assertThat(baos.toString(), containsString("---. .. stuff---"));
+        c.cleanup(ws);
```
amuniz: What is this line testing?

jglick: The call to `cleanup` you mean? Nothing more than that the standard call in the expected method sequence proceeds without error.

armfergom:
:bee:

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -46,6 +49,11 @@
  */
 public final class BourneShellScript extends FileMonitoringTask {
 
+    /** Number of times we will show launch diagnostics in a newly encountered workspace before going mute to save resources. */
+    private static /* not final */ int NOVEL_WORKSPACE_DIAGNOSTICS_COUNT = Integer.getInteger(BourneShellScript.class.getName() + ".NOVEL_WORKSPACE_DIAGNOSTICS_COUNT", 10);
+    /** Number of seconds we will wait for a controller script to be launched before assuming the launch failed. */
+    private static /* not final */ int LAUNCH_FAILURE_TIMEOUT = Integer.getInteger(BourneShellScript.class.getName() + ".LAUNCH_FAILURE_TIMEOUT", 15);
```
tfennelly: Why not `final`?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -145,6 +177,8 @@ private synchronized int pid(FilePath ws) throws IOException, InterruptedExcepti
                     status = -1;
                 }
                 return status;
+            } else if (_pid == 0 && /* compatibility */ startTime > 0 && System.currentTimeMillis() - startTime > 1000 * LAUNCH_FAILURE_TIMEOUT) {
```
tfennelly: How can `startTime > 0`? I don't see it being modified anywhere.

tfennelly:
:bee:

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
jglick: In case you want to use the `/script` console to adjust the value without restarting Jenkins.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
jglick: Set in an instance initializer.

jtnord:
:bee:

jtnord:
So this would mean that you could not upgrade this plugin whilst a durable task was inflight.

This is not unreasonable - but needs to be clearly called out in the release notes that this is guaranteed to break (whereas before you may get lucky?).  Currently the wiki does not say anything about upgrading whilst tasks are in progress (which I think it should if we are saying it is **never** expected to work but you may get lucky)

jglick:
> you could not upgrade this plugin whilst a durable task was inflight

Hmm, good point, and one I neglected in #11. I think that is fixable. If not it will indeed need to be in the release notes.

jtnord:
:bee:

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -77,10 +78,14 @@ protected FileMonitoringController doLaunch(FilePath workspace, Launcher launche
     }
 
     protected static class FileMonitoringController extends Controller {
+
+        /** Absolute path of {@link #controlDir()}. */
+        private String controlDir;
+
         /**
-         * Unique ID among all the {@link Controller}s that share the same workspace.
+         * @deprecated used only in pre-1.8
```
jtnord: nit: should also have `@Deprecated` annotation

jtnord:
:bee:

svanoort:
:bee: from me.  Long-awaited and immensely helpful.

PR #20, opened 2016-04-12
Reverting change made in https://github.com/jenkinsci/durable-task-pl…

DPD-30:
…ugin/commit/7b17a50395bc9be08926dec1ff44005696a61c9e

Trouble shooting issue https://issues.jenkins-ci.org/browse/JENKINS-34150
With loging of batch processing the hang/silent fail of Batch jobs does not occur.
The extra thread may be covering some other race condition in durabletasks.  Possibly when result file is checked for existince and batch is trying to write to it at the same time. Lock/thread race condition.


Aeon512:
I can confirm that which this change being reverted, the problems no longer appear.

jglick:
I think this is superseded by #24; can you confirm?

jglick:
Assuming this is obsolete.

PR #21, opened 2016-04-17
[JENKINS-34150] Test and solution for issue

mkaring:
A test and the solution for the following issue:
https://issues.jenkins-ci.org/browse/JENKINS-34150

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -33,6 +33,7 @@ import hudson.remoting.RemoteOutputStream;
 import hudson.remoting.VirtualChannel;
 import hudson.slaves.WorkspaceList;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.io.RandomAccessFile;
@@ -148,6 +149,9 @@ public abstract class FileMonitoringTask extends DurableTask {
             if (status.exists()) {
                 try {
                     return Integer.parseInt(status.readToString().trim());
+                } catch (FileNotFoundException ignored) {
+                    /* File is locked. Ignore and just try again later. */
+                    return null;
                 } catch (NumberFormatException x) {
                     throw new IOException("corrupted content in " + status + ": " + x, x);
                 }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -55,32 +55,46 @@ public final class WindowsBatchScript extends FileMonitoringTask {
         }
         BatchController c = new BatchController(ws);
 
-        c.getBatchFile1(ws).write(String.format("cmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
-                c.getBatchFile2(ws).getRemote().replace("%", "%%"),
-                c.getLogFile(ws).getRemote().replace("%", "%%"),
-                c.getResultFile(ws).getRemote().replace("%", "%%")
-        ), "UTF-8");
+        /*
+          This wrapper script basically only calls the actual script that is supposed to be executed. It redirects all
+          the output of the target script to the log file and stores the exit code in a result file.
+          The script uses a additional temporary to store the exit code that is moved after writing to is is done.
+          Writing the output file directly may cause problems because the check for the exit code may locate the result
+          file after it was created but before the code was written in and issue a error because of this.
+         */
+        final String nl = "\r\n";
+        c.getBatchFile1(ws).write(
+                "CHCP 65001 > NUL" + nl +
+                "SETLOCAL" + nl +
+                "SET theRealScript=" + c.getBatchFile2(ws).getRemote().replace("%", "%%") + nl +
+                "SET resultFile=" + c.getResultFile(ws).getRemote().replace("%", "%%") + nl +
+                "SET logFile=" + c.getLogFile(ws).getRemote().replace("%", "%%") + nl +
+                "SET tempResultFile=" + c.getTemporaryResultFile(ws).getRemote().replace("%", "%%") + nl +
+                "TYPE NUL > \"%tempResultFile%\"" + nl +
+                "CMD /C \"%theRealScript%\" > \"%logFile%\" 2>&1" + nl +
+                "ECHO %ERRORLEVEL% > \"%tempResultFile%\"" + nl +
+                "MOVE \"%tempResultFile%\" \"%resultFile%\" > NUL" + nl +
+                "ENDLOCAL" + nl +
+                "EXIT 0", "UTF-8");
         c.getBatchFile2(ws).write(script, "UTF-8");
 
-        Launcher.ProcStarter ps = launcher.launch().cmds("cmd", "/c", "\"\"" + c.getBatchFile1(ws) + "\"\"").envs(envVars).pwd(ws).quiet(true);
+        Launcher.ProcStarter ps = launcher.launch()
+                                          .cmds("cmd", "/C", "START \"\" /MIN /WAIT \"" + c.getBatchFile1(ws).getRemote() + '"')
+                                          .envs(envVars)
+                                          .pwd(ws)
+                                          .quiet(true);
         listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+\\\\", "") + "] Running batch script"); // details printed by cmd
         /* Too noisy, and consumes a thread:
         ps.stdout(listener);
         */
-        ps.readStdout().readStderr(); // TODO see BourneShellScript
-        c.setRunningProcess(ps.start());
+        ps.writeStdin().readStdout().readStderr(); // TODO see BourneShellScript
+        Proc process = ps.start();
         return c;
     }
 
     private static final class BatchController extends FileMonitoringController {
-        private Proc runningProcess;
-
         private BatchController(FilePath ws) throws IOException, InterruptedException {
             super(ws);
-        }
-
-        void setRunningProcess(Proc process) {
-            runningProcess = process;
         }
 
         public FilePath getBatchFile1(FilePath ws) throws IOException, InterruptedException {
@@ -91,18 +105,8 @@ public final class WindowsBatchScript extends FileMonitoringTask {
             return controlDir(ws).child("jenkins-main.bat");
         }
 
-        @Override public Integer exitStatus(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
-            Proc proc = runningProcess;
-            if ((proc != null) && proc.isAlive()) {
-                /* Don't check the file in case the process is still running. */
-                return null;
-            }
-            Integer status = super.exitStatus(workspace, launcher);
-            if (status == null) {
-                /* The process is dead, but there is no status. That is a error. */
-                return -1;
-            }
-            return status;
+        public FilePath getTemporaryResultFile(FilePath ws) throws IOException, InterruptedException {
+            return controlDir(ws).createTempFile("jenkins-result", null);
         }
 
         private static final long serialVersionUID = 1L;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -80,14 +80,18 @@ public final class WindowsBatchScript extends FileMonitoringTask {
                 "SET tempResultFile=" + escapeForBatch(c.getTemporaryResultFile(ws)) + nl +
                 "SET pidFile=" + escapeForBatch(c.getPidFile(ws)) + nl +
                 "TITLE=" + identifier + nl + /* This works regardless if the command windows is actually visible. */
-                "FOR /F \"tokens=2,9 delims=,\" %%1 IN ('tasklist /NH /FI \"IMAGENAME eq cmd.exe\" /V /FO CSV') DO (" + nl +
-                "  ECHO %%~2 | FINDSTR /C:" + identifier + " >NUL 2>&1 && ECHO %%~1 > %pidFile%" + nl +
+                "FOR /F \"tokens=*\" %%1 IN ('tasklist /NH /FI \"IMAGENAME eq cmd.exe\" /V /FO CSV') DO (" + nl +
+                "  ECHO %%1 | FINDSTR /C:" + identifier + " >NUL 2>&1 && (" + nl +
+                "    FOR /F \"tokens=2 delims=,\" %%P IN (\"%%1\") DO ( " + nl +
+                "      ECHO %%~P > \"%pidFile%\"" + nl +
+                "    )" + nl +
+                "  )" + nl +
                 ")" + nl +
                 "TYPE NUL > \"%tempResultFile%\"" + nl +
                 "CMD /C \"%theRealScript%\" > \"%logFile%\" 2>&1" + nl +
                 "ECHO %ERRORLEVEL% > \"%tempResultFile%\"" + nl +
                 "MOVE \"%tempResultFile%\" \"%resultFile%\" > NUL" + nl +
-                "DEL %pidFile% > NUL" + nl +
+                "DEL /Q %pidFile% > NUL 2>&1" + nl +
                 "ENDLOCAL" + nl +
                 "EXIT 0", "UTF-8");
         c.getMainBatchFile(ws).write(script, "UTF-8");
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -31,12 +31,19 @@ import hudson.FilePath;
 import hudson.Launcher;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.io.OutputStream;
+import java.util.UUID;
+
+import org.apache.commons.io.output.ByteArrayOutputStream;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
  * Runs a Windows batch script.
  */
 public final class WindowsBatchScript extends FileMonitoringTask {
+
+    /** Number of seconds we will wait for a controller script to be launched before assuming the launch failed. */
+    private static /* not final */ int LAUNCH_FAILURE_TIMEOUT = Integer.getInteger(WindowsBatchScript.class.getName() + ".LAUNCH_FAILURE_TIMEOUT", 15);
     private final String script;
 
     @DataBoundConstructor public WindowsBatchScript(String script) {
@@ -62,6 +69,8 @@ public final class WindowsBatchScript extends FileMonitoringTask {
           file after it was created but before the code was written in and issue a error because of this.
          */
         final String nl = "\r\n";
+        final String identifier = c.getIdentifier();
+
         c.getWrapperBatchFile(ws).write(
                 "CHCP 65001 > NUL" + nl +
                 "SETLOCAL" + nl +
@@ -69,16 +78,22 @@ public final class WindowsBatchScript extends FileMonitoringTask {
                 "SET resultFile=" + escapeForBatch(c.getResultFile(ws)) + nl +
                 "SET logFile=" + escapeForBatch(c.getLogFile(ws)) + nl +
                 "SET tempResultFile=" + escapeForBatch(c.getTemporaryResultFile(ws)) + nl +
+                "SET pidFile=" + escapeForBatch(c.getPidFile(ws)) + nl +
+                "TITLE=" + identifier + nl + /* This works regardless if the command windows is actually visible. */
+                "FOR /F \"tokens=2,9 delims=,\" %%1 IN ('tasklist /NH /FI \"IMAGENAME eq cmd.exe\" /V /FO CSV') DO (" + nl +
+                "  ECHO %%~2 | FINDSTR /C:" + identifier + " >NUL 2>&1 && ECHO %%~1 > %pidFile%" + nl +
+                ")" + nl +
                 "TYPE NUL > \"%tempResultFile%\"" + nl +
                 "CMD /C \"%theRealScript%\" > \"%logFile%\" 2>&1" + nl +
                 "ECHO %ERRORLEVEL% > \"%tempResultFile%\"" + nl +
                 "MOVE \"%tempResultFile%\" \"%resultFile%\" > NUL" + nl +
+                "DEL %pidFile% > NUL" + nl +
                 "ENDLOCAL" + nl +
                 "EXIT 0", "UTF-8");
         c.getMainBatchFile(ws).write(script, "UTF-8");
 
         Launcher.ProcStarter ps = launcher.launch()
-                                          .cmds("cmd", "/C", "START \"\" /MIN /WAIT \"" + escapeForBatch(c.getWrapperBatchFile(ws)) + '"')
+                                          .cmds("cmd", "/C", "START \"\" /MIN \"" + escapeForBatch(c.getWrapperBatchFile(ws)) + '"')
                                           .envs(envVars)
                                           .pwd(ws)
                                           .quiet(true);
@@ -88,6 +103,8 @@ public final class WindowsBatchScript extends FileMonitoringTask {
         */
         ps.writeStdin().readStdout().readStderr(); // TODO see BourneShellScript
         ps.start();
+        c.markStart();
+
         return c;
     }
 
@@ -102,8 +119,15 @@ public final class WindowsBatchScript extends FileMonitoringTask {
     }
 
     private static final class BatchController extends FileMonitoringController {
+
+        private final String identifier;
+        private long startTime;
+        private int pid;
+
         private BatchController(FilePath ws) throws IOException, InterruptedException {
             super(ws);
+
+            identifier = UUID.randomUUID().toString();
         }
 
         FilePath getWrapperBatchFile(FilePath ws) throws IOException, InterruptedException {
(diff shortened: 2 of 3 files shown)


amuniz:
Did you test this patch across Jenkins restarts? I mean, make the batch script to take longer and restart Jenkins while it is being executed.

mkaring:
I have not. How ever this patch shouldn't alter the behaviour of the plugin during a restart of Jenkins since it is not introducing anything that causes the execution to block.

In case this is a concern it may be a good idea to add a general test that verifies to behaviour of the plugin during a restart cycle. I am not even sure what I would have to check. If the restart cycle hangs or if the scripts are properly cleaned?

amuniz:
> this patch shouldn't alter the behaviour of the plugin during a restart of Jenkins since it is not introducing anything that causes the execution to block.

I suspect `runningProcess` will be always null after a restart.

> In case this is a concern it may be a good idea to add a general test that verifies to behaviour of the plugin during a restart cycle

I meant a manual test. An automated _restart_ test would be intrinsically unreliable - as you don't know how long the batch process should wait until Jenkins is restarted.

mkaring:
I am not sure what would happen to the classes during a restart. Does the plugin currently pick a running batch execution up again after a restart?

amuniz:
> Does the plugin currently pick a running batch execution up again after a restart?

Yes. That's the reason this plugin exists.

mkaring:
I understand. There may be some issues with that when using my fix. Even when falling back to the classic check it does not work, because it may trigger the hanging again.

The only proper idea I got now to get around this issue is to alter the created batch files and have them create create a file that marks the process as running that is checked before looking for the exit status file. This could act as fallback method in case the reference to the running process is lost.

I will check if that works later today. Thank you for pointing this issue out.

mkaring:
I updated this pull request with a actually working version of the script that utilizes the `START` command to execute the script. This way there is no more hanging because what ever is killing the parent process can't terminate the wrapper and the actual script any more.

I did some rewriting on the generation and the structure of the wrapper script. That was mostly done so the code is easier readable and any issues that may be caused due to stray variables or due to strange encoding problems may be avoided.

I also updated the test case to check for the issue more aggressive. It now runs multiple instances of the script in order to try to get them to lock up. The test works very reliable now. In case the problem is present, the test will fail.

All in all this is a working solution, but I am not totally happy with it because it still creates those command line windows. But I have no found any way to avoid this aside from using the scripting host, powershell or any third party tools. And all of those may be blocked in a server environment. After all the window spawning is not that big of a problem. I think Jenkins is mostly running on it's own user anyway, so it wouldn't be that much of a problem, because only that user sees those command line windows appearing.

One note that may be relevant: Those windows that now appear. If you close those by hand it cause the same hanging issue as we had before.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -54,20 +55,40 @@ public String getScript() {
...
+                                          .pwd(ws)
+                                          .quiet(true);
         listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+\\\\", "") + "] Running batch script"); // details printed by cmd
         /* Too noisy, and consumes a thread:
         ps.stdout(listener);
         */
-        ps.readStdout().readStderr(); // TODO see BourneShellScript
-        ps.start();
+        ps.writeStdin().readStdout().readStderr(); // TODO see BourneShellScript
+        Proc process = ps.start();
```
jglick: Why?

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -54,20 +55,40 @@ public String getScript() {
...
+                                          .envs(envVars)
+                                          .pwd(ws)
+                                          .quiet(true);
         listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+\\\\", "") + "] Running batch script"); // details printed by cmd
         /* Too noisy, and consumes a thread:
         ps.stdout(listener);
         */
-        ps.readStdout().readStderr(); // TODO see BourneShellScript
-        ps.start();
+        ps.writeStdin().readStdout().readStderr(); // TODO see BourneShellScript
```
jglick: Why?

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
mkaring: That is a left over. Sorry I forgot to remove it. I'll do so in case you accept the general approach of the fix. It has no impact on the behaviour of the plugin.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
mkaring: This additional function causes the process launcher to handle the  input stream internally. I added this as part of some test but since the input stream is not used by the plugin directly I saw no reason to remove it again.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -54,20 +55,40 @@ public String getScript() {
...
+        /*
+          This wrapper script basically only calls the actual script that is supposed to be executed. It redirects all
+          the output of the target script to the log file and stores the exit code in a result file.
+          The script uses a additional temporary to store the exit code that is moved after writing to is is done.
+          Writing the output file directly may cause problems because the check for the exit code may locate the result
+          file after it was created but before the code was written in and issue a error because of this.
+         */
+        final String nl = "\r\n";
+        c.getBatchFile1(ws).write(
+                "CHCP 65001 > NUL" + nl +
```
amuniz: I guess this is UTF-8, right?

amuniz:
I think this is looking good :+1:

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
mkaring: Correct. It shouldn't change much. But I wanted to be sure that the encoding of the command line context and the written files match.

mkaring:
I did some final clean up. In case there are no objections this could be merged and deployed. I don't see a better solution right now.

mkaring:
Is there any issue left that blocks this from being merged?

sonneveld:
I've noticed issues when terminating a batch script with multiple commands. If aborted during a command, it might run multiple commands after it before actually terminating (if at all, I still get hanging sometimes if aborting while running a bat file)

Example script and logs:
[JENKINS-34150.zip](https://github.com/jenkinsci/durable-task-plugin/files/235929/JENKINS-34150.zip)
First successfully terminated.  the second stopped the first ping then started the second before terminating/hanging.

mkaring:
The problem is that the aborting will not work properly, because the process is now running fully detached. If you manage to terminate the window that runs the wrapper script, it will still hang. But Jenkins shouldn't be able to do so, because there is no reference to that process kept around.

But yes... the batch will ignore any abort command.

sonneveld:
I'm only talking about the case where I stop the jenkins job through the web interface.. I am leaving the command window alone while the job runs. I referred to the windows closing because Jenkins managed to close while terminating the job it but still hung.  I think whatever fix is implemented should still allow jobs to be safely stopped through the web interface.

jglick:
> whatever fix is implemented should still allow jobs to be safely stopped through the web interface

Absolutely. You should assume that the node is a JNLP agent running on a Windows server in a dark closet somewhere, or EC2, etc. The usual means of stopping a runaway process is through the web UI, or a `timeout` in a Pipeline script, etc. etc. Interacting with the Windows desktop should be considered a corner case.

See `BourneShellScriptTest.stop`; there is not yet an equivalent test case in `WindowsBatchScriptTest`.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -54,36 +54,70 @@ public String getScript() {
...
+        /*
+          This wrapper script basically only calls the actual script that is supposed to be executed. It redirects all
+          the output of the target script to the log file and stores the exit code in a result file.
+          The script uses a additional temporary to store the exit code that is moved after writing to is is done.
+          Writing the output file directly may cause problems because the check for the exit code may locate the result
+          file after it was created but before the code was written in and issue a error because of this.
+         */
+        final String nl = "\r\n";
+        c.getWrapperBatchFile(ws).write(
+                "CHCP 65001 > NUL" + nl +
```
jtnord: I have a feeling that this will break some setups where the batch commands are written in the platform native default code page.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -148,6 +149,9 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
             if (status.exists()) {
                 try {
                     return Integer.parseInt(status.readToString().trim());
+                } catch (FileNotFoundException ignored) {
+                    /* File is locked. Ignore and just try again later. */
```
jtnord: this comment apears bogus - if the file is locked it will exist - so why is a `FileNotFoundException` thrown?

mkaring: I have not dug into the backing code of the `readToString` function. In case the file was locked the `exists` function returned `true` and the `readToString` function did throw that exception.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
mkaring: The batch files written in the pipelines are all written as UTF-8. So I added this to avoid that the scripts created by this plugin are read with the wrong encoding.

mkaring:
@jglick You have a point, how ever this is not part of this fix, because stopping the batch doesn't work currently no matter if this fix is applied or not. This pull request solely addresses the the hanging issue.

jglick:
> stopping the batch doesn't work currently

No? Did last time I tested it. Perhaps this regressed.

mkaring:
> No? Did last time I tested it. Perhaps this regressed.
We wouldn't have this discussion of the termination of the batch scripts would work properly.

amuniz:
> We wouldn't have this discussion of the termination of the batch scripts would work properly.

Well, no. This PR is about builds **not** terminating by themselves (not about forcibly terminating a build).

Actually, as noted in a JIRA comment, everything works when using a real agent (not master), even surviving a Jenkins restart or build killing. All this PR is about running a `bat` step in master (which is not recommended, although usable in demo/test cases) and that step hangs forever because `jenkins-wrapper.bat` is not fully executed so `jenkins-result.txt` is not created thus the step does not finish.

mkaring:
> which is not recommended

Why? I've got a buildserver with Jenkins that doesn't use any slaves at all. Everything happens on the master. And considering the responses on the issues related to this I am not alone.

> Well, no. This PR is about builds not terminating by themselves (not about forcibly terminating a build)

Once the hanging happens, the termination doesn't work any more too. You can't apply the timeout and you can't cancel it unless you kill the build the hard way.

Back to topic. I have a idea to get the tracking of the process working in a similar way like it works for the shell script.

mkaring:
So I updated the pull request. I altered the exit code is detected. It now looks though the task list of the host system and tries to locate the process by it's bit. The whole bit fetching in the batch script along with the checking if the process is still alive and kicking is not that straight forward and comes with some overhead because the check if the process is running launches another process that does the actual checking on the target system. That is the only idea I had to get this check working. If someone has a better idea, please share.

Regarding the potential issue with terminating the script. I wrote a test case for that case and it works fine. Not sure what else to expect.

sonneveld:
@amuniz I would like to point out that I have been solely testing with a Linux master and multiple Windows JLNP slaves. In this setup I have seen the script termination hang (and job termination issues).  I suspect the success of running a separate agent in that JIRA comment was because the problem seems to clear up temporarily if you restart the agents and master.

sonneveld:
I've tested with the new changes (11f9b05). With the example script (single batch file that runs ping multiple times with -n 20) I've seen this behaviour when terminating the job during the first ping.
1) successful termination (see 479.txt)
2) ping gets terminated, batch file continues (see 482.txt, 484.txt)
3) this is a big one: when the first ping finishes after 20 seconds, the job fails with an exit code -2, but the cmd window is still running (I can see cmd and further pings still working in process explorer on the slave node) (see 507.txt)

[logs-479-507.zip](https://github.com/jenkinsci/durable-task-plugin/files/237742/logs-479-507.zip)

mkaring:
@sonneveld The error code `-2` indicates that the batch failed to detect it's process ID. The detection is dependant on the output of a specific command (`tasklist`). The the output of this command is changed, it may cause this issue. What operating system is your slave?

sonneveld:
@mkaring Two of the nodes that I have been testing on are Windows Server 2012 R2 but I do have some Windows 10 nodes. I can test on those tomorrow.

luebbe:
I've submitted a workaround (at least it works for me) here: https://issues.jenkins-ci.org/browse/JENKINS-34150?focusedCommentId=255821&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-255821

Maybe the problem can also be solved by always returning an exit code and by adding this extra line to the code snippet [generator.]

mkaring:
@sonneveld I just checked on Windows Server 2012 R2, Windows 7 and Windows 10. The output of the commands I credited the problem to looks exactly the same. So that can't be it. What remains is that I ran all tests to this point on the master. I'll try to setup a environment with slaves as soon as I can and check what happens in that case. Maybe I did something wrong when reading the files that causes everything to break in case it runs on a slave.

The only other idea that comes to mind is that the command I am using is blocked on the slave OS. I am not aware that it is actually possible to block the `tasklist` command, but you never know. Could you check if the user running the slave is allowed to execute this command?

sonneveld:
As far as I know the slaves shouldn't be locked down. We just set them up as build nodes. When I log in as the jenkins user that runs the agents, I can run tasklist successfully:

```
Windows 2012 R2
C:\Users\jenkins>tasklist /NH /FI "PID eq 4308" /v /FO CSV
"cmd.exe","4308","RDP-Tcp#53","2","2,348 K","Running","HOSTNAME\jenkins","0:00:00","Command Prompt - cmd - tasklist  /NH /FI "PID eq 4308" /v /FO CSV"

Windows 10
C:\Users\Jenkins>tasklist /NH /FI "PID eq 4488" /v /FO CSV
"cmd.exe","4488","RDP-Tcp#0","1","2,824 K","Running","HOSTNAME\Jenkins","0:00:00","Command Prompt - tasklist  /NH /FI "PID eq 4488" /v /FO CSV"
```

mkaring:
... I just noticed why it fails. It is the difference of the locale settings of my and your operating systems.

It detects the `,` in the memory usage in your case as a separator. And this is the reason why it splits  everything wrong and doesn't detect the `cmd.exe` instance as the one currently running.

I'll try to work around that. Thanks for the input @sonneveld ! I'll get back to you once I fixed this problem.

Aeon512:
@mkaring maybe it's worth trying the approach from http://stackoverflow.com/questions/17666500/how-to-drop-all-but-last-cell-in-csv-using-cmd

In short, do not split the string within the `for /v` loop but use the whole string as an argument for a batch subroutine. In this case, the batch subroutine splits at the comma but ignores comma's within quota (since they are considered data)

mkaring:
My idea currently is to lookup the identifier in the entire line, without splitting it and in case the identifier is present, just the cell that contains the pid.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -54,36 +61,171 @@ public String getScript() {
...
+                "CMD /C \"%theRealScript%\" > \"%logFile%\" 2>&1" + nl +
+                "ECHO %ERRORLEVEL% > \"%tempResultFile%\"" + nl +
+                "MOVE \"%tempResultFile%\" \"%resultFile%\" > NUL" + nl +
+                "DEL /Q %pidFile% > NUL 2>&1" + nl +
+                "ENDLOCAL" + nl +
+                "EXIT 0", "UTF-8");
+        c.getMainBatchFile(ws).write(script, "UTF-8");
+
+        Launcher.ProcStarter ps = launcher.launch()
+                                          .cmds("cmd", "/C", "START \"\" /MIN \"" + escapeForBatch(c.getWrapperBatchFile(ws)) + '"')
```
Aeon512: I guess there should be double quotes \"\" to handle paths with spaces correctly (see JENKINS-25678)

mkaring:
@sonneveld @Aeon512 I fixed the detection of the pid so the additional comma shouldn't cause anymore trouble. I did a quick test and it looks good. I won't be able to do more extensive tests before monday.

Aeon512:
@mkaring I tested this most current version, but unfortunately i doesn't solve the problem.
Now and then the `bat` step still hangs. In this situation there is a cmd window open whose title is set to the `Identifier` and whose last command is `jenkins-main.bat 1>jenkins-log 2>&1`

mkaring:
@Aeon512 If the batch window displays this command, it should be executing your script right now. What script are you trying to run? The window that opens will not show the output of your script, since this output is redirected to the result file.

Aeon512:
@mkaring I'm trying to execute a batch file which I have previously created with the `writeFile` step. The batch file which I'm executing was already successfully executed and is also no longer listed in process explorer. Moreoever, the paths of the `jenkins-main.bat` and `jenkins-log.txt` are no longer available (already deleted by Jenkins?)

mkaring:
@Aeon512 Jenkins only deletes those files once the execution has finished. So the exit code has to be known. At this point is sounds like it works as expected. The only thing strange is the command line window you describe to be still open.

mkaring:
@Aeon512 Could you give a minimal example configuration of the scenario you are trying to execute?

Aeon512:
@mkaring Will try to get a small example working. But might take until Monday though. Will follow up asap.

mkaring:
@Aeon512 No amount of quotes was able to get rid of the problem. It seems there is a problem with the `START` command in case there is a space in the path. I was able to work around this is a fairly easy modification of the command line.

I tried it with all kinds of nasty characters in the paths and the tests don't fail. I will do some more extensive tests tomorrow. Would be nice if you run the new version in your environment.

Saucistophe:
@mkaring Do you refer to this particular bug (For completeness sake) ?http://stackoverflow.com/questions/17674255/why-does-windows-start-command-not-work-with-spaces-in-arguments-and-path

mkaring:
@Saucistophe No idea if it is a bug or just a limitation of the command. But it sounds like just the same problem.

Aeon512:
When trying to build on Win7 I still end up with three failed tests
```
Failed tests:
  WindowsBatchScriptTest.percentInPath:65->testWithPath:78 expected:<0> but was:
<-2>
  WindowsBatchScriptTest.spaceInPath2:60->testWithPath:78 expected:<0> but was:<
1>
  WindowsBatchScriptTest.spaceInPath:55->testWithPath:78 expected:<0> but was:<1
>
```

mkaring:
I have no clue how I missed that. The problem was the `CMD` call inside the wrapper script. Two quotation marks are required there. I fixed that.

Aeon512:
@mkaring 
Thanks for the update. Now all test pass and I don't have any problems anymore regarding project names with spaces.

However, when using your version of the plugin for long running batch jobs (e.g. compilation for a large C++ project which takes around 1 hour - with nearly 100% cpu and memory load all the time) the batch jobs get aborted after around 10 minutes (between 6 and 17 minutes in my testing) and simply return that the batch file finished with errorlevel -1 (in some rare cases also -2).

Looking at the console output, the batch file simply terminated in the middle of the C++ compilation job (using msbuild from Visual Studio 2012 on the console).

Compared with durable-tasks 1.7 (before this issue was introduced) the same job works fine without any problems for several days of continuously building the whole project over and over again.

Hence, something is going still wrong - but honestly i don't have any idea what this could be.

mkaring:
Okay I have no idea why the batch would be terminated under these conditions. Especially after such a long time. The error level `-2` points to the batch file not returning the process id within the expected time frame. I added this as a safeguard to ensure that it does not lock up in case something goes wrong with the PID detection. Maybe this timeout is set too short in case the system  is stressed by other processes at the same time.

Aeon512:
Well, it's not only error level `-2` but also error level `-1`. Example output

```
[build_debug_x64]          C:\Program Files (x86)\Microsoft Visual Studio 11.0\VC\bin\x86_amd64\CL.exe /c .....
[build_debug_x64]          file1.cpp
[build_debug_x64]          file2.cpp
[Pipeline] }
[Pipeline] // wrap <-- for highlighting the output
[Pipeline] }
[Pipeline] // withEnv <-- for specifying additional environment variables
[Pipeline] echo
Error:: hudson.AbortException: script returned exit code -1
```

The corresponding pipeline script is wrapped in a `try catch` block where any error is printed out in the `catch` block.

As you can see, the batch script simply stops executing the middle of the compilation process.

mkaring:
@Aeon512 Could you try to have your compiler write a log file separately? Maybe this gives a indication why the process is terminated.

jtnord:
>> which is not recommended

> Why? I've got a buildserver with Jenkins that doesn't use any slaves at all. Everything happens on the master. And considering the responses on the issues related to this I am not alone.

*  You have virtually no security if you do that all users should be treated as if they where administrators and nothing on your jenkins master should be considered secret from them. Any user that can create a job is an admin - they can access/modify jenkins configuration files on disk.  Any user that can checkin code that gets built is an admin - they can access/modify jenkins configuraton files on disk.
* an errant build script/job can take out our entire master (an errant `rm` will wipe`$JENKINS_HOME`) the master 
* Builds will fight for resources on the master which can cause a slow UI.

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
```diff
@@ -98,4 +101,69 @@ private void testWithPath(String path) throws Exception {
...
+     * did not work properly anymore. This is the reason why it is referencing to the same issue.
+     *
+     * @throws Exception
+     */
+    @Issue("JENKINS-34150")
+    @Test(timeout=5000) public void batchScriptForcedTerminate() throws Exception {
+        StreamTaskListener listener = StreamTaskListener.fromStdout();
+        FilePath ws = j.jenkins.getRootPath().child("ws");
+        Launcher launcher = j.jenkins.createLauncher(listener);
+        WindowsBatchScript script = new WindowsBatchScript("TIMEOUT /T 10");
```
jtnord: `TIMEOUT /T 10` doesn't work correctly when you have no console to read input from and will die

"ERROR: Input redirection is not supported, exiting the process immediately."

jtnord:
I believe I have a 100% reproducible testcase for this issue

It seems that in the case that the parent process has been killed (e.g. the slave dies) then all though the script terminates successfully and the wrapper terminates successfully (checked with [process monitor](https://technet.microsoft.com/en-us/sysinternals/processmonitor.aspx)) there is no attempt to create the result file.
where the parent process has not been killed I never see this issue.

All I needed to do to fix the issue I was observing is add `@echo off` as the first line of the wrapper script.  Basically I beleive it is trying to echo the commands to be run before running the commands and as there is no longer anything consuming the wrappers input/output when echoing the command it is doomed to fail (but strangly not with an exit code that implies something died!??!

sonneveld:
@jtnord I think that's one way of reproducing it but I've seen it without having killed the slave.  There might be some merit to your `@echo off` suggestion though.

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
mkaring: Right. This was not a issue, because the scenario this test was working in spawned a new window that worked as input context. The command can easily be changed to a `ping` or something like this to work around that. I just used the timeout to get a defined runtime.

mkaring:
This problem only ever happened in case the scripts were running on the master. So if you terminate any slave or not shouldn't effect this issue at all.

jtnord:
> This problem only ever happened in case the scripts were running on the master.

Well I (and a few others) observed it when run on a slave :-)

There may be multiple distinct issues all with the same symptoms. (The first unit test committed here was not fixed when I ran it on top of #24)

mkaring:
> There may be multiple distinct issues all with the same symptoms. (The first unit test committed here was not fixed when I ran it on top of #24)

I noticed this problem too, how ever the issue was not that the test ran into a timeout, indicating that the hanging problem happened, but rather this error reading from the result file.

This problem is one I fixed in my changes without mentioning. The change for this problem to happen becomes significant in case the return code file is queried very often. It happens in case the check for the presence for the file happens just in the moment the batch file tries to write it, leading to the file remaining empty. I guess the batch first touches the file handle, opens it for write and then writes to it. This process I think is interrupted by the checks after touching the file. So the file is there, but empty and this causes the exception to be thrown.

In the changes I did I worked around this by writing the return code to a temporary file and renaming the file (`move`) after the result is there. This way the filled right away when java checks for it.

jtnord:
> > There may be multiple distinct issues all with the same symptoms. (The first unit test committed here was not fixed when I ran it on top of #24)
> I noticed this problem too, how ever the issue was not that the test ran into a timeout, indicating that the hanging problem happened, but rather this error reading from the result file.

But actually by removing the timeout the test is fixed :-)
When I removed the 5 second timeout I see the unit test runs and the "getting the exit part" takes just shy of 10 seconds. Given this is a tight loop I'm not so surprised (even though I am on a 4 core machine with HT and an SSD :-) )

Aeon512:
My issues got resolved after applying #24. So far, no problems occurred.

mkaring:
@jtnord Both tests seem to work with your change. On my system it finishes before the timeout. The only "problem" that happens some times is that the `exitCode` function throws a exception because it reads a empty file. I guess this issue is depends highly on how fast the test system is.

But your change really seems to sort it out.

I will fix the tests and send them in as another pull request. Just to have some tests to this issue and to have something come out of this entire discussion. ;-) I think this pull request can be closed since #24 fixes the issue.

jtnord:
@mkaring 
> The only "problem" that happens some times is that the exitCode function throws a exception because it reads a empty file. I guess this issue is depends highly on how fast the test system is.

I had a reasonable spec machine (i7-4712HQ with SSD) so it's pretty fast.  indeed there could be a narrow race condition between the file being created and the data flushed to it, so there probably needs to be another JIRA issue and a retry added to the exit code reader.

>  I think this pull request can be closed 

Closing...

philippe-granet:
Why closing this PR?
According the documentation [https://jenkins.io/doc/pipeline/](https://jenkins.io/doc/pipeline/)
_Accordingly, pipeline functionality is:
Durable: Pipelines can survive both planned and unplanned restarts of your Jenkins master._

It doesn't actually work if your pipeline run on master nodes and you kill Jenkins. When you restart Jenkins, jobs hang.
This PR seams to fix this, no?

jglick:
> It doesn't actually work on Windows if your pipeline run on master nodes and you kill Jenkins.

The preferred fix would be as outlined in [JENKINS-27617](https://issues.jenkins-ci.org/browse/JENKINS-27617).

PR #23, opened 2016-05-16
[JENKINS-34850] Default shell is not used when configured in the glob…

marcel-hamer:
Update to use the default shell when configured in Jenkins.

[JENKINS-34850](https://issues.jenkins-ci.org/browse/JENKINS-34850)


jglick:
:-1:, this is wrong. The default shell _is_ used for the nested process, which is what your argument to the `sh` step would be, unless overridden by an explicit shebang. The _wrapper_ process is written for the Bourne shell and must only ever be run with it.

jglick:
Not a defect.

PR #24, opened 2016-05-19
[JENKINS-34150] Fixes my observed issue.

jtnord:
I have a reproducable tests case in a propratary implementation using this
code that is 100% reproducable.
The simple "`@echo off`" fixes the failing test for me.

It seems that in the case that the parent process has been killed (e.g.
the slave dies) then all though the script terminates successfully and the
wrapper terminates successfully (checked with process monitor) there is no
attempt to create the result file.
where the parent process has not been killed I never see this issue.
All I needed to do to fix the issue I was observing is add `@echo off` as
the first line of the wrapper script. Basically I believe as it is trying to
echo the commands to be run before running the commands and as there is no
longer anything consuming the wrappers input/output when echoing the
command it is doomed to fail (but strangely not with an exit code that
implies something died!??!

@reviewbybees
https://issues.jenkins-ci.org/browse/JENKINS-34150

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -54,7 +54,7 @@ public final class WindowsBatchScript extends FileMonitoringTask {
         }
         BatchController c = new BatchController(ws);
 
-        c.getBatchFile1(ws).write(String.format("cmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+        c.getBatchFile1(ws).write(String.format("@echo off \r\ncmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
                 c.getBatchFile2(ws).getRemote().replace("%", "%%"),
                 c.getLogFile(ws).getRemote().replace("%", "%%"),
                 c.getResultFile(ws).getRemote().replace("%", "%%")


jtnord:
may or may not obsolete #21

stephenc:
:bee:

rsandell:
:bee:

amuniz:
:bee: let's try it!

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -54,7 +54,7 @@ public String getScript() {
         }
         BatchController c = new BatchController(ws);
 
-        c.getBatchFile1(ws).write(String.format("cmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+        c.getBatchFile1(ws).write(String.format("@echo off \r\ncmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
```
jglick: :ant: stray space at end of line

jglick:
:bee: I think, without fully understanding the issue. But

> fixes the failing test for me

so can we copy the test from #21 to defend against regressions, or is it flaky or problematic somehow?

jtnord:
> so can we copy the test from #21 to defend against regressions, or is it flaky or problematic somehow?

I'm not sure we are all seeing the same issue, I can give it a whirl...

jtnord:
@jglick the unit test is still failing with this fix implying that there is something else as well.
However this fix is confirmed to fix the issue for at least [one other person](https://issues.jenkins-ci.org/browse/JENKINS-34150?focusedCommentId=258154&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-258154)

jglick:
Well let us try this one, because the fix is small and looks very safe and apparently fixes at least some of the reported cases.

PR #27, opened 2016-07-12
[JENKINS-26133] API for capturing output of a command

jglick:
Infrastructure for [JENKINS-26133](https://issues.jenkins-ci.org/browse/JENKINS-26133).

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -42,6 +42,7 @@ import java.util.UUID;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.MasterToSlaveFileCallable;
+import org.apache.commons.io.IOUtils;
 
 /**
  * A task which forks some external command and then waits for log and status files to be updated/created.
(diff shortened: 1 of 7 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -157,6 +158,11 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
             }
         }
 
+        @Override public byte[] getOutput(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
+            // TODO could perhaps be more efficient for large files to send a MasterToSlaveFileCallable<byte[]>
+            return IOUtils.toByteArray(getOutputFile(workspace).read());
```
rsandell: Do IOUtils have max heap size protection here or is it possible to allocate to much?

rsandell:
:bee:

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: If you are retrieving a file large enough to have to ask this question, you have got bigger problems, since the return value is going to be serialized in `program.dat`, etc.

PR #28, opened 2016-08-12
[JENKINS-37341] allow sh step to execute under z/OS USS sh

lne3:
[JENKINS-37341](https://issues.jenkins-ci.org/browse/JENKINS-37341)

- under z/OS convert input and output files of sh to/from IBM1047
- on other platforms do not convert as before

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -43,6 +43,7 @@ import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import jenkins.security.MasterToSlaveCallable;
 import org.kohsuke.stapler.DataBoundConstructor;
+import org.apache.commons.io.IOUtils;
 
 /**
  * Runs a Bourne shell script on a Unix node using {@code nohup}.
@@ -85,7 +86,9 @@ public final class BourneShellScript extends FileMonitoringTask {
             listener.getLogger().println("Warning: was asked to run an empty script");
         }
 
-        ShellController c = new ShellController(ws);
+        String encoding = ws.act(new ZosCheck()) ? "IBM1047" : null;
+
+        ShellController c = new ShellController(ws, encoding);
 
         FilePath shf = c.getScriptFile(ws);
 
@@ -95,7 +98,7 @@ public final class BourneShellScript extends FileMonitoringTask {
             String defaultShell = jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel());
             s = "#!"+defaultShell+" -xe\n" + s;
         }
-        shf.write(s, "UTF-8");
+        shf.write(s, encoding==null ? "UTF-8" : encoding);
         shf.chmod(0755);
 
         envVars.put(cookieVariable, "please-do-not-kill-me");
@@ -153,8 +156,8 @@ public final class BourneShellScript extends FileMonitoringTask {
         private int pid;
         private final long startTime = System.currentTimeMillis();
 
-        private ShellController(FilePath ws) throws IOException, InterruptedException {
-            super(ws);
+        private ShellController(FilePath ws, String encoding) throws IOException, InterruptedException {
+            super(ws, encoding);
         }
 
         public FilePath getScriptFile(FilePath ws) throws IOException, InterruptedException {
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -85,7 +86,9 @@ public String getScript() {
             listener.getLogger().println("Warning: was asked to run an empty script");
         }
 
-        ShellController c = new ShellController(ws);
+        String encoding = ws.act(new ZosCheck()) ? "IBM1047" : null;
```
jglick: Seems OK though may clash with the higher-priority #33.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -94,18 +97,41 @@ protected FileMonitoringController doLaunch(FilePath workspace, Launcher launche
...
             FilePath cd = tempDir(ws).child("durable-" + Util.getDigestOf(UUID.randomUUID().toString()).substring(0,8));
             cd.mkdirs();
             controlDir = cd.getRemote();
+            this.encoding = encoding;
         }
 
         @Override public final boolean writeLog(FilePath workspace, OutputStream sink) throws IOException, InterruptedException {
             FilePath log = getLogFile(workspace);
+            if (this.getEncoding() != null)	{
+                sink = new WriterOutputStream(new OutputStreamWriter(sink, "UTF-8"), this.getEncoding());
```
jglick: This is probably superseded by encoding fixes in #29.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -160,7 +189,9 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
 
         @Override public byte[] getOutput(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
             // TODO could perhaps be more efficient for large files to send a MasterToSlaveFileCallable<byte[]>
-            return IOUtils.toByteArray(getOutputFile(workspace).read());
+            return this.getEncoding() == null ?
```
jglick: Ditto.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -95,7 +98,7 @@ public String getScript() {
             String defaultShell = jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel());
             s = "#!"+defaultShell+" -xe\n" + s;
         }
-        shf.write(s, "UTF-8");
+        shf.write(s, encoding==null ? "UTF-8" : encoding);
```
jglick: Would we not just want to use the system encoding on the agent, regardless of OS?

lne3: For various reasons we might be in a situation that the agent does not run with the systems default encoding, i.e. the –Dfile.encoding=… jvm parameter changing it to something else. Therefore system default encoding might be a good default. However, it should be possible to set the encoding explicitly since the required script file encoding (IBM1047 on z/OS) does not necessarily match the encoding set by –Dfile.encoding.

katonica:
Hello, I am new to Jenkins and a few months into a project where we are now required to make use of Jenkins slaves running on z/OS. We are attempting to use pipelines. I took the jenkins slave from http://yourserver:port/jnlpJars/slave.jar and started it on my z/OS system and I was able to get it to connect using the -Dfile.encoding=ISO8859-1 setting. I then run into what I believe to be the same issue as described above when calling sh in my Jenkins pipeline. Is it correct that this durable-task-plugin fix would be something that would help my situation? If so are there any known workarounds while this code is reviewed and integrated? Thanks in advance for all the help!

lne3:
@katonica, thanks for the update. Unfortunately, I had myself no time to spare so far for reviewing the fix with respect to the changes in durable-task-1.13 and its dependencies.

charliegracie:
We are also now running into this issue.  Is there still a chance that this issue will be resolved?

mad8vad:
I'm also now running into this issue

RamchandV:
Any idea when is this getting resolved. I want this change to get my pipeline going :(

gorbadoc:
Doesn't this change have exactly the same goal as [z/OS Unix System Services enablement #80](https://github.com/jenkinsci/durable-task-plugin/pull/80)?  Can this change be abandoned in favor of @OldNavi's change?  What would it take to get #80 merged?

dwnusbaum:
Superseded by #80, which was merged recently and will be available in the next release.

PR #29, opened 2016-09-29
[JENKINS-38381] API to receive asynchronous notifications

jglick:
Related to https://github.com/jenkinsci/workflow-job-plugin/pull/27.

@reviewbybees esp. @oleg-nenashev


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -229,7 +284,104 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                    lastLocation = Long.parseLong(lastLocationFile.readToString());
+                }
+                FilePath logFile = controller.getLogFile(workspace);
+                long len = logFile.length();
+                if (len > lastLocation) {
+                    assert !logFile.isRemote();
+                    try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
+                        InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
+                        InputStream utf8EncodedStream;
+                        Charset nativeEncoding = Charset.defaultCharset();
```
lne3: I assume this is taking the agent's default character set. If so ...
For various reasons (e.g. [JENKINS-13091](https://issues.jenkins-ci.org/browse/JENKINS-13091)) we might be in a situation that the agent does not run with the systems default encoding, i.e. the –Dfile.encoding=… jvm parameter changing it to something else. Therefore system default encoding might be a good default. However, it should be possible to set the encoding explicitly since the required script file encoding (IBM1047 on z/OS) does not necessarily match the encoding set by –Dfile.encoding.

jglick: Resolved in #61.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -229,7 +284,104 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                        if (nativeEncoding.equals(StandardCharsets.UTF_8)) {
+                            utf8EncodedStream = locallyEncodedStream;
+                        } else {
+                            utf8EncodedStream = new ReaderInputStream(new InputStreamReader(locallyEncodedStream, nativeEncoding), StandardCharsets.UTF_8);
+                        }
+                        CountingInputStream cis = new CountingInputStream(utf8EncodedStream);
+                        handler.output(cis);
+                        lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
```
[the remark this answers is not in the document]
jglick: Now applies to #60.

jglick:
Now identical to #62, which better captures the relationship among changes.

PR #30, opened 2016-10-03
Fix Pipeline bat stuck on Windows

philippe-granet:
Problem describe here :
https://issues.jenkins-ci.org/browse/JENKINS-33164?focusedCommentId=259083&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-259083
On Windows, when master or node is killed, batch is also killed. And
when they are restarted, pipeline hang wainting jenkins-log.txt
indefinitely

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -101,6 +101,39 @@ public final class WindowsBatchScript extends FileMonitoringTask {
         public FilePath getBatchFile2(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("jenkins-main.bat");
         }
+        
+        @Override
+        public Integer exitStatus(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
+        	Integer exitStatus = super.exitStatus(workspace, launcher);
+        	if (exitStatus != null) {
+        		return exitStatus;
+        	}
+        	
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -1,116 +1,149 @@
...
-
-        @Override public String getDisplayName() {
-            return Messages.WindowsBatchScript_windows_batch();
-        }
-
-    }
-
-}
+/*
+ * The MIT License
```
jglick: Please create minimal diffs. In particular do not introduce carriage returns into repository sources.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -1,116 +1,149 @@
...
+        	if (exitStatus != null) {
+        		return exitStatus;
+        	}
+        	
+        	// Log file is locked during batch execution
+        	FilePath logFile = getLogFile(workspace);
+        	try{
+    			if(logFile.exists()){
+    				// try to write to log file
+        			// if we can, batch is ended, otherwhise an exception is thrown
```
jglick: And what if the workspace is on a network filesystem without mandatory lock support? Seems inherently unreliable.

jglick:
Does this purport to fix [JENKINS-25053](https://issues.jenkins-ci.org/browse/JENKINS-25053)?

PR #32, opened 2016-12-06
Update BourneShellScript.java

smartmarmot:
This small fix is to show the stdout for debugging otherwise there is no way of doing this.


smartmarmot:
If you have another way of doing this please let me know.

jglick:
> otherwise there is no way of doing this

```groovy
sh 'set -x; …'
```

PR #33, opened 2017-01-03
[JENKINS-40225] replace backslashes for windows shell

blatinville:
[JENKINS-40225](https://issues.jenkins-ci.org/browse/JENKINS-40225)

Tested with 
- cygwin, node connected with ssh server
- git bash using, node connected with JNLP

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -49,6 +49,8 @@ import org.kohsuke.stapler.DataBoundConstructor;
  */
 public final class BourneShellScript extends FileMonitoringTask {
 
+    private static enum OsType {DARWIN, UNIX, WINDOWS}
+
     /** Number of times we will show launch diagnostics in a newly encountered workspace before going mute to save resources. */
     private static /* not final */ int NOVEL_WORKSPACE_DIAGNOSTICS_COUNT = Integer.getInteger(BourneShellScript.class.getName() + ".NOVEL_WORKSPACE_DIAGNOSTICS_COUNT", 10);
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -98,6 +98,11 @@ public String getScript() {
         shf.write(s, "UTF-8");
         shf.chmod(0755);
 
+        scriptPath = shf.getRemote();
+        if (ws.act(new WindowsCheck())) { // JENKINS-40255
```
jglick: To avoid adding round trips this would better be collapsed with `DarwinCheck`.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -98,6 +98,16 @@ public String getScript() {
         shf.write(s, "UTF-8");
         shf.chmod(0755);
 
+        scriptPath = shf.getRemote();
+        List<String> args = new ArrayList<String>();
+
+        if (!ws.act(new DarwinCheck())) { // JENKINS-25848
+            args.add("nohup");
+        }
+        if (ws.act(new WindowsCheck())) { // JENKINS-40255
```
jglick: I guess you missed the point of my suggestion, which was to make a *single* Remoting call which check for both OS X and Windows, for example by returning an `enum`.

blatinville: I'm off this week,  I'll complete changes next week.

tiagomsmagalhaes:
Could someone approve this PR.. I'm also facing this issue and could benefit from the update. Thanks

PR #34, opened 2017-01-10
[JENKINS-40734] Escape $ in environment variable values so it does not get expanded by Launcher

jglick:
[JENKINS-40734](https://issues.jenkins-ci.org/browse/JENKINS-40734)

@reviewbybees


Adnn:
These changes seem to have broken the Windows remotes connected through Cygwin sshd.

oleg-nenashev:
Causes regressions in cases like https://issues.jenkins-ci.org/browse/JENKINS-42875 (EnvInject, valid behavior), which are probably valid. In Jenkins Launcher has been designed to expand variables, so I disagree with the fix approach. Will need to find a workaround for my plugins

PR #35, opened 2017-01-24
Fixes 'isAlive' check for remotes connected through Cygwin.

Adnn:
The isAlive check relies on 'ps -o pid= xxx' invocation to determine liveliness of a task.

As Cygwin's 'ps' does not offer the '-o' option, this invocation always returns a non-zero value, interpreted by the rest of the code as if the task completed.

The proposed fix is to make a simpler usage, in the form 'ps -p xxx', which seems to be valid on most Unix(-like) platforms.


jglick:
Needs `git merge master`.

Is `-p` guaranteed by POSIX?

Note that generally we do not test or support Cygwin and it can be assumed to not work unless you are lucky.

jglick:
No longer necessary.

PR #36, opened 2017-01-24
Quick fix for issue JENKINS-41225 ('sh' step with Cygwin remote)

Adnn:
The call to escape(envVars) on command invocation is preventing the sh step from working (at least under Cygwin environment)

This is a quick fix to allow to use this environment.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -103,17 +103,24 @@ public final class BourneShellScript extends FileMonitoringTask {
         scriptPath = shf.getRemote();
         List<String> args = new ArrayList<String>();
 
+        envVars.put(cookieVariable, "please-do-not-kill-me");
+        // The temporary variable is to ensure JENKINS_SERVER_COOKIE=durable-… does not appear even in argv[], lest it be confused with the environment.
+        Map<String, String> preparedVars = envVars;
+
         OsType os = ws.act(new getOsType());
 
         if (os != OsType.DARWIN) { // JENKINS-25848
             args.add("nohup");
         }
+
         if (os == OsType.WINDOWS) { // JENKINS-40255
             scriptPath= scriptPath.replace("\\", "/"); // cygwin sh understands mixed path  (ie : "c:/jenkins/workspace/script.sh" )
         }
+        else {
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -136,7 +143,7 @@ public String getScript() {
         cmd = cmd.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
 
         args.addAll(Arrays.asList("sh", "-c", cmd));
-        Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(true);
+        Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(preparedVars).pwd(ws).quiet(true);
```
jglick: Amending #34.

jglick:
> reverts the JENKINS-40734 fix for Windows platforms

Presumably on Cygwin only. Or maybe it does not. Since there is no test environment here which could run Cygwin, it is impossible to tell. So long as patches do not touch the behavior on supported platforms, I do not care much.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -103,17 +103,24 @@ public String getScript() {
...
 
         if (os != OsType.DARWIN) { // JENKINS-25848
             args.add("nohup");
         }
+
         if (os == OsType.WINDOWS) { // JENKINS-40255
             scriptPath= scriptPath.replace("\\", "/"); // cygwin sh understands mixed path  (ie : "c:/jenkins/workspace/script.sh" )
         }
+        else {
+            // Escaping creates issues when the connection is established through Cygwin sshd(see issue: JENKINS-41225)
```
akostadinov: I wonder why escaping should matter between connection types. Jenkins should run as a native windows process regardless of connection type. Perhaps `PATH` or other variables differ when you launch through cygwin or not. Or are variables always broken when `bash` is from cygwin regardless of slave connection type.

Original issue [https://issues.jenkins-ci.org/browse/JENKINS-41225](JENKINS-41225) seems to be about failing to find `sh`. I wonder if it can have anything to do with an issue I just found where relative path for "Shell executable" == `bash` fails to work with pipeline step but works with free-style project shell build step. Maybe this is already solved in free-style shell step and same can be applied to the pipeline step.

car-roll:

PR #37, opened 2017-02-06
retry on accessing pid file for bourne shell

gabemontero:
We have itermittently ran into the following type of exception:

```
java.io.IOException: corrupted content in /var/lib/jenkins/jobs/qwer/workspace@tmp/durable-129c54bd/pid: java.lang.NumberFormatException: For input string: ""
	at org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController.pid(BourneShellScript.java:175)
	at org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController.exitStatus(BourneShellScript.java:187)
```

But when we then look at the `/var/lib/jenkins/jobs/qwer/workspace@tmp/durable-129c54bd/pid` after the exception has occurred, it in fact has a valid PID in it.

The implication being that a timing window between when the pid file is created and when it is populated occurred.

I attempted some quick research into commits/issues/jiras to see if such an item has been discussed and mitigated in the past and found none (apologies if my search was not suffiicient).

In any event here is a proposed code change to address for all users of the plugin (vs. our code doing something similar around our calls to `exitStatus`).

Thoughts?  thanks and regards

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -179,10 +179,20 @@ public final class BourneShellScript extends FileMonitoringTask {
             if (pid == 0) {
                 FilePath pidFile = pidFile(ws);
                 if (pidFile.exists()) {
-                    try {
-                        pid = Integer.parseInt(pidFile.readToString().trim());
-                    } catch (NumberFormatException x) {
-                        throw new IOException("corrupted content in " + pidFile + ": " + x, x);
+                    for ( int tries = 30; tries > 0; tries-- ) {
+                        String _pid = "";
+                        try {
+                            _pid = pidFile.readToString().trim();
+                            pid = Integer.parseInt(_pid);
+                        } catch (NumberFormatException x) {
+                            if ( tries == 1 ) {
+                                throw new IOException("corrupted content in " + pidFile + ": " + x, x);
+                            }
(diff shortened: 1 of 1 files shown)


gabemontero:
Ah ... did find a similar situation with https://github.com/jenkinsci/durable-task-plugin/pull/17, but just at a different spot of the flow.

jglick:
There are other possible reasons for corrupted content—did you manage to reproduce a problem and show that this patch fixes it?

Might be better to just have the wrapper script write the exit status to another file and then `mv` it to the final destination, which should prevent race conditions if that is actually the explanation.

gabemontero:
hey @jglick - yes, when this PR was first crafted, we had an env seeing this with enough regularity that when I applied this change, and the problem went away, we became confident this fixed it.

Also, while waiting for this change, we implemented a similar change in our plugin that leverages DTP as a dependency, and that also addressed the problematic env at the time, and has held up since first crafted back in February.  Here is the link:  https://github.com/jenkinsci/openshift-client-plugin/blob/master/src/main/java/com/openshift/jenkins/plugins/pipeline/OcAction.java#L49-L64

Certainly we can live with our work around, but if others could benefit (i.e @oleg-nenashev feedback), I certainly would welcome removing our workaround and defer to this plugin.

thanks

jglick:
No longer necessary.

gabemontero:
thanks for the heads up on https://github.com/jenkinsci/durable-task-plugin/pull/49 @jglick 

will look to pull in v1.16 one way or the other

PR #38, opened 2017-03-10
[FIXED JENKINS-42671] - Add the compatibility warning to the UC

oleg-nenashev:
This change adds the missing compatibility warning.
So the users upgrading the dependent plugins should see this warning after the fix.

The target release is also changed to 1.13.1

https://issues.jenkins-ci.org/browse/JENKINS-42671

@reviewbybees @jglick and @daniel-beck

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
         <version>2.7</version>
     </parent>
     <artifactId>durable-task</artifactId>
-    <version>1.14-SNAPSHOT</version>
+    <version>1.13.1-SNAPSHOT</version>
     <packaging>hpi</packaging>
     <name>Durable Task Plugin</name>
     <description>Library offering an extension point for processes which can run outside of Jenkins yet be monitored.</description>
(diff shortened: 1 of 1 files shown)


jglick:
This is probably overkill since the only plugin known to be affected was `workflow-durable-task-step`, and an update to that was already released which fixes the regression. So the user does not need to do anything other than update both plugins.

oleg-nenashev:
I think it still makes sense to do it, because...

1) Actually this case was really working in EnvInject: https://issues.jenkins-ci.org/browse/JENKINS-42875
2) .... and Pipeline support is in Limbo in EnvInject (https://issues.jenkins-ci.org/browse/JENKINS-42876)
3) In the current state users do not get a warning about the potential compat issue if the plugin is being updated as a dependency. Hence it may be installed silently with a following fallout on Jenkins instances using such undocumented feature in EnvInject and maybe similar other plugins

dwnusbaum:
I just closed the linked Jira as wont-do, so I am closing this PR as well.

PR #39, opened 2017-03-19
[FIXED JENKINS-34581] Adding Powershell support

gabloe:
[JENKINS-34581](https://issues.jenkins-ci.org/browse/JENKINS-34581)

This pull request enables the Durable Task plugin to support powershell without relying on wrapping it around Windows Batch.  This depends on a second pull request into the Workflow Durable Task Step plugin repo as well.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
             <url>http://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
-    
+
     <scm>
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
--- a/pom.xml
+++ b/pom.xml
@@ -37,15 +37,6 @@
         </pluginRepository>
     </pluginRepositories>
     
-    <dependencies>
-      <dependency>
-        <groupId>org.jenkins-ci.plugins</groupId>
-        <artifactId>powershell</artifactId>
-        <version>1.3</version>
-        <optional>true</optional>
-      </dependency>
-    </dependencies>
-    
     <scm>
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.7</version>
+        <version>2.26</version>
+        <relativePath />
     </parent>
     <artifactId>durable-task</artifactId>
     <version>1.14-SNAPSHOT</version>
@@ -14,26 +15,25 @@
     <licenses>
         <license>
             <name>MIT License</name>
-            <url>http://www.opensource.org/licenses/mit-license.php</url>
+            <url>https://opensource.org/licenses/MIT</url>
         </license>
     </licenses>
     
     <properties>
-      <jenkins.version>1.609.1</jenkins.version>
-      <java.level>6</java.level>
+      <jenkins.version>1.642.3</jenkins.version>
       <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
     
     <repositories>
         <repository>
             <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
+            <url>https://repo.jenkins-ci.org/public/</url>
         </repository>
     </repositories>
     <pluginRepositories>
         <pluginRepository>
             <id>repo.jenkins-ci.org</id>
-            <url>http://repo.jenkins-ci.org/public/</url>
+            <url>https://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
 
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,16 @@
             <url>http://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
-
+    
+    <dependencies>
+      <dependency>
+        <groupId>org.jenkins-ci.plugins</groupId>
+        <artifactId>powershell</artifactId>
+        <version>1.3</version>
+        <optional>true</optional>
+      </dependency>
+    </dependencies>
+    
     <scm>
         <connection>scm:git:git://github.com/jenkinsci/${project.artifactId}-plugin.git</connection>
         <developerConnection>scm:git:git@github.com:jenkinsci/${project.artifactId}-plugin.git</developerConnection>
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -67,11 +67,13 @@ public final class PowershellScript extends FileMonitoringTask {
         
         String cmd;
         if (capturingOutput) {
-            cmd = String.format("$(& \"%s\" | Out-File -FilePath \"%s\" -Encoding UTF8) 2>&1 3>&1 4>&1 5>&1 | Out-File -FilePath \"%s\" -Encoding UTF8; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII;", 
+            cmd = String.format("$(& \"%s\" | Out-File -FilePath \"%s\" -Encoding UTF8) 2>&1 3>&1 4>&1 5>&1 | Out-File -FilePath \"%s\" -Encoding UTF8; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII; $outputWithBom = Get-Content \"%s\"; [IO.File]::WriteAllLines(\"%s\", $outputWithBom);", 
                 quote(c.getPowershellMainFile(ws)),
-                quote(c.getOutputFile(ws)),
+                quote(c.getTemporaryOutputFile(ws)),
                 quote(c.getLogFile(ws)),
-                quote(c.getResultFile(ws)));
+                quote(c.getResultFile(ws)),
+                quote(c.getTemporaryOutputFile(ws)),
+                quote(c.getOutputFile(ws)));
         } else {
             cmd = String.format("& \"%s\" *>&1 | Out-File -FilePath \"%s\" -Encoding UTF8; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII;",
                 quote(c.getPowershellMainFile(ws)),
@@ -100,7 +102,7 @@ public final class PowershellScript extends FileMonitoringTask {
     }
     
     private static String quote(FilePath f) {
-        return f.getRemote().replace("%", "%%");
+        return f.getRemote().replace("$", "`$");
     }
 
     private static final class PowershellController extends FileMonitoringController {
@@ -118,6 +120,10 @@ public final class PowershellScript extends FileMonitoringTask {
         
         public FilePath getPowershellWrapperFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("powershellWrapper.ps1");
+        }
+        
+        public FilePath getTemporaryOutputFile(FilePath ws) throws IOException, InterruptedException {
+            return controlDir(ws).child("temporaryOutput.txt");
         }
 
         private static final long serialVersionUID = 1L;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -67,7 +67,7 @@ public final class PowershellScript extends FileMonitoringTask {
         
         String cmd;
         if (capturingOutput) {
-            cmd = String.format("$Error.Clear(); & \"%s\" 3>&1 4>&1 5>&1 6>&1 | Out-File -FilePath \"%s\" -Encoding UTF8; $Error | Out-File -FilePath \"%s\" -Encoding UTF8; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII;", 
+            cmd = String.format("$(& \"%s\" | Out-File -FilePath \"%s\" -Encoding UTF8) 2>&1 3>&1 4>&1 5>&1 | Out-File -FilePath \"%s\" -Encoding UTF8; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII;", 
                 quote(c.getPowershellMainFile(ws)),
                 quote(c.getOutputFile(ws)),
                 quote(c.getLogFile(ws)),
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -91,7 +91,7 @@ public final class PowershellScript extends FileMonitoringTask {
         }
 
         Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(true);
-        listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+\\\\", "") + "] Running PowerShell script");
+        listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+(\\\\|/)", "") + "] Running PowerShell script");
         ps.readStdout().readStderr();
         ps.start();
 
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -40,6 +40,7 @@ import org.kohsuke.stapler.DataBoundConstructor;
 /**
  * Runs a Powershell script
  */
+ @Issue("JENKINS-34581")
 public final class PowershellScript extends FileMonitoringTask {
     private final String script;
     private boolean capturingOutput;
@@ -61,6 +62,7 @@ public final class PowershellScript extends FileMonitoringTask {
         List<String> args = new ArrayList<String>();
         PowershellController c = new PowershellController(ws);
         
+        // Using redirection in powershell produces extra newlines in output... replacing with [io.file]::WriteAllText to prevent corrupted output of exit code\result
         String cmd;
         if (capturingOutput) {
             cmd = String.format("$res = (& \"%s\" | Out-String) 2> \"%s\"; [io.file]::WriteAllText(\"%s\",$(if($LastExitCode -eq $null -or $LastExitCode -eq 0) {0} else {$LastExitCode})); [io.file]::WriteAllText(\"%s\",$res);", 
@@ -75,11 +77,13 @@ public final class PowershellScript extends FileMonitoringTask {
                 quote(c.getResultFile(ws)),
                 quote(c.getLogFile(ws)));
         }
-       
+        
+        // Write the script and execution wrapper to powershell files in the workspace
         c.getPowershellMainFile(ws).write(script, "UTF-8");
         c.getPowershellWrapperFile(ws).write(cmd, "UTF-8");
 
         if (launcher.isUnix()) {
+            // Open Powershell does not support ExecutionPolicy
             args.addAll(Arrays.asList("powershell", "-NonInteractive", "-File", c.getPowershellWrapperFile(ws).getRemote()));
         } else {
             args.addAll(Arrays.asList("powershell.exe", "-NonInteractive", "-ExecutionPolicy", "Bypass", "-File", c.getPowershellWrapperFile(ws).getRemote()));    
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -81,7 +81,7 @@ public final class PowershellScript extends FileMonitoringTask {
 
         // Write the script and execution wrapper to powershell files in the workspace
         c.getPowershellScriptFile(ws).write(script, "UTF-8");
-        c.getPowershellMainFile(ws).write("$error.Clear();\r\n$LastExitCode = $null;\r\ntry {\r\n& '" + quote(c.getPowershellScriptFile(ws)) + "'\r\n} catch {\r\nWrite-Error $_; exit 1;\r\n}\r\nfinally {\r\nif ($LastExitCode -ne $null -and $LastExitCode -ne 0) {\r\nexit $LastExitCode;\r\n} elseif ($error.Count -gt 0 -or !$?) {\r\nexit 1;\r\n} else {\r\nexit 0;\r\n}\r\n}", "UTF-8");
+        c.getPowershellMainFile(ws).write("try {\r\n& '" + quote(c.getPowershellScriptFile(ws)) + "'\r\n} catch {\r\nWrite-Error $_; exit 1;\r\n}\r\nfinally {\r\nif ($LastExitCode -ne $null) {\r\nexit $LastExitCode;\r\n} elseif ($error.Count -gt 0 -or !$?) {\r\nexit 1;\r\n} else {\r\nexit 0;\r\n}\r\n}", "UTF-8");
         c.getPowershellWrapperFile(ws).write(cmd, "UTF-8");
 
         if (launcher.isUnix()) {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -66,14 +66,14 @@ public final class PowershellScript extends FileMonitoringTask {
         String cmd;
         if (capturingOutput) {
             // Using redirection in PowerShell produces extra newlines in output, so I am using [io.file]::WriteAllText to prevent corrupted output of exit code
-            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"] = \"ASCII\"; & \"%s\" *> \"%s\" 2> \"%s\"; $LastExitCode > \"%s\"; $error > \"%s\";", 
+            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"] = \"UTF8\"; & \"%s\" *> \"%s\" 2> \"%s\"; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII; $error | Out-File -FilePath \"%s\" -Encoding ASCII;", 
                 quote(c.getPowershellMainFile(ws)),
                 quote(c.getOutputFile(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)),
                 quote(c.getLogFile(ws)));
         } else {
-            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"] = \"ASCII\"; & \"%s\" *> \"%s\"; $LastExitCode > \"%s\";",
+            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"] = \"UTF8\"; & \"%s\" *> \"%s\"; $LastExitCode | Out-File -FilePath \"%s\" -Encoding ASCII;",
                 quote(c.getPowershellMainFile(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)));
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -81,7 +81,7 @@ public final class PowershellScript extends FileMonitoringTask {
 
         // Write the script and execution wrapper to powershell files in the workspace
         c.getPowershellScriptFile(ws).write(script, "UTF-8");
-        c.getPowershellMainFile(ws).write("$error.Clear();\r\n$LastExitCode = $null;\r\ntry {\r\n& " + quote(c.getPowershellScriptFile(ws)) + "\r\n} catch {\r\nWrite-Error $_; exit 1;\r\n}\r\nfinally {\r\nif ($LastExitCode -ne $null -and $LastExitCode -ne 0) {\r\nexit $LastExitCode;\r\n} elseif ($error.Count -gt 0 -or !$?) {\r\nexit 1;\r\n} else {\r\nexit 0;\r\n}\r\n}", "UTF-8");
+        c.getPowershellMainFile(ws).write("$error.Clear();\r\n$LastExitCode = $null;\r\ntry {\r\n& '" + quote(c.getPowershellScriptFile(ws)) + "'\r\n} catch {\r\nWrite-Error $_; exit 1;\r\n}\r\nfinally {\r\nif ($LastExitCode -ne $null -and $LastExitCode -ne 0) {\r\nexit $LastExitCode;\r\n} elseif ($error.Count -gt 0 -or !$?) {\r\nexit 1;\r\n} else {\r\nexit 0;\r\n}\r\n}", "UTF-8");
         c.getPowershellWrapperFile(ws).write(cmd, "UTF-8");
 
         if (launcher.isUnix()) {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -66,15 +66,14 @@ public final class PowershellScript extends FileMonitoringTask {
         String cmd;
         if (capturingOutput) {
             // Using redirection in PowerShell produces extra newlines in output, so I am using [io.file]::WriteAllText to prevent corrupted output of exit code
-            cmd = String.format("$PSDefaultParameterValues = @{\"Out-File:Encoding\"=\"ASCII\"}; & \"%s\" *> \"%s\" 2> \"%s\"; [io.file]::WriteAllText(\"%s\",$LastExitCode); [io.file]::AppendAllText(\"%s\",$error);", 
+            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"] = \"ASCII\"; & \"%s\" *> \"%s\" 2> \"%s\"; $LastExitCode > \"%s\"; $error > \"%s\";", 
                 quote(c.getPowershellMainFile(ws)),
                 quote(c.getOutputFile(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)),
                 quote(c.getLogFile(ws)));
-
         } else {
-            cmd = String.format("$PSDefaultParameterValues = @{\"Out-File:Encoding\"=\"ASCII\"}; & \"%s\" *> \"%s\"; [io.file]::WriteAllText(\"%s\",$LastExitCode);",
+            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"] = \"ASCII\"; & \"%s\" *> \"%s\"; $LastExitCode > \"%s\";",
                 quote(c.getPowershellMainFile(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)));
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -76,7 +76,7 @@ public final class PowershellScript extends FileMonitoringTask {
         String cmd;
         if (capturingOutput) {
             // Using redirection in PowerShell produces extra newlines in output, so I am using [io.file]::WriteAllText to prevent corrupted output of exit code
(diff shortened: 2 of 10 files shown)


gabloe:
It seems the build system does not have powershell, which has caused the build checks to fail.

HRMPW:
@rtyler is this something that can be corrected?

gabloe:
Withdrawing this pull request to do some additional testing

jglick:
Can keep it open, no harm.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -0,0 +1,139 @@
...
+
+public class PowershellScriptTest {	
+    @Rule public JenkinsRule j = new JenkinsRule();
+
+    private StreamTaskListener listener;
+    private FilePath ws;
+    private Launcher launcher;
+
+    @Before public void vars() {
+        listener = StreamTaskListener.fromStdout();
```
jglick: This would be a good time to `Assume` that `powershell.exe` is in `%Path%` (or whatever).

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,179 @@
...
+                quote(c.getPowershellMainFile(ws)),
+                quote(c.getResultFile(ws)),
+                quote(c.getLogFile(ws)));
+        }
+        
+        // Write the script and execution wrapper to powershell files in the workspace
+        c.getPowershellMainFile(ws).write(script + "\r\nexit 0", "UTF-8");
+        c.getPowershellWrapperFile(ws).write(cmd, "UTF-8");
+        
+        // Try to load the PowerShell plugin to produce the command line arguments for running the script
```
jglick: I am not very fond of the optional dependency; do we actually gain anything from this mode?

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,179 @@
...
+        Plugin powershellPlugin = Jenkins.getInstance().getPlugin("powershell");
+        boolean notLoadedOrError = false;
+        if (powershellPlugin != null && powershellPlugin.getWrapper().isEnabled()) {
+            try {
+                ClassLoader cl = powershellPlugin.getWrapper().classLoader;
+                Class<?> powershellClass = cl.loadClass("hudson.plugins.powershell.PowerShell");
+                Object powershellInst = powershellClass.getConstructor(String.class).newInstance(script);
+                Method m = powershellClass.getMethod("buildCommandLine",FilePath.class);
+                String[] cmdLine = (String[])m.invoke(powershellInst,c.getPowershellWrapperFile(ws));
```
jglick: No need for reflection here. _If_ the plugin is enabled, you can make static calls to `PowerShell.buildCommandLine`.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
gabloe: You have a good point.  I was trying to be a good citizen by reusing some functionality from the powershell plugin.  But, I think we can safely remove the dependency and use the default params defined here as I don't believe the command line arguments for running a powershell script would change.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
gabloe: Good point.  I'll add this.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -0,0 +1,168 @@
...
+            Thread.sleep(100);
+        }
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        c.writeLog(ws, baos);
+        assertEquals(0, c.exitStatus(ws, launcher).intValue());
+        assertThat(new String(c.getOutput(ws, launcher)), containsString("Hello, World!"));
+        c.cleanup(ws);
+    }
+        
+    @Test public void output() throws Exception {
```
slide: This test doesn't seem necessary for powershell, since you already test output in other tests. I believe this test was there for batch files because of the ability to turn off echoing of the command via \@echo.

gabloe: I've removed it

jglick: @slide actually the `output` test is potentially useful for each `DurableTask` implementation, to ensure that `getOutput` works. I have no particular reason to believe the implementation inherited from `FileMonitoringController` would not work across the board, but Windows has a way of surprising us!

pom.xml
```diff
@@ -36,7 +36,7 @@
             <url>http://repo.jenkins-ci.org/public/</url>
         </pluginRepository>
     </pluginRepositories>
-
+
```
jglick: For future reference, it is nice to check your diffs when submitting PRs or PR revisions to make sure they are minimal. If you want to revert everything from a given file,

    git checkout master -- pom.xml

while reverting individual hunks is a bit more work.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    @SuppressFBWarnings(value="VA_FORMAT_STRING_USES_NEWLINE", justification="%n from master might be \\n")
+    @Override protected FileMonitoringController doLaunch(FilePath ws, Launcher launcher, TaskListener listener, EnvVars envVars) throws IOException, InterruptedException {
+        List<String> args = new ArrayList<String>();
+        PowershellController c = new PowershellController(ws);
+        
+        String cmd;
+        if (capturingOutput) {
+            // Using redirection in PowerShell produces extra newlines in output, so I am using [io.file]::WriteAllText to prevent corrupted output of exit code
```
jglick: The `trim` in `FileMonitoringController.exitStatus` does not suffice?

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,129 @@
...
+            cmd = String.format("$PSDefaultParameterValues = @{\"Out-File:Encoding\"=\"ASCII\"}; & \"%s\" *> \"%s\" 2> \"%s\"; [io.file]::WriteAllText(\"%s\",$LastExitCode); [io.file]::AppendAllText(\"%s\",$error);",
```
jglick: For my reference, I suppose the last

```ps
    [io.file]::AppendAllText("…logFile",$error)
```

is [printing error messages](http://stackoverflow.com/a/17461657/12916) to the log (like `stderr`).

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,129 @@
...
+                quote(c.getPowershellMainFile(ws)),
+                quote(c.getLogFile(ws)),
+                quote(c.getResultFile(ws)));
+        }
+
+        // Write the script and execution wrapper to powershell files in the workspace
+        c.getPowershellMainFile(ws).write("try {\r\n" + script + "\r\n} catch {\r\nWrite-Output $_; exit 1;\r\n}\r\nif ($LastExitCode -ne $null -and $LastExitCode -ne 0) {\r\nexit $LastExitCode;\r\n} elseif ($error.Count -gt 0 -or !$?) {\r\nexit 1;\r\n} else {\r\nexit 0;\r\n}", "UTF-8");
+        c.getPowershellWrapperFile(ws).write(cmd, "UTF-8");
+
+        if (launcher.isUnix()) {
```
jglick: Oh boy, [another test axis value](https://github.com/PowerShell/PowerShell#get-powershell). :-)

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,129 @@
...
+            // Open-Powershell does not support ExecutionPolicy
+            args.addAll(Arrays.asList("powershell", "-NonInteractive", "-File", c.getPowershellWrapperFile(ws).getRemote()));
+        } else {
+            args.addAll(Arrays.asList("powershell.exe", "-NonInteractive", "-ExecutionPolicy", "Bypass", "-File", c.getPowershellWrapperFile(ws).getRemote()));    
+        }
+
+        Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(true);
+        listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+\\\\", "") + "] Running PowerShell script");
```
jglick: I guess the substitution should include `/` to work on Linux.

jglick:
On XP using

```
PS …> $PSVersionTable

Name                           Value
----                           -----
CLRVersion                     2.0.50727.3655
BuildVersion                   6.0.6002.18111
PSVersion                      2.0
WSManStackVersion              2.0
PSCompatibleVersions           {1.0, 2.0}
SerializationVersion           1.1.0.1
PSRemotingProtocolVersion      2.1
```

I get:

```
Tests run: 6, Failures: 3, Errors: 1, Skipped: 0, Time elapsed: 64.126 sec <<< FAILURE! - in org.jenkinsci.plugins.durabletask.PowershellScriptTest
implicitExit(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 47.379 sec  <<< FAILURE!
java.lang.AssertionError:
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.implicitExit(PowershellScriptTest.java:97)

explicitExit(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 1.936 sec  <<< FAILURE!
java.lang.AssertionError:
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.explicitExit(PowershellScriptTest.java:82)

verbose(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 3.226 sec  <<< ERROR!
java.io.FileNotFoundException: C:\DOCUME~1\JESSEG~1\LOCALS~1\Temp\hudson3512122869690423700test\ws@tmp\durable-59e128a4\output.txt (The system cannot find the file specified)
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.verbose(PowershellScriptTest.java:138)

echoEnvVar(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 2.846 sec  <<< FAILURE!
java.lang.AssertionError:

Expected: a string containing "envvar=power$hell"
     but: was ""
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.echoEnvVar(PowershellScriptTest.java:150)
```

On Windows 10 using

```
PS …> $PSVersionTable

Name                           Value
----                           -----
PSVersion                      5.0.10586.494
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.10586.494
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

I get:

```
Tests run: 6, Failures: 4, Errors: 0, Skipped: 0, Time elapsed: 78.452 sec <<< FAILURE! - in org.jenkinsci.plugins.durabletask.PowershellScriptTest
implicitExit(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 47.451 sec  <<< FAILURE!
java.lang.AssertionError:
 ■S u c c e s s !

        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.implicitExit(PowershellScriptTest.java:97)

explicitExit(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 4.593 sec  <<< FAILURE!
java.lang.AssertionError:
 ■H e l l o ,   W o r l d !

        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.explicitExit(PowershellScriptTest.java:82)

verbose(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 4.28 sec  <<< FAILURE!
java.lang.AssertionError:

Expected: a string containing "Hello, World!"
     but: was " ■H e l l o ,   W o r l d !
 "
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.verbose(PowershellScriptTest.java:138)

echoEnvVar(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 3.899 sec  <<< FAILURE!
java.lang.AssertionError:

Expected: a string containing "envvar=power$hell"
     but: was " ■e n v v a r = p o w e r $ h e l l
 "
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.echoEnvVar(PowershellScriptTest.java:150)
```

which looks like a UTF-16 issue offhand.

On 2012 using PS 3 I get the same failures as on 10.

Is there a particular minimum Powershell or Windows version you have tested with and expect to support (in which case `help.html` in the downstream PR should say so and your test here should `Assume` that), or do these look like bugs?

I can work on getting this plugin tested on ci.jenkins.io which should offer automated Windows builds [on 2012 R2](https://github.com/jenkins-infra/documentation/blob/master/ci.adoc#node-labels).

slide:
@jglick, there are already tests that test output. The specific test I mentioned was copied from the bat tests which have a specific test for not having additional output when using specific settings, e.g., @echo off which is not valid for powershell

slide:
I don't think Windows XP should be supported...it was completely EOL'd three years ago. That's just an opinion though, I'm not sure what the official Jenkins policy for Windows support is.

slide:
It looks like PSDefaultParameterValues is only valid for PowerShell 3.0 and above. This means that Windows 7 would be the lowest version that could be supported.

gabloe:
I believe that this requires powershell 3.0 or higher unless we can find a different solution for forcing ASCII writes.

-------------------------------------------------------
 T E S T S
-------------------------------------------------------
Running InjectedTest
Running org.jenkinsci.plugins.durabletask.BourneShellScriptTest
Running org.jenkinsci.plugins.durabletask.executors.ContinuedTaskTest
Running org.jenkinsci.plugins.durabletask.PowershellScriptTest
Running org.jenkinsci.plugins.durabletask.WindowsBatchScriptTest
Tests run: 1, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 60.195 sec - in org.jenkinsci.plugins.durabletask.executors.ContinuedTaskTest
Tests run: 8, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 61.245 sec - in InjectedTest
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 64.606 sec - in org.jenkinsci.plugins.durabletask.WindowsBatchScriptTest
Tests run: 6, Failures: 0, Errors: 0, Skipped: 6, Time elapsed: 64.96 sec - in org.jenkinsci.plugins.durabletask.BourneShellScriptTest
Tests run: 6, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 65.532 sec - in org.jenkinsci.plugins.durabletask.PowershellScriptTest

Results :

Tests run: 27, Failures: 0, Errors: 0, Skipped: 6

$PSVersionTable
Name                           Value
----                           -----
PSVersion                      5.1.15063.138
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.15063.138
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

gabloe:
I'll do a little research for an alternative to using $PSDefaultParameterValues

slide:
I'm using Windows 10 and have the following and the tests fail for me too.

```
Name                           Value
----                           -----
PSVersion                      5.0.10586.672
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.10586.672
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1
```

gabloe:
Ok, I'll investigate why this could be happening.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
slide: For the output (non-error) you could just pipe to "Out-File -Encoding UTF8" instead of redirecting directly to a file, but the error output would still have the UTF-16 though.
Something like (untested)
```
String.format("& \"%s\" *>1 2> \"%s\" | Out-File -Encoding \"UTF-8\" \"%s\"; [io.file]::WriteAllText(\"%s\",$LastExitCode); [io.file]::AppendAllText(\"%s\",$error);", 
```

slide:
I still get errors, they look different now though.

```
Tests run: 30, Failures: 0, Errors: 30, Skipped: 0, Time elapsed: 262.379 sec <<< FAILURE! - in org.jenkinsci.plugins.d
rabletask.PowershellScriptTest
implicitExit(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 93.41 sec  <<< ERROR!
java.io.IOException: corrupted content in C:\Users\USER\AppData\Local\Temp\hudson796485386413990360test\ws@tmp\durabl
-df7fc94f\jenkins-result.txt: java.lang.NumberFormatException: For input string: " ■0"
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.implicitExit(PowershellScriptTest.java:90)
Caused by: java.lang.NumberFormatException: For input string: " ■0"
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.implicitExit(PowershellScriptTest.java:90)
```

slide:
This worked for me locally:

```
 String cmd;
        if (capturingOutput) {
            // Using redirection in PowerShell produces extra newlines in output, so I am using [io.file]::WriteAllText to prevent corrupted output of exit code
            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"]=\"UTF8\"; & \"%s\" *>&1 2> \"%s\" | Out-File -Encoding ASCII -FilePath \"%s\" ; Out-File -Encoding ASCII -FilePath \"%s\" -InputObject $LastExitCode ; Out-File -Encoding ASCII -FilePath \"%s\" -InputObject $error;",
                quote(c.getPowershellMainFile(ws)),
                quote(c.getLogFile(ws)),
                quote(c.getOutputFile(ws)),
                quote(c.getResultFile(ws)),
                quote(c.getLogFile(ws)));
        } else {
            cmd = String.format("$PSDefaultParameterValues[\"*:Encoding\"]=\"UTF8\"; & \"%s\" *>&1 | Out-File -Encoding ASCII -FilePath \"%s\"; Out-File -Encoding ASCII -FilePath \"%s\" -InputObject $LastExitCode;",
                quote(c.getPowershellMainFile(ws)),
                quote(c.getLogFile(ws)),
                quote(c.getResultFile(ws)));
        }
```

gabloe:
@slide your change didn't work for me.  All of the powershell tests failed.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
gabloe: It seems like it does :)

slide:
I just pulled and I am still failing.

```
org.jenkinsci.plugins.durabletask.PowershellScriptTest.verbose(org.jenkinsci.plugins.durabletask.PowershellScriptTest)
  Run 1: PowershellScriptTest.verbose:138
Expected: a string containing "Hello, World!"
     but: was "ÿ?H e l l o ,   W o r l d !
```

gabloe:
I'm kind of just shooting in the dark here since I can't reproduce the errors you're seeing :(

slide:
Maybe I can setup some vm's of various windows and powershell versions and create a table of what works and what doesn't with the current code. I'm pretty sure the defaults stuff is fairly new and won't work on many powershell versions. It specifically only works for redirection on very recent versions.

jglick:
> I believe that this requires powershell 3.0 or higher

I have no opinion, but whatever you decide please

* document the platform requirements in `help.html`
* ensure that the test `Assume`s it is using a sufficiently new version
* `git merge origin/master` so the new `Jenkinsfile` will start testing your patch on the Azure agents on ci.jenkins.io

gabloe:
@jglick Great!  I will make these changes and take advantage of the new Jenkins.io azure agents for testing

gabloe:
@jglick @slide I have replaced instanced in which I was using $PSDefaultParameterValues to simply use Out-File directly and pass the correct character encoding.  This worked locally in my testing, and has passed the automated tests as well.

jakauppila:
@gabloe Do you have the ability to capture the Powershell output separately? For example, if I have a script that utilizes `Write-Verbose` and `Write-Output`, I would want the ability to write the `Write-Verbose` to the console, but only capture stdout of `Write-Output` into a variable within the pipeline to use later on.

**Powershell Script:**
```
function Get-MyOutput
{
    [CmdletBinding()]
    Param()

    Write-Verbose 'My verbose message'
    
    Write-Output 'My Actual Message'
}

$Output = Get-MyOutput -Verbose

Write-Warning "Contents of my output variable:"
$Output
```
**Execution Output:**
```
VERBOSE: My verbose message
WARNING: Contents of my output variable:
My Actual Message
```

**Pipeline definition:**
```
powershellScriptOutput = powershell(returnStdout: true, script: 'myScript.ps1')
echo powershellScriptOutput
```

**Desired Output:**
`My Actual Message`

**Nondesired Output:**
```
VERBOSE: My verbose message
WARNING: Contents of my output variable:
My Actual Message
```

slide:
@jakauppila maybe that can be a future enhancement?

gabloe:
@slide @jakauppila By default the console output will contain all streams that are enabled, unless captureOutput is set to true, which causes the console output to contain all enabled streams except for stderr, which gets redirected to the log file.

jakauppila:
@slide @gabloe I'd be fine with it being a future enhancement, just trying to think about what the configuration options would look like so that it isn't too confusing and will provide backwards compatibility of configurations.

**Powershell Streams**
- 1 Output/Success
- 2 Error
- 3 Warning
- 4 Verbose
- 5 Debug

So if we add a feature later `returnStdout` that returns a step value of the stdout stream, but continues to write the other streams to the console, that would satisfy what I'm looking for.

From the current `bat` documentation:
> returnStdout (optional)
>     If checked, standard output from the task is returned as the step value as a String, rather than being printed to the build log. (Standard error, if any, will still be printed to the log.) You will often want to call .trim() on the result to strip off a trailing newline.

gabloe:
@jglick I think with the latest commit I have added all changes you asked for.  Let me know if you think any additional changes are necessary :)

gabloe:
@jakauppila I think the best we can try to do here is redirect streams 2-6 to the log file while keeping only stdout (stream 1) in the output file.  Right now all streams are written to the output file except stderr.  I'll see if this is something that can be easily done.

gabloe:
@jakauppila with this latest commit your use case is handled now.  If returnStdout is set to true then all other streams will be redirected to the console log and only stdout will be captured separately and returned.

jakauppila:
@gabloe Excellent, thank you!

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,129 @@
...
+
+    @DataBoundConstructor public PowershellScript(String script) {
+        this.script = script;
+    }
+    
+    public String getScript() {
+        return script;
+    }
+
+    @Override public void captureOutput() {
```
jakauppila: Should this be `returnStdout` rather than `captureOutput` just to match the option for `bat`?

gabloe: Nope, returnStdout is exposed to the user via workflow-durable-task-step, but all that does is call captureOutput()

gabloe: So the behavior is the same.  You will use returnStdout :)

gabloe:
One of the Linux environments in ci.jenkins.io is flaking out on the bourneShellScriptTest.  But all other environments are passing.  I don't think it has anything to do with my PR.

slide:
@jglick You ok with this now?

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
gabloe: :)

i386:
For bonus points it would be amazing to see an update of https://jenkins.io/doc/pipeline/tour/running-multiple-steps/ to include PowerShell - get the word out 💃

gabloe:
@i386 That would be fantastic

westlakem:
any updates on this? I would love to use PowerShell, especially since cmd (bat) is starting to be deprecated.
http://news.softpedia.com/news/microsoft-replaces-command-prompt-with-powershell-in-latest-windows-10-build-510328.shtml

gabloe:
@westlakem The work is finished from my perspective

slide:
I think we are just waiting on @jglick to approve the changes and merge the PR.

gabloe:
I've pushed a small fix to prevent the output file (if returnStdout is used) from containing a BOM.  Otherwise output will have some erroneous characters at the beginning (ï»¿)

jakauppila:
@jglick Any movement on this? I'd love to test it out.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -0,0 +1,173 @@
...
+
+    @Before public void vars() throws IOException, InterruptedException {
+        listener = StreamTaskListener.fromStdout();
+        ws = j.jenkins.getRootPath().child("ws");
+        launcher = j.jenkins.createLauncher(listener);
+        Properties properties = System.getProperties();
+        String pathSeparator = properties.getProperty("path.separator");
+        String[] paths = System.getenv("PATH").split(pathSeparator);
+        boolean powershellExists = false;
+        String cmd = launcher.isUnix()?"powershell":"powershell.exe";
```
jglick: Note to self (this is probably too involved in ci.jenkins.io infra details for most contributors to take on): would be useful to patch `Jenkinsfile` to require `platforms: ['docker', 'windows']` and then make this test when run on Linux use `docker-fixtures` to launch a container which somehow mixes `JavaContainer` with `microsoft/powershell`, for example by adding [these installation instructions](https://github.com/PowerShell/PowerShell/blob/master/docs/installation/linux.md#ubuntu-1604). Would allow Powershell for Linux to be tested on CI.

src/main/resources/org/jenkinsci/plugins/durabletask/Messages.properties
```diff
@@ -1,2 +1,3 @@
 BourneShellScript.bourne_shell=Bourne shell
 WindowsBatchScript.windows_batch=Windows batch
+PowershellScript.powershell=Powershell script
```
jglick: **PowerShell** acc. to https://msdn.microsoft.com/powershell

(though this UI is not generally shown anyway)

src/main/resources/org/jenkinsci/plugins/durabletask/PowershellScript/help-script.html
```diff
@@ -0,0 +1,3 @@
+<div>
+    A Powershell Script.  Tested on Powershell 3.x+
```
jglick: ditto

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -0,0 +1,173 @@
...
+        if (powershellExists == true) {
+            List<String> args = new ArrayList<String>();
+            args.addAll(Arrays.asList(cmd,"$PSVersionTable.PSVersion.Major"));
+            Launcher.ProcStarter ps = launcher.launch().cmds(args).quiet(true);
+            ps.readStdout();
+            Proc proc = ps.start();
+            byte[] buffer = new byte[5];
+            while (proc.getStdout().read(buffer) > 0) {}
+            String psVersionStr = new String(buffer);
+            int psVersion = Integer.parseInt(psVersionStr.trim());
```
jglick: Replace these last four lines with:

```java
String psVersionStr = IOUtils.toString(proc.getStdout());
int psVersion;
try {
    psVersion = Integer.parseInt(psVersionStr.trim());
} catch (NumberFormatException x) {
    throw new AssumptionViolatedException(psVersionStr, x);
}
```

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -0,0 +1,173 @@
...
+                powershellExists = true;
+                break;
+            }
+        }
+        Assume.assumeTrue("This test should only run if powershell is available", powershellExists == true);
+        
+        // Assume Powershell major version is at least 3
+        if (powershellExists == true) {
+            List<String> args = new ArrayList<String>();
+            args.addAll(Arrays.asList(cmd,"$PSVersionTable.PSVersion.Major"));
```
jglick: Replace these last two lines with:

```java
List<String> args = new ArrayList<>(Collections.singleton(cmd));
if (!launcher.isUnix()) {
    args.addAll(Arrays.asList("-ExecutionPolicy", "Bypass"));
}
args.add("$PSVersionTable.PSVersion.Major");
```

jglick: BTW

```bash
$ docker run -it microsoft/powershell powershell -ExecutionPolicy Bypass '$PSVersionTable.PSVersion.Major'
6
```

so I am not sure why we need to exclude this argument on Linux.

jglick:
After those fixes, I still get

```
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 93.14 sec <<< FAILURE! - in org.jenkinsci.plugins.durabletask.PowershellScriptTest
verbose(org.jenkinsci.plugins.durabletask.PowershellScriptTest)  Time elapsed: 10.328 sec  <<< ERROR!
java.nio.file.FileSystemException:
C:\Users\…\AppData\Local\Temp\jenkinsTests.tmp\jenkins…test\ws@tmp\durable-…\temporaryOutput.txt: The process cannot access the file because it is being used by another process.

        at sun.nio.fs.WindowsException.translateToIOException(WindowsException.java:86)
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:97)
        at sun.nio.fs.WindowsException.rethrowAsIOException(WindowsException.java:102)
        at sun.nio.fs.WindowsFileSystemProvider.implDelete(WindowsFileSystemProvider.java:269)
        at sun.nio.fs.AbstractFileSystemProvider.delete(AbstractFileSystemProvider.java:103)
        at java.nio.file.Files.delete(Files.java:1126)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at hudson.Util.deleteFile(Util.java:255)
        at hudson.FilePath.deleteRecursive(FilePath.java:1203)
        at hudson.FilePath.deleteContentsRecursive(FilePath.java:1212)
        at hudson.FilePath.deleteRecursive(FilePath.java:1194)
        at hudson.FilePath.access$1000(FilePath.java:190)
        at hudson.FilePath$14.invoke(FilePath.java:1173)
        at hudson.FilePath$14.invoke(FilePath.java:1170)
        at hudson.FilePath.act(FilePath.java:990)
        at hudson.FilePath.act(FilePath.java:968)
        at hudson.FilePath.deleteRecursive(FilePath.java:1170)
        at org.jenkinsci.plugins.durabletask.FileMonitoringTask$FileMonitoringController.cleanup(FileMonitoringTask.java:187)
        at org.jenkinsci.plugins.durabletask.PowershellScriptTest.verbose(PowershellScriptTest.java:165)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.internal.runners.statements.RunBefores.evaluate(RunBefores.java:26)
        at org.jvnet.hudson.test.JenkinsRule$2.evaluate(JenkinsRule.java:534)
        at org.junit.rules.RunRules.evaluate(RunRules.java:20)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:325)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:78)
        at org.junit.runners.BlockJUnit4ClassRunner.runChild(BlockJUnit4ClassRunner.java:57)
        at org.junit.runners.ParentRunner$3.run(ParentRunner.java:290)
        at org.junit.runners.ParentRunner$1.schedule(ParentRunner.java:71)
        at org.junit.runners.ParentRunner.runChildren(ParentRunner.java:288)
        at org.junit.runners.ParentRunner.access$000(ParentRunner.java:58)
        at org.junit.runners.ParentRunner$2.evaluate(ParentRunner.java:268)
        at org.junit.runners.ParentRunner.run(ParentRunner.java:363)
        at org.apache.maven.surefire.junit4.JUnit4Provider.execute(JUnit4Provider.java:367)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeWithRerun(JUnit4Provider.java:274)
        at org.apache.maven.surefire.junit4.JUnit4Provider.executeTestSet(JUnit4Provider.java:238)
        at org.apache.maven.surefire.junit4.JUnit4Provider.invoke(JUnit4Provider.java:161)
        at org.apache.maven.surefire.booter.ForkedBooter.invokeProviderInSameClassLoader(ForkedBooter.java:290)
        at org.apache.maven.surefire.booter.ForkedBooter.runSuitesInProcess(ForkedBooter.java:242)
        at org.apache.maven.surefire.booter.ForkedBooter.main(ForkedBooter.java:121)


Results :

Tests in error:
  PowershellScriptTest.verbose:165 ╗ FileSystem C:\Users\…\AppData\Local\Temp...

Tests run: 6, Failures: 0, Errors: 1, Skipped: 0
```

jglick:
(though not reproducibly)

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
gabloe: ExecutionPolicy is not a thing in PowerShell for Linux. It's reported as a bug in PowerShells GitHub issue tracker, but it's not going to be fixed because it's not expected to have any impact on Linux. Most likely the argument is ignored, but I think it should be omitted as a sanity measure. https://github.com/PowerShell/PowerShell/issues/478

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -155,6 +164,7 @@
         c.writeLog(ws, baos);
         assertEquals(0, c.exitStatus(ws, launcher).intValue());
         assertThat(baos.toString(), containsString("Hello, World!"));
+        c.stop(ws,launcher);
```
jglick: Why would this be necessary? The test failure seemed to indicate a file handle leak as far as I can tell, and the actual `DurableTaskStep` would not normally be calling `stop` here (only if you deliberately abort the build).

gabloe:
This is my mistake.  I was testing this and meant to delete that line before committing.  I'll remove it.

jglick:
Last test run (with `stop` call removed) passed for me, so I am not sure what the issue with `temporaryOutput.txt` is.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -0,0 +1,140 @@
...
+                quote(c.getResultFile(ws)));
+        }
+
+        // Write the script and execution wrapper to powershell files in the workspace
+        c.getPowershellScriptFile(ws).write(script, "UTF-8");
+        c.getPowershellMainFile(ws).write("try {\r\n& '" + quote(c.getPowershellScriptFile(ws)) + "'\r\n} catch {\r\nWrite-Error $_; exit 1;\r\n}\r\nfinally {\r\nif ($LastExitCode -ne $null) {\r\nexit $LastExitCode;\r\n} elseif ($error.Count -gt 0 -or !$?) {\r\nexit 1;\r\n} else {\r\nexit 0;\r\n}\r\n}", "UTF-8");
+        c.getPowershellWrapperFile(ws).write(cmd, "UTF-8");
+
+        if (launcher.isUnix()) {
+            // Open-Powershell does not support ExecutionPolicy
```
jtnord: no but powershell in bash in ubuntu in windows in creators update does :-)

PR #40, opened 2017-04-05
[JENKINS-39307] fixed missing EnvVars in ps command

kufi:
This PR adds the env vars from the shell command to the corresponding ps command.

This fixes the following problem: https://issues.jenkins-ci.org/browse/JENKINS-39307

The cause for the unexpected abort is the following:

When starting a docker container on a different docker-Host, set by the env-var DOCKER_HOST, the sh commands get correctly piped through to the container, but the ps command will not go to the correct docker host, because the env vars are not set.

In our setup we run Jenkins in a docker container and use the same docker-host to run the pipeline docker containers. This means that we have to explicitely set the DOCKER_HOST to something different than localhost.

To reproduce, run the following script, but set the DOCKER_HOST on the node to an external Host. The "sleep 10" command will be aborted and the echo will never run.

```
pipeline {
    agent {
        docker {
            image 'ubuntu:16.04'
        }
    }
    stages {
        stage('Run') {
            steps {
                sh 'sleep 10'
                sh 'echo "Test"'
        }
    }
}
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -39,6 +39,7 @@ import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.WeakHashMap;
+import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.model.Jenkins;
 import jenkins.security.MasterToSlaveCallable;
@@ -162,7 +163,9 @@ public final class BourneShellScript extends FileMonitoringTask {
 
         private int pid;
         private final long startTime = System.currentTimeMillis();
-        private EnvVars envVars = new EnvVars();
+
+        @CheckForNull
+        private EnvVars envVars;
 
         private ShellController(FilePath ws, EnvVars envVars) throws IOException, InterruptedException {
             super(ws);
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -84,7 +84,7 @@ protected FileMonitoringController doLaunch(FilePath workspace, Launcher launche
     /**
      * JENKINS-40734: blocks the substitutions of {@link EnvVars#overrideExpandingAll} done by {@link Launcher}.
      */
-    protected static Map<String, String> escape(EnvVars envVars) {
+    static Map<String, String> escape(EnvVars envVars) {
```
jglick: Unnecessary (and incompatible); revert.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -162,9 +162,11 @@ public String getScript() {
 
         private int pid;
         private final long startTime = System.currentTimeMillis();
+        private EnvVars envVars;
```
jglick: This is going to be null after deserialization of a `sh` step in a build that started running prior to plugin upgrade, so you must handle that.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -62,7 +62,7 @@
     @DataBoundConstructor public BourneShellScript(String script) {
         this.script = Util.fixNull(script);
     }
-    
+
```
jglick: As a general rule, avoid whitespace changes in PRs so as to keep the diff minimal.

kufi:
Ok. I fixed the problems.

I'm not really sure how to test this automatically as a unit test, as it could only be reproduced if the test would actually run in the described setup.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -162,9 +162,11 @@ public String getScript() {
 
         private int pid;
         private final long startTime = System.currentTimeMillis();
+        private EnvVars envVars = new EnvVars();
```
jglick: This initializer is useless—it is always overridden by the constructor.

You can set fields in restored objects using a `readResolve` method, but it would be simpler to just mark this field `@CheckForNull` as a reminder that it will occasionally be unset, and have `exitStatus` behave accordingly.

jglick:
Yeah I think this just needs a minor technical fix.

kufi:
Fixed the null value check.

jglick:
Some positive results claimed in [JENKINS-35370](https://issues.jenkins-ci.org/browse/JENKINS-35370) also.

hvisser:
FWIW, this fixed JENKINS-35370 for me too.

jglick:
No longer necessary.

PR #41, opened 2017-04-25
Try a Jenkinsfile

jglick:
Pending [INFRA-1168](https://issues.jenkins-ci.org/browse/INFRA-1168) and a successful build. Would be very helpful for Windows-specific changes like #39.

@reviewbybees

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,1 @@
+buildPlugin(jenkinsVersions: [null, '2.32.3'])
--- a/pom.xml
+++ b/pom.xml
@@ -3,7 +3,8 @@
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
-        <version>2.7</version>
+        <version>2.26</version>
+        <relativePath />
     </parent>
     <artifactId>durable-task</artifactId>
     <version>1.14-SNAPSHOT</version>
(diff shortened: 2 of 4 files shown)


Jenkinsfile
```diff
@@ -0,0 +1 @@
+buildPlugin(jenkinsVersions: [null, '2.32.3'])
```
oleg-nenashev: Didn't know `null` works here, but it is really useful in such context

pom.xml
```diff
@@ -14,26 +15,25 @@
...
             <name>MIT License</name>
-            <url>http://www.opensource.org/licenses/mit-license.php</url>
+            <url>https://opensource.org/licenses/MIT</url>
         </license>
     </licenses>
     
     <properties>
-      <jenkins.version>1.609.1</jenkins.version>
-      <java.level>6</java.level>
+      <jenkins.version>1.642.3</jenkins.version>
```
oleg-nenashev: Not sure why you would need it, but OK since you are a maintainer.
1.625 would be enough for try-with-resources

jglick:
> Not sure why you would need it

```
[INFO] --- maven-enforcer-plugin:1.3.1:enforce (display-info) @ durable-task ---
[INFO] Restricted to JDK 1.6 yet org.codehaus.groovy:groovy-all:jar:2.4.7:provided contains org/codehaus/groovy/runtime/NioGroovyMethods.class targeted to JDK 1.7
[INFO] Restricted to JDK 1.6 yet org.jenkins-ci.main:remoting:jar:3.4.1:provided contains org/jenkinsci/remoting/RoleChecker.class targeted to JDK 1.7
[WARNING] Rule 2: org.apache.maven.plugins.enforcer.EnforceBytecodeVersion failed with message:
Found Banned Dependency: org.codehaus.groovy:groovy-all:jar:2.4.7
Found Banned Dependency: org.jenkins-ci.main:remoting:jar:3.4.1
Use 'mvn dependency:tree' to locate the source of the banned dependencies.
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
```

`java.level=6` and `jenkins.version=2.32.3` do not mix. Probably the library could be patched to automatically set a newer `java.level` when overriding `jenkins.version`, but this seemed like more hassle than it was worth for now. Anyway Pipeline plugins use 1.642.x and most changes in this plugin are delivering Pipeline features so it seemed silly to use such an ancient baseline.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -173,7 +174,9 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
 
         @Override public byte[] getOutput(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
             // TODO could perhaps be more efficient for large files to send a MasterToSlaveFileCallable<byte[]>
-            return IOUtils.toByteArray(getOutputFile(workspace).read());
+            try (InputStream is = getOutputFile(workspace).read()) {
```
jglick: I think this fixed [JENKINS-43639](https://issues.jenkins-ci.org/browse/JENKINS-43639).

PR #42, opened 2017-05-29
Revert "Trying JNA instead."

bgamari:
This reverts commit 3c10cc0cb0a4738646966494fbeaab54bca115e1.

JNA breaks on Windows under Cygwin as there is no shared libc available. Moreover, I believe this is no longer necessary; I have no trouble using this plugin with current JNR and FreeBSD-11.0.

Fixes JENKINS-38682.


bgamari:
Never mind, unfortunately this still doesn't fix the issue as `getpgid` is unimplemented by `jnr-posix` on Windows..

bgamari:
Never mind, unfortunately this still doesn't fix the issue as `getpgid` is unimplemented by `jnr-posix` on Windows..

PR #43, opened 2017-05-30
[JENKINS-38682] - Use waitpid to poll for process liveness

bgamari:
This reverts commit 3c10cc0cb0a4738646966494fbeaab54bca115e1 which
moved the process liveness checking implementation from JNR to JNA.
This broke on Windows as there is no shared libc available.

Instead, we move back to JNR, but using waitpid instead of getpgid to
perform the polling. This is a generic enough interfacec that it should
have very consistent support across POSIX operating systems and can even
be supported on Windows (in a future JNR release).

Finally, I believe the original motivation cited for moving to JNA in
3c10cc0 no longer applies; I have no trouble using this plugin with
current JNR and FreeBSD-11.0.

Fixes [JENKINS-38682](https://issues.jenkins-ci.org/browse/JENKINS-38682).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1,5 +1,20 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-constants</artifactId>
+            <version>0.9.10-SNAPSHOT</version>
+        </dependency>
+        <!-- Commented out until we can assume new jnr-posix >= 3.0.42 is in
+             core due to the need for waitpid on Windows.
+        <dependency>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-posix</artifactId>
+            <version>3.0.42-SNAPSHOT</version>
+        </dependency>
+        -->
+    </dependencies>
     <parent>
         <groupId>org.jenkins-ci.plugins</groupId>
         <artifactId>plugin</artifactId>
--- a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
@@ -99,12 +99,21 @@ final class ProcessLiveness {
         }
     }
 
+    /**
+     * Worker-side implementation of process liveness check.
+     *
+     * @throws IllegalStateException when we are unable to find a functional
+     * process-liveness checking mechanism.
+     */
     private static final class Liveness extends MasterToSlaveCallable<Boolean,RuntimeException> {
         private final int pid;
         Liveness(int pid) {
             this.pid = pid;
         }
-        @Override public Boolean call() throws RuntimeException {
+        @Override public Boolean call() throws IllegalStateException {
+            final String fallback_msg = "jnr doesn't have waitpid support; "+
+                "falling back to the JNA LibC implementation";
+
             // Try jnr's waitpid
             // We need to painstakingly extract the function via reflection to
             // avoid class-loading errors in the event that the available jnr
@@ -113,6 +122,10 @@ final class ProcessLiveness {
             // This silliness is equivalent to,
             //     int[] code = new int[1];
             //     return jnr.waitpid(pid, code, WaitFlags.WNOHANG.intValue()) != -1;
+            //
+            // TODO: Replace by the direct call once Jenkins core is updated to 2.64+
+            // which has https://github.com/jenkinsci/jenkins/pull/2904.
+
             try {
                 ClassLoader loader = getClass().getClassLoader();
                 Object jnr = PosixAPI.jnr();
@@ -130,13 +143,13 @@ final class ProcessLiveness {
                 Integer ret = (Integer) waitpidMethod.invoke(jnr, pid, code, WNOHANG);
                 return ret != -1;
             } catch (InvocationTargetException x) {
-                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+                LOGGER.log(Level.WARNING, fallback_msg, x);
             } catch (IllegalAccessException x) {
-                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+                LOGGER.log(Level.WARNING, fallback_msg, x);
             } catch (ClassNotFoundException x) {
-                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+                LOGGER.log(Level.WARNING, fallback_msg, x);
             } catch (NoSuchMethodException x) {
-                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+                LOGGER.log(Level.WARNING, fallback_msg, x);
             }
 
             // JNR-POSIX used to fail on FreeBSD at least, so try JNA as well.
--- a/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
@@ -28,8 +28,12 @@ import com.sun.jna.Library;
 import com.sun.jna.Native;
 import hudson.Launcher;
 import hudson.model.TaskListener;
+import hudson.os.PosixAPI;
+import hudson.os.PosixException;
 import hudson.remoting.VirtualChannel;
 import java.io.IOException;
+import java.lang.reflect.Method;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Collections;
 import java.util.Map;
 import java.util.WeakHashMap;
@@ -37,6 +41,11 @@ import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.security.MasterToSlaveCallable;
+import jnr.constants.platform.WaitFlags;
+import jnr.posix.POSIX;
+import jnr.posix.POSIXFactory;
+import jnr.posix.POSIXHandler;
+import jnr.posix.util.DefaultPOSIXHandler;
 
 /**
  * Utility class to track whether a given process is still alive.
@@ -76,13 +85,13 @@ final class ProcessLiveness {
     private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launcher) throws IOException, InterruptedException {
         if (launcher instanceof Launcher.LocalLauncher || launcher instanceof Launcher.RemoteLauncher) {
             try {
-                boolean alive = channel.call(new Liveness(pid));
-                LOGGER.log(Level.FINER, "{0} is alive? {1}", new Object[] {pid, alive});
-                return alive;
-            } catch (RuntimeException x) {
+                return channel.call(new Liveness(pid));
+            } catch (IllegalStateException x) {
                 LOGGER.log(Level.WARNING, "cannot determine liveness of " + pid, x);
-                return true;
+            } catch (PosixException x) {
+                LOGGER.log(Level.WARNING, "cannot determine liveness of " + pid, x);
             }
+            return true;
         } else {
             // Using a special launcher; let it decide how to do this.
             // TODO perhaps this should be a method in Launcher, with the following fallback in DecoratedLauncher:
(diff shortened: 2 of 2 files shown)


bgamari:
The bits enabling Windows support can be found in https://ci.appveyor.com/project/jnr/jnr-posix/build/198/job/jmxp6unc18afs32r.

bgamari:
Ping?

bgamari:
Pinging @reviewbybees

slide:
@jenkinsci/code-reviewers

pom.xml
```diff
@@ -1,5 +1,17 @@
...
+    <dependencies>
+        <dependency>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-constants</artifactId>
+            <version>0.9.9</version>
+        </dependency>
+        <dependency>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-posix</artifactId>
+            <version>3.0.41</version>
```
oleg-nenashev: It won't work reliably in such way since Jenkins core may push its own lib to the classloader.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -89,33 +95,26 @@ private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launch
...
-            // JNR-POSIX does not seem to work on FreeBSD at least, so using JNA instead.
-            LibC libc = LibC.INSTANCE;
-            if (libc.getpgid(0) == -1) {
-                throw new IllegalStateException("getpgid does not seem to work on this platform");
+            try { // loadPOSIX sends problems to stderr, useless if we are on a slave
+                Method loadNativePOSIX = POSIXFactory.class.getDeclaredMethod("loadNativePOSIX", POSIXHandler.class);
+                loadNativePOSIX.setAccessible(true);
+                loadNativePOSIX.invoke(null, new DefaultPOSIXHandler());
+            } catch (NoSuchMethodException x) {
+                LOGGER.log(Level.WARNING, "loadNativePOSIX no longer exists?", x);
```
oleg-nenashev: It worth trying LibC in this case. If you wrap it by `isLinux()` check and proper logging in `else`, it should be safe enough

jglick:
Without an automated (or even manual) test environment for Cygwin I cannot really evaluate PRs like these. Generally I would consider `sh` simply unsupported on Windows—you should rather use `bat`, or (when finished) `ps`.

As noted in [JENKINS-25503](https://issues.jenkins-ci.org/browse/JENKINS-25503), a more reliable solution would probably be to use a statically linked helper application cross-compiled for various platforms.

pom.xml
bgamari: I'm a bit confused; this is simply reverting to the approached used prior to 3c10cc0; wouldn't this argument had held prior to this 3c10cc0 as well? I'm admittedly rather unfamiliar with the Jenkins class loading scheme; under what circumstances would the version pushed to the builder differ from version required here?

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
bgamari: Sure, I'd be happy to do that.

bgamari:
> Without an automated (or even manual) test environment for Cygwin I cannot really evaluate PRs like these. Generally I would consider sh simply unsupported on Windows—you should rather use bat, or (when finished) ps.

To provide some context, this work was done as part of an on-going effort to move the continuous integration infrastructure of the [Glasgow Haskell Compiler](http://haskell.org/ghc) from Phabricator's Harbormaster to Jenkins.

GHC is a compiler which targets a multitude of platforms and has reasonably complex testing requirements. While it is technically feasible to reimplement this test logic for Windows, needing to do so would be a regression from Harbormaster.

Consequently, I would be happy to perform whatever testing necessary to address your concerns regarding this patch.

Alternatively, I would be happy to implement the scheme suggested in JENKINS-25503 if it would be easier to accept. In the ticket you suggest Go for implementing the wrapper; would you accept C and C++ instead? Not only am I more familiar with these languages but I suspect compilers are significantly easier to come by. Also, I would likely want to consult with you for guidance on integrating this scheme into the Jenkins build and deployment system as my knowledge of Java packaging practices is quite limited.

pom.xml
oleg-nenashev: So there is an issue with that...

1) The ProcessLiveness code is being executed on Agent
2) Agent loads only one version of library to its JAR cache
3) Depending on the call, it may download an old 3.0.1 version from the core (before your patch) or the new one in the plugin

OTOH the plugin library is a priority for core side IIRC. But I would not really rely on availability of classes from 3.0.41. So I would use reflection to retrieve the stuff in `ProcessLiveness` implementation and comment out this dependency

bgamari:
So what is the status of this? It's quite unclear to me whether @jglick's comment was intended to mean that the patch as-written is unacceptable or merely difficult for upstream to test.

oleg-nenashev:
@bgamari Patch as written is inacceptable to me (see the comment above), but IMHO it requires only small changes to become acceptable. 

From what I see from @jglick's comment, as a maintainer he just does not commit to keep other changes compatible with `sh()` on Windows since he does not test it now. But I'd guess he will be fine with any changes while they do not cause regressions on Unix systems.

Otherwise this regression fix may cause other regressions. I will perform manual testing once my comments are addressed.

oleg-nenashev:
@bgamari So do you plan to work on it? If no, I could probably dedicate some time next week

bgamari:
Oleg Nenashev <notifications@github.com> writes:

> @bgamari So do you plan to work on it? If no, I could probably dedicate some time next week
>
I do but I just returned from a camping trip, so I'm catching up on more
urgent matters at the moment. I can try to get to it this weekend.

bgamari:
@oleg-nenashev how does this look?

oleg-nenashev:
@bgamari Generally looks good to me, but IMHO `Level.WARNING` is too high. Jenkins will be frequently checking statuses, hence the logs will be spammed. Maybe FINE or FINEST is better

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -94,14 +103,48 @@ private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launch
...
+            } catch (IllegalAccessException x) {
+                LOGGER.log(Level.WARNING, "jnr doesn't have waitpid support", x);
+            } catch (ClassNotFoundException x) {
+                LOGGER.log(Level.WARNING, "jnr doesn't have waitpid support", x);
+            } catch (NoSuchMethodException x) {
+                LOGGER.log(Level.WARNING, "jnr doesn't have waitpid support", x);
+            }
+
+            // JNR-POSIX used to fail on FreeBSD at least, so try JNA as well.
             LibC libc = LibC.INSTANCE;
```
oleg-nenashev: Here you can explicitly check if the instance is a Windows one and skip this step then.Or...  Any chance JVM is successful to call libc.so on Windows with Cygwin/MinGW/whatever?

bgamari:
@oleg-nenashev, how does this look now?

rtyler:
@gabloe This pull request may also be of interest to you since @bgamari has been working to make Jenkins and Windows play more nicely together :smile_cat:

oleg-nenashev:
@bgamari please do not squash commits while the review is in progress. Especially after a 1-month delay. It makes reviewers to review everything from scratch.

If we need to squash PRs, it can be done during the merge operation.

pom.xml
```diff
@@ -1,5 +1,20 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
     <modelVersion>4.0.0</modelVersion>
+    <dependencies>
+        <dependency>
+            <groupId>com.github.jnr</groupId>
+            <artifactId>jnr-constants</artifactId>
+            <version>0.9.10-SNAPSHOT</version>
```
oleg-nenashev: Any release ETA?

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -94,14 +103,49 @@ private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launch
...
+                return ret != -1;
+            } catch (InvocationTargetException x) {
+                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+            } catch (IllegalAccessException x) {
+                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+            } catch (ClassNotFoundException x) {
+                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+            } catch (NoSuchMethodException x) {
+                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
+            }
```
oleg-nenashev: maybe also SecurityException, but it's not something going to work well in Jenkins anyway

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -94,14 +103,49 @@ private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launch
...
+                Class WaitFlags = loader.loadClass("jnr.constants.platform.WaitFlags");
+                Method intValue = WaitFlags.getDeclaredMethod("intValue");
+                Integer WNOHANG = (Integer) intValue.invoke(Enum.valueOf(WaitFlags, "WNOHANG"));
+
+                int[] code = new int[1];
+                waitpidMethod.setAccessible(true); // the implementations are package-local
+                Integer ret = (Integer) waitpidMethod.invoke(jnr, pid, code, WNOHANG);
+                return ret != -1;
+            } catch (InvocationTargetException x) {
+                LOGGER.log(Level.FINE, "jnr doesn't have waitpid support", x);
```
oleg-nenashev: Add "falling back to the JNA LibC implementation" text to the log entry?

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -94,14 +103,49 @@ private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launch
             this.pid = pid;
         }
         @Override public Boolean call() throws RuntimeException {
-            // JNR-POSIX does not seem to work on FreeBSD at least, so using JNA instead.
+            // Try jnr's waitpid
```
oleg-nenashev: It makes sense to add `//TODO: Replace by the direct call once Jenkins core is updated to 2.64+ with https://github.com/jenkinsci/jenkins/pull/2904`

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -94,14 +103,49 @@ private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launch
             this.pid = pid;
         }
         @Override public Boolean call() throws RuntimeException {
```
oleg-nenashev: Would be great to document possible exceptions in the `Liveness` Javadoc

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
```diff
@@ -74,13 +83,13 @@ public static boolean isAlive(VirtualChannel channel, int pid, Launcher launcher
     private static boolean _isAlive(VirtualChannel channel, int pid, Launcher launcher) throws IOException, InterruptedException {
         if (launcher instanceof Launcher.LocalLauncher || launcher instanceof Launcher.RemoteLauncher) {
             try {
-                boolean alive = channel.call(new Liveness(pid));
-                LOGGER.log(Level.FINER, "{0} is alive? {1}", new Object[] {pid, alive});
-                return alive;
-            } catch (RuntimeException x) {
+                return channel.call(new Liveness(pid));
```
oleg-nenashev: You could probably check Jenkins core version here: `Jenkins.getVersion().isNewerThan(new VersionNumber("2.64"))` in order to optimize the call on the agent side. and to check JNR version before it . But it's rather polishing, which may actually break some use-cases with pre-loaded JAR files

pom.xml
bgamari: I'll ping upstream to find out.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
bgamari: Let's hold off on this.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
bgamari: Done.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
bgamari: Done.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
bgamari: Done.

src/main/java/org/jenkinsci/plugins/durabletask/ProcessLiveness.java
bgamari: I've not encountered this one but we're now catching it anyways.

bgamari:
How does this look now, @oleg-nenashev>

bgamari:
@oleg-nenashev, ping?

oleg-nenashev:
sorry, missed pings under tons of notifications

pom.xml
oleg-nenashev: Still a snapshot, not releasable

jglick:
No longer necessary.

bgamari:
@jglick, for posterity can you explain why this is no longer necessary?

bgamari:
Ahh, never mind; I see: #49.

PR #44, opened 2017-07-19
Fix issue with unicode chars in PowerShell

gabloe:
This PR fixes an issue that I discovered while using the PowerShell step that causes Unicode characters to not be interpreted properly.  I've also added a new test that checks for proper interpretation of Unicode chars.

Example:
`node {
   powershell 'Write-Output "Hellõ"'
}`

Before this fix:
> ﻿HellÃµ

After this fix:
> Hellõ

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,1 @@
+buildPlugin(jenkinsVersions: [null, '2.32.3'])
--- a/pom.xml
+++ b/pom.xml
@@ -37,15 +37,6 @@
         </pluginRepository>
     </pluginRepositories>
     
-    <dependencies>
-      <dependency>
-        <groupId>org.jenkins-ci.plugins</groupId>
-        <artifactId>powershell</artifactId>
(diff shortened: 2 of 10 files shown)


gabloe:
PowerShell by default adds a BOM when writing Unicode text out to a file.  This is a bug, and is meant to be fixed soon as per the RFC:
https://github.com/PowerShell/PowerShell-RFC/blob/master/2-Draft-Accepted/RFC0020-DefaultFileEncoding.md

This PR adds a workaround that better supports Unicode characters in Jenkins pipeline code.  The prior workaround only supported BOM-free output when using returnStdout, while this prevents a BOM in all cases.

rtyler:
@gabloe This pull request looks like it needs to be rebased against the `master` or moved to a different branch. The commit history is messy with a lot of stuff in your fork's master branch as far as I can tell.

gabloe:
@rtyler Good point.  The history does look super screwy.  I'll rebase against master.

PR #45, opened 2017-08-01
Fixing issues with unicode characters in PowerShell

gabloe:
This PR fixes an issue that I discovered while using the PowerShell step that causes Unicode characters to not be interpreted properly.  I've also added a new test that checks for proper interpretation of Unicode chars.

Example:
`node {
   powershell 'Write-Output "Hellõ"'
}`

Before this fix:
> ﻿HellÃµ

After this fix:
> Hellõ

PowerShell by default adds a BOM when writing Unicode text out to a file.  This is a bug, and is meant to be fixed soon as per the RFC:
https://github.com/PowerShell/PowerShell-RFC/blob/master/2-Draft-Accepted/RFC0020-DefaultFileEncoding.md

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -40,6 +40,8 @@ import java.io.IOException;
 import org.kohsuke.stapler.DataBoundConstructor;
 import java.io.FileInputStream;
 import java.io.InputStreamReader;
+import java.io.OutputStream;
+import java.nio.charset.Charset;
 
 /**
  * Runs a Powershell script
@@ -47,7 +49,7 @@ import java.io.InputStreamReader;
 public final class PowershellScript extends FileMonitoringTask {
     private final String script;
     private boolean capturingOutput;
-
+    
     @DataBoundConstructor public PowershellScript(String script) {
         this.script = script;
     }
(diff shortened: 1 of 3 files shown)


gabloe:
@rtyler This PR looks cleaner :)

gabloe:
@rtyler can I get a review of this PR?  Thanks!

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -104,26 +159,32 @@ public String getScript() {
...
-            return controlDir(ws).child("powershellMain.ps1");
-        }
         
-        public FilePath getPowershellScriptFile(FilePath ws) throws IOException, InterruptedException {
+        public FilePath getPowerShellScriptFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("powershellScript.ps1");
         }
         
-        public FilePath getPowershellWrapperFile(FilePath ws) throws IOException, InterruptedException {
+        public FilePath getPowerShellWrapperFile(FilePath ws) throws IOException, InterruptedException {
```
jglick: Please avoid gratuitous changes when filing a PR. It just makes the diff impossible to review effectively.

gabloe: Okay, I'll remember that in the future.  I think I just wanted to fix the casing to be consistent since in some places I use "Powershell" and in others "PowerShell".

svanoort:
I don't use Windows and can't write PowerShell, so can't really review that content well.  But nothing looks blatantly crazy and I see it includes test coverage and it passes existing Windows tests plus other reviews, so... approving and merging.

gabloe:
Thanks for the quick response :)

jglick:
Does this fix [JENKINS-46496](https://issues.jenkins-ci.org/browse/JENKINS-46496) by any chance?

There are a ton of open bugs in this plugin relating to the `powershell` step, are you watching them?

gabloe:
Yes, it does.  I have not actually seen these bugs.  Is there a way for me to be automatically notified when a bug is opened against powershell in pipeline?  Either way I will respond to these bugs.

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #46, opened 2017-08-24
Apply a reasonable timeout to ps commands

jglick:
Generally want to avoid `ProcStarter.join()` without a timeout, in case the process hangs for some reason.

@reviewbybees


jglick:
@reviewbybees done

oleg-nenashev:
The only thing I can imagine is an extremely long classloading. Bt even in such case it will fail and try again, right?

svanoort:
@scoheb  It's released as of 13 Oct

PR #48, opened 2017-09-21
fix: use correct syntax ps command that checks if process is alive.

iocanel:
Method isAlive internally uses `ps -o pid=<pid>`, to check if a pid exists or not.

I've tried running `ps -o pid=<pid>` in many different system and it always returns 0, regardless of the pid.

As a result when its called like: `ps -o pid=99999` it always return 0, leading eventually to `does not seem able to determine whether processes are alive or not`

I either miss something obvious here, or the command should be `ps -o pid -p <pid>` instead, as this pull request suggest.


jglick:
This is likely a more complicated version of #35, anyway. Do you use Cygwin, or what was the motivation for this change?

PR #47, opened 2017-09-11
Try to be robust against LinkageError’s from JNA

jglick:
An installation encountering https://github.com/kohsuke/akuma/issues/15 which would normally be [swallowed here](https://github.com/jenkinsci/jenkins/blob/b649939009f4a8c3356a953ca8d98a96244346c8/core/src/main/java/jenkins/slaves/restarter/UnixSlaveRestarter.java#L29-L45) nonetheless had trouble with warnings printed during Pipeline builds:

```
… FINE    o.j.p.w.s.d.DurableTaskStep$Execution#check: could not check /…
java.lang.UnsatisfiedLinkError: /tmp/jna-…/jna….tmp: /tmp/jna-…/jna….tmp: failed to map segment from shared object: Operation not permitted
    at java.lang.ClassLoader$NativeLibrary.load(Native Method)
    at java.lang.ClassLoader.loadLibrary0(ClassLoader.java:2045)
    at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1928)
    at java.lang.Runtime.load0(Runtime.java:809)
    at java.lang.System.load(System.java:1094)
    at com.sun.jna.Native.loadNativeDispatchLibraryFromClasspath(Native.java:851)
    at com.sun.jna.Native.loadNativeDispatchLibrary(Native.java:826)
    at com.sun.jna.Native.<clinit>(Native.java:140)
    at com.sun.akuma.CLibrary.<clinit>(CLibrary.java:89)
    at com.sun.akuma.JavaVMArguments.resolvePID(JavaVMArguments.java:128)
    at com.sun.akuma.JavaVMArguments.ofLinux(JavaVMArguments.java:116)
    at com.sun.akuma.JavaVMArguments.of(JavaVMArguments.java:104)
    at com.sun.akuma.JavaVMArguments.current(JavaVMArguments.java:92)
    at jenkins.slaves.restarter.UnixSlaveRestarter.canWork(UnixSlaveRestarter.java:29)
    at jenkins.slaves.restarter.JnlpSlaveRestarterInstaller$2.call(JnlpSlaveRestarterInstaller.java:66)
    at jenkins.slaves.restarter.JnlpSlaveRestarterInstaller$2.call(JnlpSlaveRestarterInstaller.java:52)
Caused: java.lang.NoClassDefFoundError: com/sun/jna/Native : cannot initialize class because prior initialization attempt failed
[...]

PR #49, opened 2017-11-01
[JENKINS-47791] Remove ProcessLiveness and PID tracking in favor of a simple timestamp check on the log file

jglick:
[JENKINS-47791](https://issues.jenkins-ci.org/browse/JENKINS-47791)

Allows us to close #48, #47, #43, #40, #37, and #35, and probably solve a bunch of thorny issues.

Main code changes:

- [X] basic change
- [x] update to `getDiagnostics`
- [x] make timing behaviors controllable by system property

Automated test coverage (downstream of https://github.com/jenkinsci/docker-fixtures/pull/9):

- [x] Docker-based test on Ubuntu
- [x] Docker-based test on CentOS
- [X] test to check for unreaped zombies
- [x] test of script with `#!` interpreter
- [X] `BourneShellScriptTest` passes on Mac OS X acc. to @michaelneale
- [x] does not regress `workflow-durable-task-step` tests
- [X] does not regress `docker-workflow` tests: https://github.com/jenkinsci/docker-workflow-plugin/pull/121
- [x] does not regress `WorkflowPluginTest` (ATH)

Manual sanity checks I expect to do:

- [x] run on master node
- [X] `docker run -i --rm --name agent -u root jenkinsci/slave:3.7-1 java -jar /usr/share/jenkins/slave.jar`
- [x] run on a JNLP agent
- [x] run in FreeBSD
- [X] run on Mac OS X (at least )
- [x] reboot agent’s computer
- [x] `STOP`/`CONT`/`TERM`/`KILL` main and wrapper processes
- [X] run on Bash layer for Windows 2016 Server

Other desirable sanity checks (volunteers welcome!):

- [ ] actual run on Mac OS X
- [ ] run on Cygwin
- [ ] use in Kubernetes (@scoheb seems to say it works?)
- [ ] use on heavily loaded system

@reviewbybees


jglick:
Produces zombie processes but maybe that was true before and is only solvable with `setsid`.

jglick:
Nope, that seems to be a genuine regression. Maybe I need to bite the bullet and implement [JENKINS-25503](https://issues.jenkins-ci.org/browse/JENKINS-25503) already.

jglick:
No, `setsid` does not help, with or without `--wait`.

michaelneale:
ran on OSX `mvn -Dtest=BourneShellScriptTest clean test`

```
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 115.365 s - in org.jenkinsci.plugins.durabletask.BourneShellScriptTest
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 7, Failures: 0, Errors: 0, Skipped: 0
[INFO] 
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 02:09 min
[INFO] Finished at: 2017-11-02T08:50:02-07:00
[INFO] Final Memory: 51M/964M
[INFO] ------------------------------------------------------------------------
```

scoheb:
This PR has helped make https://github.com/jenkinsci/kubernetes-plugin/pull/244 more stable...

abayer:
Approved modulo test failure - kicked off another run to see if it was just noise.

jglick:
@abayer see #50 for the test failure.

abayer:
@jglick Ah, gotcha.

svanoort:
@jglick What happens if ntp runs on the agent and the clock changes?

jglick:
> What happens if ntp runs on the agent and the clock changes?

IIRC `ntp` is designed to adjust the clock gradually to avoid any discontinuities. If an admin abruptly changes the system clock forward, then the next `touch` or process output should trigger `seeing recent log file modifications`. If abruptly backward, then I suppose the `sh` step will exit with -1 status.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -171,47 +202,36 @@ public FilePath getScriptFile(FilePath ws) throws IOException, InterruptedExcept
...
+            if (lastCheck == 0) {
+                LOGGER.log(Level.FINE, "starting check in {0}", controlDir);
+                lastCheck = now;
+            } else if (now > lastCheck + TimeUnit.SECONDS.toNanos(HEARTBEAT_CHECK_INTERVAL)) {
+                lastCheck = now;
+                long currentTimestamp = getLogFile(workspace).lastModified();
+                if (currentTimestamp == 0) {
+                    LOGGER.log(Level.FINE, "apparently never started in {0}", controlDir);
+                    return recordExitStatus(workspace, -2);
+                } else if (checkedTimestamp > 0 && currentTimestamp < checkedTimestamp + TimeUnit.SECONDS.toMillis(HEARTBEAT_MINIMUM_DELTA)) {
```
svanoort: 🐛 Need to check absolute value of timestamp differences -- that way we're covered if an agent has its clock skewed and then NTP runs to reset the clocks (possible to an earlier time).  It happens more often than one would think (have seen it introduce production issues in distributed systems using timestamps). 

Now it's possible skew will cause a build to report to report under HEARTBEAT_MINIMUM even when valid but from what I've seen usually the skews are either on the order of a second or two *OR* very large -- this will cover the clock skew case much better.

jglick:
> how is this going to work in terms of compatibility with durable tasks launched and running before the master restarts to install the update with this?

Well `ExecutorStepTest.serialForm` passes, if you trust that, but it is probably not too relevant.

Basically the `pid` file will exist but be ignored. The heartbeat task will not be running. So, if the process is continuing to produce output, it will run to completion; if it is stalled but alive, the step will exit with -1.

svanoort:
> Basically the pid file will exist but be ignored. The heartbeat task will not be running. So, if the process is continuing to produce output, it will run to completion; if it is stalled but alive, the step will exit with -1.

Mmm.  That's about what I feared -- probably worth noting in the release notes, but since it's a one-time thing and this fix is very impactful in a positive way (and 99% of processes should be producing regular output) I guess that's okay.

jglick:
OK. I am still watching an ATH build to look for obvious problems with https://github.com/jenkinsci/docker-fixtures/pull/10 before I release 1.2; I _just_ found and fixed a problem with https://github.com/jenkinsci/acceptance-test-harness/pull/375 (not affecting this PR at all).

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -151,4 +176,54 @@ public void smokeTest() throws Exception {
...
+            baos = new ByteArrayOutputStream();
+            assertEquals(0, dockerLauncher.launch().cmds("ps", "-e", "-o", "pid,stat,command").stdout(new TeeOutputStream(baos, System.out)).join());
+        } while (baos.toString().contains(" sleep "));
+        assertThat("no zombies running", baos.toString(), not(containsString(" Z ")));
+        s.toComputer().disconnect(new OfflineCause.UserCause(null, null));
+    }
+
+    @Test public void runWithCommandLauncher() throws Exception {
+        assumeTrue("Docker required for this test", new Docker().isAvailable());
+        runOnDocker(new DumbSlave("docker", "/home/jenkins/agent", new SimpleCommandLauncher("docker run -i --rm --name agent jenkinsci/slave:3.7-1 java -jar /usr/share/jenkins/slave.jar")));
```
jglick: @MarkEWaite points me to tini, which is apparently available as an `--init` option, so I should check whether that _also_ solves the zombie issue without `wait`, and if so, recommend its usage on the image page.

PR #50, opened 2017-11-09
Trying to diagnose some ContinuedTaskTest failures

jglick:
Saw some presumably unrelated failures in #49:

```
java.lang.AssertionError: expected:<1> but was:<0>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.jenkinsci.plugins.durabletask.executors.ContinuedTaskTest$1.perform(ContinuedTaskTest.java:55)
```

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
@@ -60,7 +60,7 @@ public class ContinuedTaskTest {
         });
         QueueTaskFuture<FreeStyleBuild> b1 = p.scheduleBuild2(0);
         r.jenkins.getQueue().schedule(new TestTask(cntA, false), 0);
-        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 1);
+        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 3);
         // cntB task ought to run first, then b1 and the cntA task in either order
         r.createSlave(label);
         r.assertBuildStatusSuccess(b1);
--- a/src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
@@ -60,7 +60,7 @@ public class ContinuedTaskTest {
         });
         QueueTaskFuture<FreeStyleBuild> b1 = p.scheduleBuild2(0);
         r.jenkins.getQueue().schedule(new TestTask(cntA, false), 0);
-        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 3);
+        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 0);
         // cntB task ought to run first, then b1 and the cntA task in either order
         r.createSlave(label);
         r.assertBuildStatusSuccess(b1);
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
```diff
@@ -51,15 +54,14 @@
         final AtomicInteger cntB = new AtomicInteger();
         p.getBuildersList().add(new TestBuilder() {
             @Override public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-                assertEquals(0, cntA.get());
```
jglick: see below

src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
```diff
@@ -51,15 +54,14 @@
...
                 assertEquals(1, cntB.get());
                 return true;
             }
         });
         QueueTaskFuture<FreeStyleBuild> b1 = p.scheduleBuild2(0);
         r.jenkins.getQueue().schedule(new TestTask(cntA, false), 0);
         r.jenkins.getQueue().schedule(new TestTask(cntB, true), 1);
-        // cntB task ought to run first, then b1, then the cntA task
+        // cntB task ought to run first, then b1 and the cntA task in either order
```
jglick: If there was some reason why `b1` would have to run before `cntA`, I have forgotten it. Seems like the order should be arbitrary.

src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
```diff
@@ -51,15 +54,14 @@
         final AtomicInteger cntB = new AtomicInteger();
         p.getBuildersList().add(new TestBuilder() {
             @Override public boolean perform(AbstractBuild<?, ?> build, Launcher launcher, BuildListener listener) throws InterruptedException, IOException {
-                assertEquals(0, cntA.get());
                 assertEquals(1, cntB.get());
```
jglick: This was the failing assertion.

src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
```diff
@@ -51,15 +54,14 @@
...
                 return true;
             }
         });
         QueueTaskFuture<FreeStyleBuild> b1 = p.scheduleBuild2(0);
         r.jenkins.getQueue().schedule(new TestTask(cntA, false), 0);
-        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 1);
-        // cntB task ought to run first, then b1, then the cntA task
+        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 0);
+        // cntB task ought to run first, then b1 and the cntA task in either order
```
jglick: If there was some reason why `b1` would have to run before `cntA`, I have forgotten it. Seems like the order should be arbitrary.

src/test/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTaskTest.java
```diff
@@ -51,15 +54,14 @@
...
-                assertEquals(0, cntA.get());
                 assertEquals(1, cntB.get());
                 return true;
             }
         });
         QueueTaskFuture<FreeStyleBuild> b1 = p.scheduleBuild2(0);
         r.jenkins.getQueue().schedule(new TestTask(cntA, false), 0);
-        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 1);
-        // cntB task ought to run first, then b1, then the cntA task
+        r.jenkins.getQueue().schedule(new TestTask(cntB, true), 0);
```
jglick: I think this might fix the failures.

PR #51, opened 2017-11-16
Fixed JENKINS-46876, JENKINS-46508, JENKINS-46496, JENKINS-48057, and JENKINS-47797

gabloe:
Fixing several issues with powershell hanging and incorrect handling of exit codes

[JENKINS-46876](https://issues.jenkins-ci.org/browse/JENKINS-46876) [JENKINS-46508](https://issues.jenkins-ci.org/browse/JENKINS-46508)  [JENKINS-46496](https://issues.jenkins-ci.org/browse/JENKINS-46496)  [JENKINS-48057](https://issues.jenkins-ci.org/browse/JENKINS-48057)  [JENKINS-47797 ](https://issues.jenkins-ci.org/browse/JENKINS-47797 )

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -192,15 +192,18 @@ public final class PowershellScript extends FileMonitoringTask {
                                              "  exit $LASTEXITCODE;\r\n" + 
                                              "}", powershellBinary, powershellArgs, quote(c.getPowerShellScriptFile(ws)));
         
+        // Add an explicit exit to the end of the script so that exit codes are propagated
+        String scriptWithExit = script + "\r\nexit $LASTEXITCODE;";
+        
         if (launcher.isUnix()) {
             // There is no need to add a BOM with Open PowerShell
             c.getPowerShellHelperFile(ws).write(helperScript, "UTF-8");
-            c.getPowerShellScriptFile(ws).write(script, "UTF-8");
+            c.getPowerShellScriptFile(ws).write(scriptWithExit, "UTF-8");
             c.getPowerShellWrapperFile(ws).write(scriptWrapper, "UTF-8");
         } else {
             // Write the Windows PowerShell scripts out with a UTF8 BOM
             writeWithBom(c.getPowerShellHelperFile(ws), helperScript);
-            writeWithBom(c.getPowerShellScriptFile(ws), script);
+            writeWithBom(c.getPowerShellScriptFile(ws), scriptWithExit);
             writeWithBom(c.getPowerShellWrapperFile(ws), scriptWrapper);
         }
         
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -120,7 +120,7 @@ public final class PowershellScript extends FileMonitoringTask {
         ")\r\n" +
         "  $exceptionCaught = $null\r\n" +
         "  try {\r\n" +
-        "    [System.Text.Encoding] $encoding = [System.Text.UTF8Encoding]::new( $false );\r\n" +
+        "    [System.Text.Encoding] $encoding = New-Object System.Text.UTF8Encoding $false;\r\n" +
         "    [System.Console]::OutputEncoding = [System.Console]::InputEncoding = $encoding;\r\n" +
         "    [System.IO.Directory]::SetCurrentDirectory( $PWD );\r\n" +
         "    $null = New-Item $LogFile -ItemType File -Force;\r\n" +
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -153,7 +153,7 @@ public final class PowershellScript extends FileMonitoringTask {
             args.addAll(Arrays.asList("powershell.exe", "-NonInteractive", "-ExecutionPolicy", "Bypass", "-File", quote(c.getPowerShellMainFile(ws))));    
         }
 
-        Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(false);
+        Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(true);
         listener.getLogger().println("[" + ws.getRemote().replaceFirst("^.+(\\\\|/)", "") + "] Running PowerShell script");
         ps.readStdout().readStderr();
         ps.start();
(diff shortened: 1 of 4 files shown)


gabloe:
@abayer @svanoort @jglick 
Can this PR please be reviewed?

jakauppila:
It would be great if this was pulled in; this has been biting us on a few production jobs.

jglick:
> Can this PR please be reviewed?

Just “request reviews” in GitHub as needed.

gabloe:
The issue seems to be caused by powershell unable to handle spaces in the path to the workspace.  Most likely this happens when the workspace is something like "C:\Program Files (x86)\Jenkins\workspace\<Job>" or something similar.  The result is that the process hangs indefinitely, which is incredibly odd that it doesn't just fail.  I could add a test that creates a workspace with spaces.  However, this PR adds a change to the parameters used when invoking powershell that will not cause it to suffer from this issue.

gabloe:
@jglick I would gladly request reviews but it seems that option is disabled for me. Under "Reviewers" I see no link/button or a gear icon like I normally see in PR's that I submit.

jglick:
> I could add a test that creates a workspace with spaces. However, this PR adds a change to the parameters used when invoking powershell that will not cause it to suffer from this issue.

So temporarily revert the `src/main/` changes, add the `src/test/` reproducing the bug, then reapply the `src/main/` part to confirm the fix.

gabloe:
I've done just that.  I reverted the change temporarily, added a test to introduce a workspace with spaces in the path, ran the unit tests and the new powershell test timed out consistently after ~180 seconds.  So I can confirm that spaces in the workspace path causes the powershell process to hang.  Then I pulled in my change and tested again and the powershell tests (including the new test) ran with 100% success.

gabloe:
Could someone install powershell core on the Linux azure images being used to run these tests?  Otherwise, there is no test coverage on powershell for Linux (aka powershell core).  Binaries/instructions can be found here: https://github.com/PowerShell/PowerShell/blob/master/README.md

gabloe:
The tests are failing prematurely due to what I think is a .NET compatibility issue.  I am investigating the issue now.  Meanwhile, can someone tell me details about the test environment?  What version of Windows, .NET, and PowerShell are installed?

chris5287:
Any progress on this review, as we stuck on v1.15 until this fix is released

jakauppila:
Any updates on this? I'd love to see it merged so we could upgrade.

gabloe:
@abayer @svanoort Ping for a review.

svanoort:
I am really not qualified to review Powershell, but will give it a once-over as best I can and am trying to get @jtnord to give it a look before merge (since he is qualified to review)

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -65,7 +66,7 @@
         String pathSeparator = properties.getProperty("path.separator");
         String[] paths = System.getenv("PATH").split(pathSeparator);
         boolean powershellExists = false;
-        String cmd = launcher.isUnix()?"powershell":"powershell.exe";
+        String cmd = launcher.isUnix()?"pwsh":"powershell.exe";
```
jtnord: Ahh and here be dragons...
People are using cygwin sshd to launch slaves on windows which I think ends up telling Jenkins that the OS is Unix, and well the powershell executable is `powershell` and not `pwsh` that it is on linux....

given that this will break any existing setup I am probably against this change here.

svanoort:
Seems reasonable AFAICT, assuming James's comment can be addressed (if so I'm happy to merge -- he's given permission to dismiss the review once it is)

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
gabloe: Good point. I did not consider Cygwin here. How would you prefer this to be handled? I could test for both binaries regardless of platform, but then the issue is that I assume pwsh for *nix and powershell.exe for windows in \src\main\java\org\jenkinsci\plugins\durabletask\PowershellScript.java, so I'd need extra logic to determine which binary should be used, which seems like extra overhead for just running a script.

Is there a way to detect if Cygwin is being used?

gabloe:
It would be fairly easy to extend this class to provide PS core functionality only. But this would break anyone using the powershell step to run PS core, since they would need to replace powershell with pwsh.  It would be great to be able to handle both sceneries under the powershell step.

Get Outlook for iOS<https://aka.ms/o0ukef>

jtnord:
I would keep calling powershell in this pr.  A change to do something different may require extra thought and so as this pr has useful fixes as is should not be held up with those discussions.

jglick:
cf. #55

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -86,68 +86,127 @@ public String getScript() {
...
-        "  exit 1;\r\n" + 
-        "} finally {\r\n" +
-        "  if ($LastExitCode -ne $null) {\r\n" + 
-        "    exit $LastExitCode;\r\n" +
-        "  } elseif ($error.Count -gt 0 -or !$?) {\r\n" + 
-        "    exit 1;\r\n" +
-        "  } else {\r\n" +
-        "    exit 0;\r\n" +
+        String helperScript = "function New-StreamWriter {\r\n" +
+        "[CmdletBinding()]\r\n" +
```
jglick: AFAICT this is a single string constant, so it would be more readable and maintainable to move it to `src/main/resources/org/jenkinsci/plugins/durabletask/powershellHelper.ps1` and copy it directly from there (`FilePath.copyFrom` with `Class.getResource`) to `getPowerShellHelperFile(ws)`.

jglick:
Sorry I introduced a merge conflict here, but it should be simple to solve—just adding `, listener` to `exitStatus` calls in the test.

src/main/resources/org/jenkinsci/plugins/durabletask/powershellHelper.ps1
```diff
@@ -0,0 +1,85 @@
+﻿# By default PowerShell adds a byte order mark (BOM) to the beginning of a file when using Out-File with a unicode encoding such as UTF8.
```
jtnord: @batmat set this repo to commit as "unix style line ends" so you should probably make sure this file has binary attributes in git to make sure it has Windows line endings (if line endings matter - but the previous code was using `\r\n`)

jtnord: line endings do not seem to matter

gabloe:
@jtnord Agreed. My latest commit preserves the old binary name. Ideally the edition of PowerShell (eg. Windows PowerShell or PowerShell Core) should be user definable.  Something like:
`powershell edition: 'PowerShellCore', script: '...'`

For now it can be temporarily resolved by creating a symlink that maps powershell to pwsh on *nix.

svanoort:
Thanks @gabloe @jtnord Merging now and will cut a release in the next day or two (investigating something that might require a durable-task fix, and want to include it in this release if needed).

chris5287:
@svanoort Any news on a new version release as our Jenkins CI is currently broken as we updated `Blue Ocean` (which depends on the newer `Durable Task` v1.16 via `Pipeline: Nodes and Processes` which has the Powershell bug in)?

oleo65:
@chris5287 we had the very same issue with broken powershell (even in v 1.17).
I ended up reverting `Pipeline: Nodes and Processes` to 2.17 and `Durable Task` to 1.15 which repaired our pipelines.
Working but still not the best solution.

svanoort:
@oleo65 @chris5287 @gabloe Just released now as of version 2.18 -- there were some local test failures with the master branch that needed investigation before it was safe to release.

PR #52, opened 2017-11-20
Check that we are running a sufficiently recent version of Docker

jglick:
Cf. [changelog](https://docs.docker.com/release-notes/docker-engine/#1130-2017-01-18). Follows up #49.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
@@ -31,6 +31,7 @@ import hudson.plugins.sshslaves.SSHLauncher;
 import hudson.slaves.DumbSlave;
 import hudson.slaves.OfflineCause;
 import hudson.util.StreamTaskListener;
+import hudson.util.VersionNumber;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -58,8 +60,11 @@
 
     @Rule public DockerRule<CentOSFixture> dockerCentOS = new DockerRule<>(CentOSFixture.class);
 
-    @Before public void unix() {
+    @BeforeClass public static void unixAndDocker() throws Exception {
         assumeTrue("This test is only for Unix", File.pathSeparatorChar==':');
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        assumeThat("`docker version` could be run", new Launcher.LocalLauncher(StreamTaskListener.fromStderr()).launch().cmds("docker", "version", "--format", "{{.Client.Version}}").stdout(new TeeOutputStream(baos, System.err)).stderr(System.err).join(), is(0));
+        assumeThat("Docker must be at least 1.13.0 for this test (uses --init)", new VersionNumber(baos.toString().trim()), greaterThanOrEqualTo(new VersionNumber("1.13.0")));
```
recena: @jglick What do you think whether one requires a minimum version of docker?

jglick: Not sure I understand the question.

PR #53, opened 2017-11-21
Specify java.level=7

jglick:
#49 inadvertently switched the plugin to use Java 8 despite a baseline of 2.7.x, due to a decision in https://github.com/jenkinsci/plugin-pom/pull/83 which I think was dangerous.

@reviewbybees

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,7 @@
     
     <properties>
       <jenkins.version>2.7.3</jenkins.version>
+      <java.level>7</java.level>
       <findbugs.failOnError>false</findbugs.failOnError>
     </properties>
     


jglick:
BTW I searched for some way to configure `plugin-pom/pom.xml` such that it would _force_ its children to specify `java.level`, but without success. Using a bogus default value does not work, since the build of `plugin-pom` itself makes use of `${java.level}` in signature checks and the like. The Enforcer rule `requireProperty` accepts inherited properties. Would have to define a new rule, I suppose; seems like a hassle.

batmat:
@jglick http://www.mojohaus.org/extra-enforcer-rules/requirePropertyDiverges.html ?

PR #54, opened 2017-12-06
[JENKINS-48415] PowerShell pipeline step hangs with spaces in path

mcaley:
Replaced double quotes with single quotes when launching PowerShell so that the string isn't escaped in the middle.


gabloe:
Hi @mcaley,
This issue is already fixed in my PR: https://github.com/jenkinsci/durable-task-plugin/pull/51

mcaley:
Awesome, thanks a lot @gabloe

PR #55, opened 2018-01-04
[Fixed JENKINS-48803] Updating PowerShell executable name on linux

LiamLeane:
The reasoning is really stupid and I certainly hope they setup an alias in a future release but powershell is now pwsh, this is in the latest RC and is part of the GA release for powershell next week. Given all previous releases of Powershell for Linux were pre-release I propose that detecting which executable name to use should not be required; if the durable task plugin is updated then so too should powershell (alternatively they can manually alias the command until a Powershell update is applied).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -143,7 +143,7 @@ public final class PowershellScript extends FileMonitoringTask {
 
         if (launcher.isUnix()) {
             // Open-Powershell does not support ExecutionPolicy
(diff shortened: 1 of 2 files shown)


gabloe:
This change needs to come in with some additional design. Ideally, I would like to see it as an option in  [Workflow Durable Task Step](https://github.com/jenkinsci/workflow-durable-task-step-plugin) and then exposed in durable task so that the user can select which edition of PowerShell they want to target (Windows PowerShell, or PowerShell Core). Otherwise, there are too many different configurations to handle that go beyond checking for Windows or *nix, such as the scenario @jtnord pointed out.

casz:
This does not consider Windows that may have pwsh installed

PR #56, opened 2018-01-24
Updates

jglick:
Noticed some recent PR failures relating to `docker-fixtures` and thought I would update that.

@reviewbybees


svanoort:
Looks OK to me but need to see tests pass from CI.

jglick:
Interesting…I specifically tested this locally (with a clean image cache) and it worked. Will try harder to reproduce.

PR #57, opened 2018-01-25
[JENKINS-48300] Add an overload for exitStatus taking TaskListener

jglick:
See discussion in [JENKINS-48300](https://issues.jenkins-ci.org/browse/JENKINS-48300). If we are going to return -1 (or -2) from `sh`, rather than an actual code, at least print some explanatory message.

@reviewbybees

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -207,8 +207,8 @@ public final class BourneShellScript extends FileMonitoringTask {
             return controlDir(ws).child("pid");
         }
 
-        @Override public Integer exitStatus(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
-            Integer status = super.exitStatus(workspace, launcher);
+        @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
+            Integer status = super.exitStatus(workspace, launcher, listener);
             if (status != null) {
                 LOGGER.log(Level.FINE, "found exit code {0} in {1}", new Object[] {status, controlDir});
                 return status;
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,8 +207,8 @@ private FilePath pidFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("pid");
         }
 
-        @Override public Integer exitStatus(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
-            Integer status = super.exitStatus(workspace, launcher);
+        @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
```
svanoort: 🐛  Binary compatibility breakage.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -159,7 +161,7 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
         }
 
         // TODO would be more efficient to allow API to consolidate writeLog with exitStatus (save an RPC call)
-        @Override public Integer exitStatus(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
+        @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
```
svanoort: 🐛 Darn it binary compatibility.

jglick:
What makes you think this is binary-incompatible? I introduced new method overloads and delegated from the old ones. The two methods you marked were merely overrides.

svanoort:
Ah, you're right -- I missed that without the full context.

jglick:
@reviewbybees done

PR #59, opened 2018-02-09
Using simpler Queue.withLock overload

jglick:
Unrelated simplification pulled out of #29.

@reviewbybees


svanoort:
Ahaha, "well let's use a callable that returns nothing" vs. "oh hey we could use a runnable"

PR #60, opened 2018-02-09
[JENKINS-52165] Controller.watch API

jglick:
[JENKINS-52165](https://issues.jenkins-ci.org/browse/JENKINS-52165) Just the minimum changes pulled out of #29 to enable push rather than pull logging from durable tasks.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -207,8 +207,8 @@ public final class BourneShellScript extends FileMonitoringTask {
             return controlDir(ws).child("pid");
         }
 
-        @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
-            Integer status = super.exitStatus(workspace, launcher, listener);
+        @Override protected Integer exitStatus(FilePath workspace, TaskListener listener) throws IOException, InterruptedException {
+            Integer status = super.exitStatus(workspace, listener);
             if (status != null) {
                 LOGGER.log(Level.FINE, "found exit code {0} in {1}", new Object[] {status, controlDir});
                 return status;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
@@ -44,6 +44,20 @@ import javax.annotation.Nonnull;
 public abstract class Controller implements Serializable {
 
     /**
+     * Begins watching the process asynchronously, so that the master may receive notification when output is available or the process has exited.
+     * This should be called as soon as the process is launched, and thereafter whenever reconnecting to the agent.
+     * You should not call {@link #writeLog} or {@link #cleanup} in this case; you do not need to call {@link #exitStatus(FilePath, Launcher)} frequently,
+     * though it is advisable to still call it occasionally to verify that the process is still running.
+     * @param workspace the workspace in use
+     * @param handler a remotable callback
+     * @param listener a remotable destination for messages
+     * @throws UnsupportedOperationException when this mode is not available, so you must fall back to polling {@link #writeLog} and {@link #exitStatus(FilePath, Launcher)}
+     */
+    public void watch(@Nonnull FilePath workspace, @Nonnull Handler handler, @Nonnull TaskListener listener) throws IOException, InterruptedException, UnsupportedOperationException {
+        throw new UnsupportedOperationException("Asynchronous mode is not implemented in " + getClass().getName());
+    }
+
+    /**
      * Obtains any new task log output.
      * Could use a serializable field to keep track of how much output has been previously written.
      * @param workspace the workspace in use
@@ -55,7 +69,7 @@ public abstract class Controller implements Serializable {
     /**
      * Checks whether the task has finished.
      * @param workspace the workspace in use
-     * @param launcher a way to start processes
+     * @param launcher a way to start processes (currently unused)
      * @param logger a way to report special messages
      * @return an exit code (zero is successful), or null if the task appears to still be running
      */
@@ -86,7 +100,7 @@ public abstract class Controller implements Serializable {
      * Intended for use after {@link #exitStatus(FilePath, Launcher)} has returned a non-null status.
      * The result is undefined if {@link DurableTask#captureOutput} was not called before launch; generally an {@link IOException} will result.
      * @param workspace the workspace in use
-     * @param launcher a way to start processes
+     * @param launcher a way to start processes (currently unused)
      * @return the output of the process as raw bytes (may be empty but not null)
      */
     public @Nonnull byte[] getOutput(@Nonnull FilePath workspace, @Nonnull Launcher launcher) throws IOException, InterruptedException {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -38,6 +38,8 @@ import hudson.slaves.WorkspaceList;
 import hudson.util.StreamTaskListener;
 import java.io.File;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
 import java.io.OutputStream;
 import java.io.RandomAccessFile;
 import java.io.StringWriter;
@@ -63,6 +65,7 @@ import javax.annotation.Nonnull;
 import jenkins.MasterToSlaveFileCallable;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.input.CountingInputStream;
+import org.apache.commons.io.input.ReaderInputStream;
 
 /**
  * A task which forks some external command and then waits for log and status files to be updated/created.
@@ -253,14 +256,23 @@ public abstract class FileMonitoringTask extends DurableTask {
          *         or null if not performing transcoding because it was not requested or the data was already thought to be in UTF-8
          */
         private static @CheckForNull ByteBuffer maybeTranscode(@Nonnull byte[] data, @CheckForNull String charset) {
-            if (charset == null) { // no transcoding requested, do raw copy and YMMV
+            Charset cs = transcodingCharset(charset);
+            if (cs == null) {
+                return null;
+            } else {
+                return StandardCharsets.UTF_8.encode(cs.decode(ByteBuffer.wrap(data)));
+            }
+        }
+
+        private static @CheckForNull Charset transcodingCharset(@CheckForNull String charset) {
+            if (charset == null) {
                 return null;
             } else {
                 Charset cs = charset.isEmpty() ? Charset.defaultCharset() : Charset.forName(charset);
                 if (cs.equals(StandardCharsets.UTF_8)) { // transcoding unnecessary as output was already UTF-8
                     return null;
                 } else { // decode output in specified charset and reëncode in UTF-8
-                    return StandardCharsets.UTF_8.encode(cs.decode(ByteBuffer.wrap(data)));
+                    return cs;
                 }
             }
         }
@@ -384,12 +396,14 @@ public abstract class FileMonitoringTask extends DurableTask {
         private final FilePath workspace;
         private final Handler handler;
         private final TaskListener listener;
+        private final @CheckForNull Charset cs;
 
         Watcher(FileMonitoringController controller, FilePath workspace, Handler handler, TaskListener listener) {
             this.controller = controller;
             this.workspace = workspace;
             this.handler = handler;
             this.listener = listener;
+            cs = FileMonitoringController.transcodingCharset(controller.charset);
         }
 
         @Override public void run() {
@@ -405,7 +419,9 @@ public abstract class FileMonitoringTask extends DurableTask {
                 if (len > lastLocation) {
                     assert !logFile.isRemote();
                     try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
-                        CountingInputStream cis = new CountingInputStream(Channels.newInputStream(ch.position(lastLocation)));
+                        InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
+                        InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
+                        CountingInputStream cis = new CountingInputStream(utf8EncodedStream);
                         handler.output(cis);
                         lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
                     }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -293,6 +293,7 @@ public abstract class FileMonitoringTask extends DurableTask {
     private static ScheduledExecutorService watchService;
     private synchronized static ScheduledExecutorService watchService() {
         if (watchService == null) {
+            // TODO 2.105+ use ClassLoaderSanityThreadFactory
             watchService = new /*ErrorLogging*/ScheduledThreadPoolExecutor(5, new NamingThreadFactory(new DaemonThreadFactory(), "FileMonitoringTask watcher"));
         }
         return watchService;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -142,7 +142,7 @@ public abstract class FileMonitoringTask extends DurableTask {
      * Tails a log file and watches for an exit status file.
      * Must be remotable so that {@link #watch} can transfer the implementation.
      */
-    protected static class FileMonitoringController extends Controller {
+    protected static class FileMonitoringController extends Controller { // TODO implements Remotable when available (*not* SerializableOnlyOverRemoting)
 
         /** Absolute path of {@link #controlDir(FilePath)}. */
         String controlDir;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -499,6 +499,21 @@ public abstract class FileMonitoringTask extends DurableTask {
             } catch (Exception x) {
                 // note that LOGGER here is going to the agent log, not master log
                 LOGGER.log(Level.WARNING, "giving up on watching " + controller.controlDir, x);
+                // Typically this will have been inside Handler.output, e.g.:
+                // hudson.remoting.ChannelClosedException: channel is already closed
+                //         at hudson.remoting.Channel.send(Channel.java:667)
+                //         at hudson.remoting.ProxyOutputStream.write(ProxyOutputStream.java:143)
+                //         at hudson.remoting.RemoteOutputStream.write(RemoteOutputStream.java:110)
+                //         at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1793)
+                //         at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1769)
+                //         at org.apache.commons.io.IOUtils.copy(IOUtils.java:1744)
+                //         at org.jenkinsci.plugins.workflow.steps.durable_task.DurableTaskStep$HandlerImpl.output(DurableTaskStep.java:503)
+                //         at org.jenkinsci.plugins.durabletask.FileMonitoringTask$Watcher.run(FileMonitoringTask.java:477)
+                // Thus we assume the log sink is hopeless and the Watcher task dies.
+                // If and when the agent is reconnected, a new watch call will be made and we will resume streaming.
+                // last-location.txt will record the last successfully written block of output;
+                // we cannot know reliably how much of the problematic block was actually received by the sink,
+                // so we err on the side of possibly duplicating text rather than losing text.
             }
         }
 
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -476,8 +476,11 @@ public abstract class FileMonitoringTask extends DurableTask {
                         InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
                         CountingInputStream cis = new CountingInputStream(locallyEncodedStream);
                         InputStream utf8EncodedStream = cs == null ? cis : new ReaderInputStream(new InputStreamReader(cis, cs), StandardCharsets.UTF_8);
-                        handler.output(utf8EncodedStream);
-                        lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
+                        try {
+                            handler.output(utf8EncodedStream);
+                        } finally {
+                            lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
+                        }
                     }
                 }
                 if (exitStatus != null) {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -474,11 +474,8 @@ public abstract class FileMonitoringTask extends DurableTask {
                     try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
                         InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
                         InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
-                        try {
-                            handler.output(utf8EncodedStream);
-                        } finally {
-                            lastLocationFile.write(Long.toString(ch.position()), null);
-                        }
+                        handler.output(utf8EncodedStream);
+                        lastLocationFile.write(Long.toString(ch.position()), null);
                     }
                 }
                 if (exitStatus != null) {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -474,9 +474,9 @@ public abstract class FileMonitoringTask extends DurableTask {
                     assert !logFile.isRemote();
                     try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
                         InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
-                        InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
-                        CountingInputStream cis = new CountingInputStream(utf8EncodedStream);
-                        handler.output(cis);
+                        CountingInputStream cis = new CountingInputStream(locallyEncodedStream);
+                        InputStream utf8EncodedStream = cs == null ? cis : new ReaderInputStream(new InputStreamReader(cis, cs), StandardCharsets.UTF_8);
+                        handler.output(utf8EncodedStream);
                         lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
                     }
                 }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -69,7 +69,6 @@ import javax.annotation.Nonnull;
 import jenkins.MasterToSlaveFileCallable;
 import jenkins.security.MasterToSlaveCallable;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.io.input.CountingInputStream;
 import org.apache.commons.io.input.ReaderInputStream;
 import org.apache.commons.io.output.CountingOutputStream;
 import org.apache.commons.io.output.WriterOutputStream;
@@ -474,12 +473,11 @@ public abstract class FileMonitoringTask extends DurableTask {
                     assert !logFile.isRemote();
                     try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
                         InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
-                        CountingInputStream cis = new CountingInputStream(locallyEncodedStream);
-                        InputStream utf8EncodedStream = cs == null ? cis : new ReaderInputStream(new InputStreamReader(cis, cs), StandardCharsets.UTF_8);
+                        InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
                         try {
                             handler.output(utf8EncodedStream);
                         } finally {
-                            lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
+                            lastLocationFile.write(Long.toString(ch.position()), null);
                         }
                     }
                 }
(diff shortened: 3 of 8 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -248,7 +275,102 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+            return controlDir(workspace).child("last-location.txt");
+        }
+
+        private static final long serialVersionUID = 1L;
+    }
+
+    private static ScheduledExecutorService watchService;
+    private synchronized static ScheduledExecutorService watchService() {
+        if (watchService == null) {
+            watchService = new /*ErrorLogging*/ScheduledThreadPoolExecutor(5, new NamingThreadFactory(new DaemonThreadFactory(), "FileMonitoringTask watcher"));
```
svanoort: TODO Use the new `ClassLoaderSanityThreadFactory` when available from core

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -248,7 +275,103 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                    lastLocation = Long.parseLong(lastLocationFile.readToString());
+                }
+                FilePath logFile = controller.getLogFile(workspace);
+                long len = logFile.length();
+                if (len > lastLocation) {
+                    assert !logFile.isRemote();
+                    try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
+                        CountingInputStream cis = new CountingInputStream(Channels.newInputStream(ch.position(lastLocation)));
+                        handler.output(cis);
+                        lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
```
jglick: Seems that this would _not_ suffer from [JENKINS-37575](https://issues.jenkins-ci.org/browse/JENKINS-37575)-like problems. We intentionally write `lastLocationFile` only after a successful call to `output`, so that if there is an agent channel outage we do not lose output. In such a case this `Watcher` simply dies; if and when the agent is reconnected, a new one should start at the same position.

src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
```diff
@@ -43,6 +43,20 @@
...
+    /**
+     * Begins watching the process asynchronously, so that the master may receive notification when output is available or the process has exited.
+     * This should be called as soon as the process is launched, and thereafter whenever reconnecting to the agent.
+     * You should not call {@link #writeLog} or {@link #cleanup} in this case; you do not need to call {@link #exitStatus(FilePath, Launcher)} frequently,
+     * though it is advisable to still call it occasionally to verify that the process is still running.
+     * @param workspace the workspace in use
+     * @param handler a remotable callback
+     * @param listener a remotable destination for messages
+     * @throws UnsupportedOperationException when this mode is not available, so you must fall back to polling {@link #writeLog} and {@link #exitStatus(FilePath, Launcher)}
+     */
```
oleg-nenashev: IOException needs to be documented. https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/65 suggests that it happens iff the node is offline, but there is no evidence of that here

src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
```diff
@@ -86,7 +100,7 @@
      * Intended for use after {@link #exitStatus(FilePath, Launcher)} has returned a non-null status.
      * The result is undefined if {@link DurableTask#captureOutput} was not called before launch; generally an {@link IOException} will result.
      * @param workspace the workspace in use
-     * @param launcher a way to start processes
+     * @param launcher a way to start processes (currently unused)
```
oleg-nenashev: same

src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
```diff
@@ -55,7 +69,7 @@
     /**
      * Checks whether the task has finished.
      * @param workspace the workspace in use
-     * @param launcher a way to start processes
+     * @param launcher a way to start processes (currently unused)
```
oleg-nenashev: The method is overridable, and the downstream implementation may actually use it but inherit Javadocs. So I would rather not touch it

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -100,6 +108,10 @@ protected FileMonitoringController doLaunch(FilePath workspace, Launcher launche
         return m;
     }
 
+    /**
+     * Tails a log file and watches for an exit status file.
+     * Must be remotable so that {@link #watch} can transfer the implementation.
+     */
     protected static class FileMonitoringController extends Controller {
```
oleg-nenashev: `SerializableOnlyOverRemoting` once it gets to 2.107+ ? Not the best API, but we do not have better annotations now (should be fixed eventually once we get to any kind of static analysis)

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -194,19 +206,33 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
         static final StatusCheck STATUS_CHECK_INSTANCE = new StatusCheck();
 
         @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
+            return exitStatus(workspace, listener);
+        }
+
+        /**
+         * Like {@link #exitStatus(FilePath, Launcher, TaskListener)} but not requesting a {@link Launcher}, which would not be available in {@link #watch} mode anyway.
+         */
+        protected Integer exitStatus(FilePath workspace, TaskListener listener) throws IOException, InterruptedException {
```
oleg-nenashev: Add CheckForNull so nobody unboxes it to `int` with FindBugs being silent?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -194,19 +206,33 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
...
 
         @Override public byte[] getOutput(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
-            // TODO could perhaps be more efficient for large files to send a MasterToSlaveFileCallable<byte[]>
+            return getOutput(workspace);
+        }
+
+        /**
+         * Like {@link #getOutput(FilePath, Launcher)} but not requesting a {@link Launcher}, which would not be available in {@link #watch} mode anyway.
+         */
+        protected byte[] getOutput(FilePath workspace) throws IOException, InterruptedException {
```
oleg-nenashev: Maybe it's better to introduce a new method with optional Launcher so that it's not ignored in the implementations?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -272,7 +298,103 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+        @Override public void run() {
+            try {
+                Integer exitStatus = controller.exitStatus(workspace, listener); // check before collecting output, in case the process is just now finishing
+                long lastLocation = 0;
+                FilePath lastLocationFile = controller.getLastLocationFile(workspace);
+                if (lastLocationFile.exists()) {
+                    lastLocation = Long.parseLong(lastLocationFile.readToString());
+                }
+                FilePath logFile = controller.getLogFile(workspace);
+                long len = logFile.length();
```
oleg-nenashev: Maybe check the file's existence just in case? otherwise the watcher will give up immediately (IOException ). But IIUC the code it may want to wait to avoid race conditions. Or not?

src/main/java/org/jenkinsci/plugins/durabletask/Handler.java
```diff
@@ -0,0 +1,67 @@
...
+import java.io.Serializable;
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+/**
+ * A remote handler which may be sent to an agent and handle process output and results.
+ * If it needs to communicate with the master, you may use {@link VirtualChannel#export}.
+ * @see Controller#watch
+ */
+public abstract class Handler implements Serializable {
```
oleg-nenashev: SerializableOnlyOverRemoting after 2.107+

src/main/java/org/jenkinsci/plugins/durabletask/Handler.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Notification that new process output is available.
+     * <p>Should only be called when at least one byte is available.
+     * Whatever bytes are actually read will not be offered on the next call, if there is one; there is no need to close the stream.
+     * <p>There is no guarantee that output is offered in the form of complete lines of text,
+     * though in the typical case of line-oriented output it is likely that it will end in a newline.
+     * <p>Buffering is the responsibility of the caller, and {@link InputStream#markSupported} may be false.
+     * @param stream a way to read process output which has not already been handled
+     * @throws Exception if anything goes wrong, this watch is deactivated
```
oleg-nenashev: I would rather use a more narrow exception type in `throws` so that Handler implementers do not start throwing RuntimeExceptions. But you will have to catch them when invoking Handler methods anyway...

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -194,19 +206,33 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
...
+         */
+        protected byte[] getOutput(FilePath workspace) throws IOException, InterruptedException {
             try (InputStream is = getOutputFile(workspace).read()) {
                 return IOUtils.toByteArray(is);
             }
         }
 
         @Override public final void stop(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
             launcher.kill(Collections.singletonMap(COOKIE, cookieFor(workspace)));
+            // TODO after 10s, if the control dir still exists, write a flag file and have the Watcher shut down (interrupting any ongoing handler.output call if possible)
```
oleg-nenashev: Makes sense to add logging at least

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -272,7 +298,103 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                    }
+                    handler.exited(exitStatus, output);
+                    controller.cleanup(workspace);
+                } else {
+                    // Could use an adaptive timeout as in DurableTaskStep.Execution in polling mode,
+                    // though less relevant here since there is no network overhead to the check.
+                    watchService().schedule(this, 100, TimeUnit.MILLISECONDS);
+                }
+            } catch (Exception x) {
+                // note that LOGGER here is going to the agent log, not master log
```
oleg-nenashev: so Runtime exceptions will be non-recoverable here... Looks fine assuming that we want Handler implementations to be written properly.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Could do that, but actually I think I want to mark this

```java
@Restricted(ProtectedExternally.class)
```

as a reminder that only this class is expected to be calling it.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Not sure I follow what you are saying.

To be clear, this overload needed to be introduced so it can be called from the watch task, which cannot produce a (usable) `Launcher`; the original overload was not using its `launcher` parameter anyway.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: If the file does not exist (because the wrapper script is still launching, or failed to launch), `length()` will be zero, thus nothing will happen on this iteration and we check again in another 100ms.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: If there is an error because the channel is closed, then if and when a new agent is started, a new watcher should be created; so, fine. If there is some other kind of error, then yes this is fatal—we stop sending status updates from this process.

src/main/java/org/jenkinsci/plugins/durabletask/Handler.java
jglick: Perhaps, yes. As discussed above.

src/main/java/org/jenkinsci/plugins/durabletask/Handler.java
jglick: I do not really care what type is thrown. If anything is caught, the watch task shuts down.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Never mind—plain `protected` already accomplishes that. Can annotate nullness.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,9 +207,8 @@ private FilePath pidFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("pid");
         }
 
-        // TODO run as one big MasterToSlaveCallable<Integer> to avoid extra network roundtrips
```
jglick: (was added in #66 but is obsolete here)

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -252,7 +304,7 @@ private void runOnDocker(DumbSlave s) throws Exception {
         dockerWS.child("eastern").write("Čau!\n", "ISO-8859-2");
         dockerWS.child("mixed").write("¡Čau → there!\n", "UTF-8");
         Launcher dockerLauncher = s.createLauncher(listener);
-        assertEncoding("control: no transcoding", "latin", null, "¡Ole!", "ISO-8859-1", dockerWS, dockerLauncher);
+        assertEncoding("control: no transcoding", "latin", null, "¡Ole!", "ISO-8859-1", false, dockerWS, dockerLauncher);
         assertEncoding("test: specify particular charset (UTF-8)", "mixed", "UTF-8", "¡Čau → there!", "UTF-8", dockerWS, dockerLauncher);
         assertEncoding("test: specify particular charset (unrelated)", "eastern", "ISO-8859-2", "Čau!", "UTF-8", dockerWS, dockerLauncher);
         assertEncoding("test: specify agent default charset", "latin", "", "¡Ole!", "UTF-8", dockerWS, dockerLauncher);
```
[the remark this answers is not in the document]
jglick: Tried that, but did not really work out. Just catting a text file with a newline does not do what you expect, and echoing of commands produces garbage. Generally, I doubt we can support use of encodings which are not supersets of ASCII, though #28 purports to add EBCDIC support.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -252,64 +300,4 @@ private void runOnDocker(DumbSlave s, int sleepSeconds) throws Exception {
...
-        @Override public String call() throws RuntimeException {
-            return Charset.defaultCharset().name();
-        }
-    }
-    private void assertEncoding(String description, String file, String charset, String expected, String expectedEncoding, FilePath dockerWS, Launcher dockerLauncher) throws Exception {
-        assertEncoding(description, file, charset, expected, expectedEncoding, false, dockerWS, dockerLauncher);
-        assertEncoding(description, file, charset, expected, expectedEncoding, true, dockerWS, dockerLauncher);
-    }
-    private void assertEncoding(String description, String file, String charset, String expected, String expectedEncoding, boolean output, FilePath dockerWS, Launcher dockerLauncher) throws Exception {
-        System.err.println(description + " (output=" + output + ")"); // TODO maybe this should just be moved into a new class and @RunWith(Parameterized.class) for clarity
```
jglick: Did that refactoring, adding the `watch` parameter in the process.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -367,7 +399,111 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+        @Override public void watch(FilePath workspace, Handler handler, TaskListener listener) throws IOException, InterruptedException, ClassCastException {
+            workspace.actAsync(new StartWatching(this, handler, listener));
+            LOGGER.log(Level.FINE, "started asynchronous watch in {0}", controlDir);
+        }
+
+        /**
+         * File in which a last-read position is stored if {@link #watch} is used.
+         */
+        public FilePath getLastLocationFile(FilePath workspace) throws IOException, InterruptedException {
+            return controlDir(workspace).child("last-location.txt");
```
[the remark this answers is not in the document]
jglick: Anyway would be a topic for https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/63 not this PR.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -367,7 +399,114 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                if (len > lastLocation) {
+                    assert !logFile.isRemote();
+                    try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
+                        InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
+                        CountingInputStream cis = new CountingInputStream(locallyEncodedStream);
+                        InputStream utf8EncodedStream = cs == null ? cis : new ReaderInputStream(new InputStreamReader(cis, cs), StandardCharsets.UTF_8);
+                        try {
+                            handler.output(utf8EncodedStream);
+                        } finally {
+                            lastLocationFile.write(Long.toString(lastLocation + cis.getByteCount()), null);
```
dwnusbaum: Could you use `ch.position()` instead and get rid of `CountingInputStream` entirely, or am I missing something?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -251,11 +269,25 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
...
         }
 
         @Override public byte[] getOutput(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
+            return getOutput(workspace);
+        }
+
+        /**
+         * Like {@link #getOutput(FilePath, Launcher)} but not requesting a {@link Launcher}, which would not be available in {@link #watch} mode anyway.
+         */
+        protected byte[] getOutput(FilePath workspace) throws IOException, InterruptedException {
```
svanoort: If we're going to rework the way output is done, I'd *really* like to return a stream-able data implementation rather than raw `byte[]` - this ensures we're not allocating giant wads of memory to collect all the data if something produces logs quickly.  Whether it's a Stream, Channel, whatever I do not care as much. 

Consider also something like 'writeOutput' that internally feeds data to a sink.

It's also just plain more elegant and easier to work with in general.

src/main/java/org/jenkinsci/plugins/durabletask/Controller.java
```diff
@@ -43,6 +44,21 @@
...
+     * This should be called as soon as the process is launched, and thereafter whenever reconnecting to the agent.
+     * You should not call {@link #writeLog} or {@link #cleanup} in this case; you do not need to call {@link #exitStatus(FilePath, Launcher)} frequently,
+     * though it is advisable to still call it occasionally to verify that the process is still running.
+     * @param workspace the workspace in use
+     * @param handler a remotable callback
+     * @param listener a remotable destination for messages
+     * @throws IOException if initiating the watch fails, for example with a {@link ChannelClosedException}
+     * @throws UnsupportedOperationException when this mode is not available, so you must fall back to polling {@link #writeLog} and {@link #exitStatus(FilePath, Launcher)}
+     */
+    public void watch(@Nonnull FilePath workspace, @Nonnull Handler handler, @Nonnull TaskListener listener) throws IOException, InterruptedException, UnsupportedOperationException {
```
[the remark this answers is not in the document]
jglick: During `DurableTaskStep.Execution.start`: the `sh` step would fail. During `getWorkspace`: we retry later.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -367,7 +399,114 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+        }
+
+        private static final long serialVersionUID = 1L;
+    }
+
+    private static ScheduledExecutorService watchService;
+    private synchronized static ScheduledExecutorService watchService() {
+        if (watchService == null) {
+            // TODO 2.105+ use ClassLoaderSanityThreadFactory
+            watchService = new /*ErrorLogging*/ScheduledThreadPoolExecutor(5, new NamingThreadFactory(new DaemonThreadFactory(), "FileMonitoringTask watcher"));
```
svanoort: Wish we could use a thread pool that grows as needed for load (with an upper limit) but still supports scheduling.  Sigh.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -251,11 +269,25 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
...
+            return getOutput(workspace);
+        }
+
+        /**
+         * Like {@link #getOutput(FilePath, Launcher)} but not requesting a {@link Launcher}, which would not be available in {@link #watch} mode anyway.
+         */
+        protected byte[] getOutput(FilePath workspace) throws IOException, InterruptedException {
             return getOutputFile(workspace).act(new MasterToSlaveFileCallable<byte[]>() {
                 @Override public byte[] invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
                     byte[] buf = FileUtils.readFileToByteArray(f);
```
svanoort: Sigh, so much wish for streaming use.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: Actually... we could still count location that way using a CountingOutputStream I believe.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: ( I know, I know, not part of the standard library )

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Probably. My concern here is to have `last-location.txt` record the last position that was actually delivered to the `Handler`, in case something broke the streaming midway. It is maybe clearer that this would correspond to how much of the `locallyEncodedStream` was processed as opposed to the behavior of `newInputStream`, but OTOH that method does claim that the stream is unbuffered so I guess it should work and be simpler. (And anyway I have not managed to test this scenario.) I will try it.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Was already discussed earlier. This method is only used when you specify `returnStdout: true`, a fairly specialized case which is anyway only suited for small amounts of output—in the case of a `workflow-cps` caller, the entire resulting `String` gets saved to `program.dat`.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Ditto, unnecessary here.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: Damnit, we've gone in circles on these PRs so long that I'm repeating myself -- at least I'm consistent though! 😆

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -367,7 +398,113 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                long len = logFile.length();
+                if (len > lastLocation) {
+                    assert !logFile.isRemote();
+                    try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
+                        InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
+                        InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
+                        try {
+                            handler.output(utf8EncodedStream);
+                        } finally {
+                            lastLocationFile.write(Long.toString(ch.position()), null);
```
svanoort: 🐜 Noted that without atomic writes here, if the agent is restarted during writing, then we lose the position within the log and re-send *I think* the entire log.  Potential for looping if somehow we get repeated failures (very unlikely though).

svanoort: What if the handler throws an exception (likely IOException) while writing the output and the output isn't actually saved?  Suppose, for example that we can't connect to whatever is supposed to receive the log output.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -367,7 +398,113 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                }
+                FilePath logFile = controller.getLogFile(workspace);
+                long len = logFile.length();
+                if (len > lastLocation) {
+                    assert !logFile.isRemote();
+                    try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
+                        InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
+                        InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
+                        try {
+                            handler.output(utf8EncodedStream);
```
svanoort: @jglick Is there an enclosing Timeout somewhere here so we can't hang indefinitely?  May be somewhere non-obvious since this has so many levels of indirection that my brain is throwing a StackOverflowException trying to process it, but if there *isn't* one there should be, and if there *is* one please note it somewhere obvious so this doesn't get flagged again.

Note also: as *long* as you haven't advanced the location marker it *should* be safe to try re-sending later... assuming this doesn't cause the whole thing to bomb out.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: Wait, at that point will we bomb out completely and never try to send the log again?  Gah, this code is _literally_ giving me headaches to follow the flow of execution because it's so indirect.

@jglick C'mon man, a _few_ more explanatory comments please?

svanoort: Let me say what I *expect*: if the handler throws an exception (probably `IOExeption`) while trying to handle output, or takes longer than some timeout period, we break out (without mutating the lastLocationFile), and retry at least once after an appropriate delay.  

At some point if issues are resolved so the handler passes, we *should* (via mechanism unspecified) be able to get the entire log re-sent from the point of failure, including any new content that was added.

jglick: > if the handler throws an exception (probably `IOExeption`)

I need to note in a comment that the default handler does not currently ever throw `IOException` as a result of a sink issue (only as a result of a local file error).

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: What if it *halts*?  I'm not clear what the result is there, and what if that happens a bunch of times in a row?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -367,7 +398,110 @@ public FilePath getOutputFile(FilePath workspace) throws IOException, Interrupte
...
+                }
+                FilePath logFile = controller.getLogFile(workspace);
+                long len = logFile.length();
+                if (len > lastLocation) {
+                    assert !logFile.isRemote();
+                    try (FileChannel ch = FileChannel.open(Paths.get(logFile.getRemote()), StandardOpenOption.READ)) {
+                        InputStream locallyEncodedStream = Channels.newInputStream(ch.position(lastLocation));
+                        InputStream utf8EncodedStream = cs == null ? locallyEncodedStream : new ReaderInputStream(new InputStreamReader(locallyEncodedStream, cs), StandardCharsets.UTF_8);
+                        handler.output(utf8EncodedStream);
+                        lastLocationFile.write(Long.toString(ch.position()), null);
```
dwnusbaum: I think a comment explaining that this is intentionally not in a finally block so that we duplicate data rather than skip it when an error occurs would be helpful.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Changed as of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/63/commits/62aeb458e7b1da123fc77228c9821ca95a00ddbd.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Not sure what you mean by “it halts” in this context.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
dwnusbaum: If there is a non-remoting error, then it looks like the build will keep running until `controller.exitStatus` returns `null in `DurableTaskStep.Execution#check`. Would it helpful/possible to add some way to indicate to DurableTaskStep that logging appears to be broken? A passing build with broken logs would be strange, although failing the build because logging is broken doesn't seem any better.

If the channel is closed, then everything should retry in `DurableTask#getWorkspace` during the next call to `DurableTaskStep.Execution#check` so that seems fine to me, although the 5 minute recurrence period seems a little long.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Just added that.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: > Would it helpful/possible to add some way to indicate to DurableTaskStep that logging appears to be broken?

Possibly. I will think about another method in `Handler` for this purpose. As you note, in the case of a closed channel, the agent likely cannot communicate that error by definition, and `getWorkspace` is responsible for fixing things up a few minutes later. But in the case of a cloudy log implementation it could be useful to indicate to the master that some other error occurred, logs may be faulty, and we should rewatch when possible. Whether anything _else_ in the build gets successfully logged at that point, who knows.

> A passing build with broken logs would be strange, although failing the build because logging is broken doesn't seem any better.

Yeah it seems a bit of a toss-up. My general expectation is that if something strange happens without a clear recovery procedure, we should err on the side of just failing the build.

jglick: https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/63/commits/c859378dc5759222abd3caee63371052515366b5 _may_ produce better outcomes for cloudy log sinks. Without a test case it is hard to predict exactly what the failure modes will be, much less what response a user would actually find helpful.

`ExecutorStepTest.shellOutputAcrossRestart` does now exercise the code path under discussion in the case of the stock Remoting sink. The aforementioned commit does not change the behavior in that test, because it is indeed a `ChannelClosedException` that you get when the agent is disconnected from the master side. (If you just kill the agent JVM via either SIGTERM or SIGKILL, then the watcher task dies with it and you are also OK.)

PR #61, opened 2018-02-12
[JEP-206] Define API for gathering command output in a local encoding

jglick:
[JEP-206](https://github.com/jenkinsci/jep/blob/master/jep/206/README.adoc)

Extracted from #29. Downstream of #78.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
@@ -31,6 +31,8 @@ import hudson.Launcher;
 import hudson.model.AbstractDescribableImpl;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.nio.charset.Charset;
+import javax.annotation.Nonnull;
 
 /**
  * A task which may be run asynchronously on a build node and withstand disconnection of the slave agent.
@@ -61,6 +63,25 @@ public abstract class DurableTask extends AbstractDescribableImpl<DurableTask> i
      */
     public void captureOutput() throws UnsupportedOperationException {
         throw new UnsupportedOperationException("Capturing of output is not implemented in " + getClass().getName());
+    }
+
+    /**
+     * Requests that a specified charset be used to transcode process output.
+     * The encoding of {@link Controller#writeLog} and {@link Controller#getOutput} is then presumed to be UTF-8.
+     * If not called, no translation is performed.
+     * @param cs the character set in which process output is expected to be
+     */
+    public void charset(@Nonnull Charset cs) {
+        // by default, ignore
+    }
+
+    /**
+     * Requests that the node’s system charset be used to transcode process output.
+     * The encoding of {@link Controller#writeLog} and {@link Controller#getOutput} is then presumed to be UTF-8.
+     * If not called, no translation is performed.
+     */
+    public void defaultCharset() {
+        // by default, ignore
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
@@ -32,6 +32,8 @@ import hudson.model.AbstractDescribableImpl;
 import hudson.model.TaskListener;
 import java.io.IOException;
 import java.nio.charset.Charset;
+import java.util.logging.Level;
+import java.util.logging.Logger;
 import javax.annotation.Nonnull;
 
 /**
@@ -39,6 +41,8 @@ import javax.annotation.Nonnull;
  * Should have a descriptor, and a {@code config.jelly} for form data binding.
  */
 public abstract class DurableTask extends AbstractDescribableImpl<DurableTask> implements ExtensionPoint {
+
+    private static final Logger LOGGER = Logger.getLogger(DurableTask.class.getName());
 
     @Override public DurableTaskDescriptor getDescriptor() {
         return (DurableTaskDescriptor) super.getDescriptor();
@@ -72,7 +76,7 @@ public abstract class DurableTask extends AbstractDescribableImpl<DurableTask> i
      * @param cs the character set in which process output is expected to be
      */
     public void charset(@Nonnull Charset cs) {
-        // by default, ignore
+        LOGGER.log(Level.WARNING, "The charset method should be overridden in {0}", getClass().getName());
     }
 
     /**
@@ -81,7 +85,7 @@ public abstract class DurableTask extends AbstractDescribableImpl<DurableTask> i
      * If not called, no translation is performed.
      */
     public void defaultCharset() {
-        // by default, ignore
+        LOGGER.log(Level.WARNING, "The defaultCharset method should be overridden in {0}", getClass().getName());
     }
 
 }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -24,6 +24,7 @@
 
 package org.jenkinsci.plugins.durabletask;
 
+import com.google.common.io.Files;
 import hudson.EnvVars;
 import hudson.FilePath;
 import hudson.Launcher;
@@ -54,6 +55,7 @@ import javax.annotation.CheckForNull;
 import javax.annotation.Nonnull;
 import jenkins.MasterToSlaveFileCallable;
 import org.apache.commons.io.FileUtils;
+import org.apache.commons.io.IOUtils;
 
 /**
  * A task which forks some external command and then waits for log and status files to be updated/created.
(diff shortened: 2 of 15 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -58,19 +64,34 @@
 
     private static final String COOKIE = "JENKINS_SERVER_COOKIE";
 
+    /**
+     * Charset name to use for transcoding, or the empty string for node system default, or null for no transcoding.
```
svanoort: Wait, empty string and null have different handling? 

Uh..... 🐛 - if the behavior is different that should be a custom string, i.e. 'DEFAULT' or 'PRESERVE' or something, because otherwise this is going to trip us up somewhere.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -174,13 +205,42 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
+         */
+        private static @CheckForNull ByteBuffer maybeTranscode(@Nonnull byte[] data, @CheckForNull String charset) {
+            if (charset == null) { // no transcoding requested, do raw copy and YMMV
+                return null;
+            } else {
+                Charset cs = charset.isEmpty() ? Charset.defaultCharset() : Charset.forName(charset);
+                if (cs.equals(StandardCharsets.UTF_8)) { // transcoding unnecessary as output was already UTF-8
+                    return null;
+                } else { // decode output in specified charset and reëncode in UTF-8
+                    return StandardCharsets.UTF_8.encode(cs.decode(ByteBuffer.wrap(data)));
```
svanoort: 🐛 We should do everything we can to allocate and reuse a buffer here -- i.e. create a transient byte[] / Buffer that is used for storing encoded output, rather than `StandardCharsets.UTF_8.encode` -- otherwise we're going to generate massive amounts of memory garbage.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -174,13 +205,42 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
+                }
             }
         }
 
+        @Override public byte[] getOutput(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
+            return getOutputFile(workspace).act(new MasterToSlaveFileCallable<byte[]>() {
+                @Override public byte[] invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
+                    byte[] buf = FileUtils.readFileToByteArray(f);
```
svanoort: 🐛 Allocating and reading the whole file rather than doing streaming handling (or doing a chunk at a time with the buffer).  This is a problem because it will generate tons of memory garbage and may create demand excessive memory spikes that impact stability.  

Also, streaming APIs are just a more natural fit with our normal logging model if possible.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -149,7 +175,12 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
                         // TODO is this efficient for large amounts of output? Would it be better to stream data, or return a byte[] from the callable?
```
svanoort: I agree with this comment... really really this is the place to stream it.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: It is internal only, but sure a constant could be introduced for clarity.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Well, hardly “massive” as this is typically dwarfed by unrelated overhead related to launching and joining processes, but I can check if there is some premature optimization possible here if you care.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: This is only for `returnStdout: true`, where we are anyway returning a `String` to the program. There is no reason to even attempt to stream anything.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: See the other set of PRs which deprecate this code path anyway.

jglick:
> process a chunk at a time in a streaming fashion

That happens in #60. Prior to that, the _current_ output processing mode of `FileMonitoringController` is not streaming. This PR does not change that.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: Separate paths for optimization - process launch/join has some fixed overheads and some scaling issues that are addressed by changes in the algorithm.  

If we're injecting new APIs though we should ensure the API is friendly to efficient implementations -- coupling directly to newly-allocated-and-not-reused `byte[]` will bite us.

jglick:
So to summarize, if this is in incorporated into #62, which performs streaming transcoding where necessary, then memory buffers are allocated only in case you use the relatively uncommon mode `returnStdout: true` (in which case we are _already_ allocating a `String`, and doing more with it besides—typically saving it to `program.dat`); and even then only when the process was actually using a non-UTF-8 output encoding, which in practice means stuff running on Windows. This does not seem like a valuable target for optimization.

jglick:
sorry @abayer & @rsandell, you cannot be reviewers any more

src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
```diff
@@ -63,4 +65,23 @@ public void captureOutput() throws UnsupportedOperationException {
...
     }
 
+    /**
+     * Requests that a specified charset be used to transcode process output.
+     * The encoding of {@link Controller#writeLog} and {@link Controller#getOutput} is then presumed to be UTF-8.
+     * If not called, no translation is performed.
+     * @param cs the character set in which process output is expected to be
+     */
+    public void charset(@Nonnull Charset cs) {
+        // by default, ignore
```
oleg-nenashev: Add some FINE/DEBUG logging to indicate that the method is not implemented?

src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
```diff
@@ -63,4 +65,23 @@ public void captureOutput() throws UnsupportedOperationException {
...
+    }
+
+    /**
+     * Requests that the node’s system charset be used to transcode process output.
+     * The encoding of {@link Controller#writeLog} and {@link Controller#getOutput} is then presumed to be UTF-8.
+     * If not called, no translation is performed.
+     */
+    public void defaultCharset() {
+        // by default, ignore
```
oleg-nenashev: same

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
oleg-nenashev: +1 to @jglick. It would be reasonable if the operation supported any parameters (e.g. tail or so), but currently the change does not make the code worse

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -236,4 +239,141 @@ private void runOnDocker(DumbSlave s) throws Exception {
         runOnDocker(new DumbSlave("docker", "/home/jenkins/agent", new SimpleCommandLauncher("docker run -i --rm --name agent --init jenkinsci/slave:3.7-1 java -jar /usr/share/jenkins/slave.jar")));
     }
 
+    @Issue("JENKINS-31096")
+    @Test public void encoding() throws Exception {
```
oleg-nenashev: This test is unreadable IMHO. Would it be possible to split it to several tests or at least explicitly indicate test stages? Assert logic could be also refactored to a separate method

src/main/java/org/jenkinsci/plugins/durabletask/DurableTask.java
jglick: Could do that. In practice there are only three implementations anyway, all in this plugin.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
jglick: Yes, it could be refactored to use a helper assertion method with parameters.

jglick:
> There are other durable task implementations

Well, [with a combined total of <200 installations](https://jenkins.io/doc/developer/extensions/durable-task/), and probably those plugins are mistakes that should be deleted, but yes.

> so it would need explicit changelog

Sure.

jglick:
In https://github.com/jenkinsci/jep/pull/128#discussion_r197574441 @svanoort asked what it would take to ensure that newly produced non-ASCII content from durable tasks is correctly transcoded into builds which were started using a non-UTF-8 character set but continued running across the upgrade. AFAICT `FileMonitoringController.transcodingCharset` and `.maybeTranscode` and (in #62) `Watcher.run` would need to know the correct target encoding. For the `Watcher.run`, this information is available from `TaskListener`, but only as of Jenkins 2.91 (https://github.com/jenkinsci/jenkins/pull/3122) and not even then from this context since it is `ProtectedExternally`. For other usages (`writeLog`, `getOutput`), we do not even have a `TaskListener`. So the target encoding would need to be injected explicitly, probably in the `DurableTask.charset` & `.defaultCharset` methods, with the caller (`DurableTaskStep.Execution.start`) looking up the `Run.getCharset` when a `Run` is available. Doable, but makes the code more complicated, and hardly seems worth the bother for a very minor bug affecting a small minority of users and even then at most once.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -125,43 +149,47 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
                         byte[] buf = new byte[(int) toRead];
                         raf.readFully(buf);
-                        sink.write(buf);
-                    } finally {
-                        raf.close();
+                        ByteBuffer transcoded = maybeTranscode(buf, charset);
+                        if (transcoded == null) {
+                            sink.write(buf);
+                        } else {
+                            Channels.newChannel(sink).write(transcoded);
```
jglick: This may be writing a number of bytes to `CountingOutputStream` which is different than `toRead` (when the full write succeeds). Probably the transcoding needs to be done on the master side, which will be a little trickier since any `defaultCharset` call needs to happen on the agent side.

jglick: Note that the bug, if there is one, is rendered obsolete by #62 which renders this dead code.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
jglick: (resolved)

jglick:
Reproduced the suspected problem (some characters missing from output when doing transcoding) in the test and fixed it. I now suspect that #62 suffers from an analogous bug, but that is at least further down the road.

jglick:
> won't this run into issues with external log storage

No, because as mentioned before (https://github.com/jenkinsci/durable-task-plugin/pull/61#discussion_r204948161), the affected code is no longer used (a deprecated code path) when used with external log storage.

jglick:
> test cases for character encoding situations around external log storage

The interesting logic is already covered by tests here; the external logging systems just assume they are getting UTF-8, and keep it that way. Anyway, as we develop “live” tests for such implementations we can include some sanity checks for encoding handling. Certainly it is a far lower priority since any problems would not be regression risks.

PR #63, opened 2018-02-26
avoid bash error

caio2k:
While executing a k8s pipeline the command used to launch a script resulted in error.
`bash: syntax error near unexpected token 'do'`


svanoort:
@caio2k This appears to cause test failures: https://ci.jenkins.io/blue/organizations/jenkins/Plugins%2Fdurable-task-plugin/detail/PR-63/1/tests

Could you please take a look and amend it as needed?  Thanks!

car-roll:

PR #65, opened 2018-03-01
[JENKINS-25519] If unable to read exitStatus file then wait briefly and retry one more time to allow write to complete

svanoort:
Proposed solution to error like this, encountered when statusCode file exists but is empty, potentially when created but not written to yet, or not written fully:

> java.lang.NumberFormatException: For input string: ""
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:592)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.jenkinsci.plugins.durabletask.FileMonitoringTask$FileMonitoringController.exitStatus(FileMonitoringTask.java:168)
> Caused: java.io.IOException: corrupted content in $SOMEPLACE 
> at org.jenkinsci.plugins.durabletask.FileMonitoringTask$FileMonitoringController.exitStatus(FileMonitoringTask.java:170)
	at org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController.exitStatus(BourneShellScript.java:211)

Variant of [JENKINS-25519](https://issues.jenkins-ci.org/browse/JENKINS-25519)

This may not be the *best* solution to the problem in my opinion but should generally solve it.

Other options:
* Return null until the file parses cleanly (normal timeout will apply once the launching script dies and log file no longer is touched).
* Write and look for a delimiter on the exitStatus file to indicate that the file is fully written (avoids theoretical issues with only '1' being written when status code might actually be '102' or something).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -163,11 +163,17 @@ public abstract class FileMonitoringTask extends DurableTask {
         // TODO would be more efficient to allow API to consolidate writeLog with exitStatus (save an RPC call)
         @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
             FilePath status = getResultFile(workspace);
-            if (status.exists()) {
+            if (status.exists() && status.length() > 0) {
                 try {
                     return Integer.parseInt(status.readToString().trim());
                 } catch (NumberFormatException x) {
-                    throw new IOException("corrupted content in " + status + ": " + x, x);
+                    LOGGER.log(Level.FINE, "Unable to parse exit code file, waiting briefly and retrying - file "+status+" error message: "+x.getMessage());
+                    try {
+                        Thread.sleep(100);
+                        return Integer.parseInt(status.readToString().trim());
+                    } catch (NumberFormatException nfe) {
+                        throw new IOException("corrupted content in " + status + ": " + nfe, nfe);
+                    }
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -163,11 +163,17 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
         // TODO would be more efficient to allow API to consolidate writeLog with exitStatus (save an RPC call)
         @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
             FilePath status = getResultFile(workspace);
-            if (status.exists()) {
+            if (status.exists() && status.length() > 0) {
```
jglick: Well this alone would fix the issue, right?

svanoort: Theoretically, yes, if there's no bogus whitespace or other stuff in there.

jglick:
Ah, maybe I was thinking of #37—similar problem (at least on the face of it), different file.

If the problem is indeed a simple race condition, tricks like this are not the best fix. Rather, the wrapper script should be changed from

```sh
…; echo $? > …/exit
```

to

```
…; echo $? > …/exit.tmp; mv …/exit.tmp …/exit
```

which ought to be atomic.

svanoort:
@jglick Has requested a test-comment, please ignore this.  @cloudbees/team-arc

svanoort:
Closing in favor of #66

PR #66, opened 2018-03-03
Better fix to [JENKINS-25519] -- avoid errors when trying to read exitStatus from durable task before finished writing

svanoort:
Superior version of https://github.com/jenkinsci/durable-task-plugin/pull/65

Proposed solution to error like this, encountered when statusCode file exists but is empty, potentially when created but not written to yet, or not written fully.  Was confirmed that previous version removed the errors with the "corrupt" status file content and greatly reduced failure rates.

> java.lang.NumberFormatException: For input string: ""
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:592)
	at java.lang.Integer.parseInt(Integer.java:615)
	at org.jenkinsci.plugins.durabletask.FileMonitoringTask$FileMonitoringController.exitStatus(FileMonitoringTask.java:168)
> Caused: java.io.IOException: corrupted content in $SOMEPLACE 
> at org.jenkinsci.plugins.durabletask.FileMonitoringTask$FileMonitoringController.exitStatus(FileMonitoringTask.java:170)
	at org.jenkinsci.plugins.durabletask.BourneShellScript$ShellController.exitStatus(BourneShellScript.java:211)

Variant of [JENKINS-25519](https://issues.jenkins-ci.org/browse/JENKINS-25519)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -138,9 +138,12 @@ public final class BourneShellScript extends FileMonitoringTask {
         String cmd;
         FilePath logFile = c.getLogFile(ws);
         FilePath resultFile = c.getResultFile(ws);
+        if (resultFile.exists()) {
+            resultFile.delete();  // Maybe overly cautious, but better safe than sorry, similarly we should make sure no duplicate logfile?
+        }
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
-            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s'; wait",
+            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,
@@ -149,9 +152,10 @@ public final class BourneShellScript extends FileMonitoringTask {
                 scriptPath,
                 c.getOutputFile(ws),
                 logFile,
+                resultFile,
                 resultFile);
         } else {
-            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s'; wait",
+            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,
@@ -159,6 +163,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                 cookieVariable,
                 scriptPath,
                 logFile,
+                resultFile,
                 resultFile);
         }
         cmd = cmd.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
@@ -231,7 +236,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                         if (pidFile.exists()) {
                             listener.getLogger().println("still have " + pidFile + " so heartbeat checks unreliable; process may or may not be alive");
                         } else {
-                            listener.getLogger().println("wrapper script does not seem to be touching the log file in " + controlDir);
+                            listener.getLogger().println("s " + controlDir);
                             listener.getLogger().println("(JENKINS-48300: if on a laggy filesystem, consider -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.HEARTBEAT_CHECK_INTERVAL=300)");
                             return recordExitStatus(workspace, -1);
                         }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -152,8 +152,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                 scriptPath,
                 c.getOutputFile(ws),
                 logFile,
-                resultFile,
-                resultFile);
+                resultFile, resultFile, resultFile);
         } else {
             cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
@@ -163,8 +162,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                 cookieVariable,
                 scriptPath,
                 logFile,
-                resultFile,
-                resultFile);
+                resultFile, resultFile, resultFile);
         }
         cmd = cmd.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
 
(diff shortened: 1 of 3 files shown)


svanoort:
@reviewbybees

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -138,9 +138,12 @@ public String getScript() {
         String cmd;
         FilePath logFile = c.getLogFile(ws);
         FilePath resultFile = c.getResultFile(ws);
+        if (resultFile.exists()) {
+            resultFile.delete();  // Maybe overly cautious, but better safe than sorry, similarly we should make sure no duplicate logfile?
```
jglick: Well, the whole control directory should not exist before we start, or should be empty.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,6 +210,7 @@ private FilePath pidFile(FilePath ws) throws IOException, InterruptedException {
             return controlDir(ws).child("pid");
         }
 
+        // TODO run as one big MasterToSlaveCallable<Integer> to avoid extra network roundtrips
```
jglick: Irrelevant after #60.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -231,7 +235,7 @@ private FilePath pidFile(FilePath ws) throws IOException, InterruptedException {
                         if (pidFile.exists()) {
                             listener.getLogger().println("still have " + pidFile + " so heartbeat checks unreliable; process may or may not be alive");
                         } else {
-                            listener.getLogger().println("wrapper script does not seem to be touching the log file in " + controlDir);
+                            listener.getLogger().println("s " + controlDir);
```
jglick: Uh…?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -160,20 +165,39 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
-        // TODO would be more efficient to allow API to consolidate writeLog with exitStatus (save an RPC call)
-        @Override public Integer exitStatus(FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
-            FilePath status = getResultFile(workspace);
-            if (status.exists()) {
-                try {
-                    return Integer.parseInt(status.readToString().trim());
-                } catch (NumberFormatException x) {
-                    throw new IOException("corrupted content in " + status + ": " + x, x);
+        /** Avoids excess round-tripping when reading status file. */
+        static class StatusCheck extends MasterToSlaveFileCallable<Integer> {
```
jglick: This is just gratuitous merge conflict creation against #60. Better to keep the patch short and to the point.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -160,20 +165,39 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
+            @Override
+            @CheckForNull
+            public Integer invoke(File f, VirtualChannel channel) throws IOException, InterruptedException {
+                if (f.exists() && f.length() > 0) {
+                    try {
+                        String fileString = Files.readFirstLine(f, Charset.defaultCharset());
```
jglick: Anyway you could just get rid of the `exists` check and use a single `readToString` call, catching `FileNotFoundException` and ignoring.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -160,20 +165,39 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
+                    } catch (NumberFormatException x) {
+                        throw new IOException("corrupted content in " + f + ": " + x, x);
+                    }
                 }
-            } else {
                 return null;
             }
         }
 
+        static final StatusCheck STATUS_CHECK_INSTANCE = new StatusCheck();
```
jglick: You know `act` is doing object allocation anyway, right?

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -60,17 +60,18 @@ public String getScript() {
         BatchController c = new BatchController(ws);
 
         String cmd;
+        String quotedResultFile = quote(c.getResultFile(ws));
         if (capturingOutput) {
-            cmd = String.format("@echo off \r\ncmd /c \"\"%s\"\" > \"%s\" 2> \"%s\"\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+            cmd = String.format("@echo off \r\ncmd /c \"\"%s\"\" > \"%s\" 2> \"%s\"\r\necho %%ERRORLEVEL%% > \"%s.tmp\"\r\nmove \"%s.tmp\" \"%s\"\r\n",
```
jglick: And PowerShell?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
svanoort: Assuming #60 is fundamentally reliable. We can always remove the comment if it tests clean.  And we may have to do yet another emergency fix due to something introduced in #49 before that gets integrated.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
svanoort: Erm, I have no idea where that came from.  Best guess It was a hotkeying error somewhere, but fixing now.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
svanoort: Powershell is caught by the 0-length check -- I have no idea how we'd even begin applying this to it and was planning to ask @gabloe or @jtnord because they know the Windows side of things. 

Can at least still write a Batch script though, so I included that.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: We have one particular user who has longer latencies and this change likely helps keep them under the timeouts - thus it is hardly "gratuitous."

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
svanoort: If it's all agent-local then it doesn't really matter, does it?  Also: as a rule of thumb in Java one should not be throwing Exceptions to signal routine and expected conditions.  They come with a higher than normal level of baggage. 

As opposed to Python where that's the idiomatic way to signal things like an Iterator being done.

jglick: Well, Remoting itself is a pretty high level of baggage, which would swamp any tiny overhead of `Throwable.fillInStackTrace` in this case.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
jglick: Sensible.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Minor note: use of `defaultCharset` could theoretically cause issues on z/OS. Since we are expecting this file to be ASCII (it should in fact just contain `[0-9]+`), it is safer and clearer to use `StandardCharsets.ASCII`.

svanoort: A really critical point - it looks like this is already broken, per https://github.com/jenkinsci/durable-task-plugin/pull/28 -- I think we'd need to address it in more comprehensive work to get it to play nicely in any case.

PR #67, opened 2018-03-07
[Fixed JENKINS-50029] Fix issues with PowerShell error handling

gabloe:
This PR fixes issues with backwards compatibility of error handling and error propagation, and ensures that non-fatal errors do not erroneously fail a job while also ensuring that $ErrorActionPreference is honored. The previous logic of relying on the $Error automatic variable is incorrect because it disregards the value of $ErrorActionPreference.


svanoort:
As far as I can tell this looks reasonable to me (version checks, testcases, etc), but what I know about Powershell could be written on the back of a single notecard.  @jtnord would you be able to take a quick peek and approve if you think it's releasable?

svanoort:
Thanks @jtnord I'm merging now and will aim to cut a release shortly

PR #68, opened 2018-03-08
[JENKINS-TBD] Fix hung Batch steps

svanoort:
Should fix a regression [JENKINS-50025](https://issues.jenkins-ci.org/browse/JENKINS-50025) introduced in #66 -- a subtle issue that may revolve around a quirk of path handling in Windows. 

This triggers symptoms matching [JENKINS-34150](https://issues.jenkins-ci.org/browse/JENKINS-34150).  

After an hour of investigation with the user (results noted in ticket) and without an ability to easily reproduce the issue, **I am backing out the previous change to the BatchScriptDurableTask.**  

Other components of #66 (the zero-length file check) should adequately ensure Windows coverage for [JENKINS-25519](https://issues.jenkins-ci.org/browse/JENKINS-25519), and I'd rather go back to a known-good version than risk a second regression.  This version was tested with the user and confirmed to work correctly.


svanoort:
Regression inexplicably triggered on some systems by #66

svanoort:
@reviewbybees

PR #69, opened 2018-04-30
Fix test failure under non-english Windows

Wadeck:
If you are using a non-english version of Windows, your PowerShell is configured to run in your language. And so, the test PowershellScriptTest#specialStreams() that will test for the presence of "VERBOSE" will receive String that are dependent on your language.
In french, I see : 
```
COMMENTAIRES : Hello, Verbose!
AVERTISSEMENT : Hello, Warning!
DÉBOGUER : Hello, Debug!
```

instead of the expected

```
VERBOSE: Hello, Verbose!
WARNING: Hello, Warning!
DEBUG: Hello, Debug!
```

By fixing the language before running the command, this will solve this problem (and not impact other test cases). If you are using a english Windows, you can just change the en-US to fr-FR to see the difference.

@reviewbybees

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
@@ -204,7 +204,8 @@ public class PowershellScriptTest {
     }
     
     @Test public void specialStreams() throws Exception {
-        DurableTask task = new PowershellScript("$VerbosePreference = \"Continue\"; " +
+        DurableTask task = new PowershellScript("[Threading.Thread]::CurrentThread.CurrentUICulture = 'en-US'; " +
+                                                "$VerbosePreference = \"Continue\"; " +
                                                 "$WarningPreference = \"Continue\"; " +
                                                 "$DebugPreference = \"Continue\"; " +
                                                 "Write-Verbose \"Hello, Verbose!\"; " +


dwnusbaum:
@gabloe looks like you have been working on some related code, would be great if you could take a look!

gabloe:
I'm not sure if [Threading.Thread]::CurrentThread.CurrentUICulture has any impact on the locale in PowerShell.

> C:\> [Threading.Thread]::CurrentThread.CurrentUICulture = 'fr-FR';
> C:\> Write-Verbose "Test" -Verbose
> VERBOSE: Test

I do not have experience with PowerShell locale settings, but I can try to help.

Wadeck:
@gabloe 

You need to apply the change in Culture in the same line, like 
```powershell
C:> [Threading.Thread]::CurrentThread.CurrentUICulture = 'fr-FR'; Write-Verbose "Test" -Verbose
```
to see an impact.

gabloe:
No, that shouldn't be the case.  And it does not produce the intended result.

> PS C:\windows\system32> [Threading.Thread]::CurrentThread.CurrentUICulture = 'fr-FR'; Write-Verbose "Test" -Verbose
> VERBOSE: Test

Wadeck:
@gabloe you already overcome my skills in PowerShell, cannot help more. 

I just know in my machine with a french Windows, the two separate commands provide french messages and both on the same line provides the english version (I switch fr-FR to en-US)

Do you think it is linked to the fact you do not have the french language pack installed ? (pure hypothesis)

gabloe:
I installed the French language pack before testing the command. I haven't ever played around with language/culture settings, so I am unsure about the proper way to set this up.

gabloe:
I understand the issue here.  The issue is that with the current way that PowerShell is invoked two separate processes are created on the Jenkins host.  The first process executes the second and pipes the resulting stdout/stderr to log files, among other things.  The reason for this is due to a quirk in PowerShell which causes certain output hints to be removed from the console output upon redirection/merging with stdout, but by using a second process that issue is mitigated.  Setting [Threading.Thread]::CurrentThread.CurrentUICulture only has an effect in the parent powershell process, which is why it's not working.

I have sent a PR https://github.com/jenkinsci/durable-task-plugin/pull/70 that simultaneously reduces the PowerShell footprint to a single process, and preserves the output hints for VERBOSE, WARNING, and DEBUG streams, but has the side effect of only supporting English.

car-roll:

PR #70, opened 2018-05-08
[Fixed JENKINS-50840] Reduce PowerShell footprint

gabloe:
Currently the PowerShell pipeline step consumes 2 PowerShell processes on the Jenkins host. The reason for this was because of a quirk which causes redirected console output to lose the appropriate stream designation when merged with stdout, but was mitigated by wrapping execution in a second call to powershell.  This PR introduces a fix that reduces the footprint of PowerShell to a single process, which in turn reduces the Jenkins host load, and artificially injects the appropriate stream designations into the console output.

A side effect of this is that localization/culture of three common PowerShell output hints are lost: VERBOSE, WARNING, and DEBUG. This has some direct impact on the changes introduced in https://github.com/jenkinsci/durable-task-plugin/pull/69, in that on a host that has non-English localization the stream designation hints will be in English instead of the correct culture defined in the host system. I believe that the benefits of reduced load outweigh the cost of losing localization of 3 output hints.

This also fixes JENKINS-50840 by adding special handling to buffer PowerShell output from format-* cmdlets

@reviewbybees


gabloe:
This PR was closed because after further testing it seems this causes more harm than good.  I am researching some better approaches.

PR #72, opened 2018-06-07
Incrementalified

jglick:


src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -49,7 +49,7 @@
      */
     boolean isContinued();
 
-    @Restricted(DoNotUse.class) // implementation
+    @Restricted(NoExternalUse.class) // implementation
```
jglick: https://github.com/kohsuke/access-modifier/pull/12

pom.xml
```diff
@@ -42,13 +44,13 @@
...
         <url>https://github.com/jenkinsci/${project.artifactId}-plugin</url>
-        <tag>HEAD</tag>
+        <tag>${scmTag}</tag>
     </scm>
     <dependencies>
         <dependency>
             <groupId>org.jenkins-ci.test</groupId>
             <artifactId>docker-fixtures</artifactId>
-            <version>1.6</version>
+            <version>1.8</version>
```
oleg-nenashev: JTH still does not bundle it iiuc

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
jglick: or perhaps not

pom.xml
jglick: No, it is a separate dependency.

src/test/resources/org/jenkinsci/plugins/durabletask/CentOSFixture/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM centos:7.4.1708
 RUN yum -y install \
-        openssh-server-7.4p1-13.el7_4 \
-        java-1.8.0-openjdk-headless-1.8.0.151-1.b12.el7_4 \
+        openssh-server \
```
svanoort: @jglick Why?

jglick: See https://github.com/jenkinsci/durable-task-plugin/pull/72/commits/621cfbef0232e176059e834fa465d8a3fcff8272 and a long series of PRs in `docker-fixtures`.

PR #74, opened 2018-06-25
[JENKINS-37575] Keep track of how much content has been copied by writeLog even if the callable is interrupted

jglick:
See [JENKINS-37575](https://issues.jenkins-ci.org/browse/JENKINS-37575) where there is information to reproduce. This is dead code as of #60 + https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/63 but may as well fix the bug so long as the old code path can be run at least with a kill switch.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -126,7 +126,6 @@ public abstract class FileMonitoringTask extends DurableTask {
         @Override public final boolean writeLog(FilePath workspace, OutputStream sink) throws IOException, InterruptedException {
             FilePath log = getLogFile(workspace);
             CountingOutputStream cos = new CountingOutputStream(sink);
-            long start = System.currentTimeMillis();
             try {
                 log.act(new WriteLog(lastLocation, new RemoteOutputStream(cos)));
                 return cos.getByteCount() > 0;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -50,7 +50,7 @@ import java.util.logging.Level;
 import java.util.logging.Logger;
 import jenkins.MasterToSlaveFileCallable;
 import org.apache.commons.io.IOUtils;
-import org.jenkinsci.remoting.RoleChecker;
+import org.apache.commons.io.output.CountingOutputStream;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -125,43 +125,40 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
                 if (len > lastLocation) {
-                    RandomAccessFile raf = new RandomAccessFile(f, "r");
-                    try {
+                    try (RandomAccessFile raf = new RandomAccessFile(f, "r")) {
                         raf.seek(lastLocation);
                         long toRead = len - lastLocation;
                         if (toRead > Integer.MAX_VALUE) { // >2Gb of output at once is unlikely
                             throw new IOException("large reads not yet implemented");
                         }
-                        // TODO is this efficient for large amounts of output? Would it be better to stream data, or return a byte[] from the callable?
```
svanoort: Why is this line removed?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -125,43 +125,40 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
...
                         byte[] buf = new byte[(int) toRead];
```
svanoort: Still not a fan of the fact that we chose to do a complete buffer read (since potentially we could have a very large block of data).  I'm itching to change this once these logging PRs are all integrated.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: Because the speculation is obsolete—the PR is picking another approach entirely.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: #60 (when active) renders this dead code, replaced by something that (potentially) streams content.

jglick:
`writeLog` counts bytes. If you think there is a bug in #61, ask it there.

recampbell:
@jglick  What is the next step for this PR?

jglick:
@recampbell I am awaiting reviews.

svanoort:
I needed to double check yet another interaction with #61 to make sure this wouldn't need amendment with the changes there but I think we're fine now.

PR #75, opened 2018-06-27
[JENKINS-50892] Ensure the toucher process exits if the main wrapper process is killed

jglick:
To the extent that I follow the analyses in [JENKINS-50892](https://issues.jenkins-ci.org/browse/JENKINS-50892) and [JENKINS-46283](https://issues.jenkins-ci.org/browse/JENKINS-46283) I suspect this will fix the problem. Something is still fishy—under some circumstances, something is killing the main wrapper process (the one whose `pid` is recorded here) and I do not know why or how to reproduce that.

Thanks to [this tip](https://stackoverflow.com/a/14880926/12916) for the liveness checking trick. Someone with a Mac ought to check that this works on OS X.

Note that superficially this looks to restore some stuff I deleted in #49. The difference here is that the PID checks are being done in an external process (so, no problems with JNA), and from a process in the same kernel namespace (so, no problems with `withDockerContainer` and its K8S equivalent I hope).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -140,7 +140,7 @@ public final class BourneShellScript extends FileMonitoringTask {
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
-            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid -p $pid | grep -q $pid && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,
@@ -151,7 +151,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                 logFile,
                 resultFile, resultFile, resultFile);
         } else {
-            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid -p $pid | grep -q $pid && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,


svanoort:
@dwnusbaum Can you give this one a look?

svanoort:
(I'll also need to review as well, but this probably needs several set of eyes)

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -140,7 +140,7 @@ public String getScript() {
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
-            cmd = String.format("{ while [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid -p $pid | grep -q $pid && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
```
dwnusbaum: Would it be better to use `grep -q "^$pid$"` in case the pid is a substring of some other valid pid on the agent? (Same for below as well)

dwnusbaum:
Tested the change locally and it seems to work fine for me on OS X. I `kill -15`'d the parent process for a long-running shell step, and the child process touching the file exited immediately while script.sh kept running. When the shell step finished, it hung for a few seconds and then exited with "wrapper script does not seem to be touching the log file in ...".

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
oeuftete: This apparently just caused a problem for me in the wild:

```
15:25:26 ps: unrecognized option: p
15:25:26 BusyBox v1.27.2 (2017-12-12 10:41:50 GMT) multi-call binary.
15:25:26 
15:25:26 Usage: ps [-o COL1,COL2=HEADER]
15:25:26 
15:25:26 Show list of processes
15:25:26 
15:25:26 	-o COL1,COL2=HEADER	Select columns for display
```

Had to revert to the previous version.

dwnusbaum: @oeuftete Best to open a bug report on https://issues.jenkins-ci.org/secure/Dashboard.jspa. It would be great if you could include the output of  `ps -o pid`, `ps -o pid=`, and `ps -o pid=HEADER` on your system.

svanoort: @oeuftete Also, please include your operating system and version for the **build agent** in that ticket.

dwnusbaum: Looks like [BusyBox's ps](https://busybox.net/downloads/BusyBox.html?#ps) only has `-o` and `-T` options, so we'd have to omit the `-p` option and pass a stricter pattern to `grep` for this to work with that version of `ps`.

Notably, Alpine Linux's Official Docker image only has BusyBox's `ps` installed by default, see gliderlabs/docker-alpine#173, so we probably want to fix it.

carlossg: jenkins-slave image is adding `procps` in https://github.com/jenkinsci/docker-slave/pull/28

carlossg: https://issues.jenkins-ci.org/browse/JENKINS-52847

PR #77, opened 2018-08-01
[JENKINS-52847] Avoid using -p to support basic implementations of ps

dwnusbaum:
See [JENKINS-52847](https://issues.jenkins-ci.org/browse/JENKINS-52847) and  https://github.com/jenkinsci/durable-task-plugin/pull/75#discussion_r206994271.

Some implementations of `ps` don't support the `-p` option, but we can work around that by passing a more complicated regex to `grep`.

I manually tested the change on OS X and Alpine Linux to make sure it works on both platforms.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -140,7 +140,7 @@ public final class BourneShellScript extends FileMonitoringTask {
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
-            cmd = String.format("pid=$$; { while ps -o pid -p $pid | grep -q $pid && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid | grep -q \"^$pid$\" && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,
@@ -151,7 +151,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                 logFile,
                 resultFile, resultFile, resultFile);
         } else {
-            cmd = String.format("pid=$$; { while ps -o pid -p $pid | grep -q $pid && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid | grep -q \"^$pid$\" && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -140,7 +140,7 @@ public final class BourneShellScript extends FileMonitoringTask {
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
-            cmd = String.format("pid=$$; { while ps -o pid | grep -q \"^$pid$\" && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid | grep -q \"^\\s*$pid$\" && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
                 controlDir,
                 resultFile,
                 logFile,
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -140,7 +140,7 @@ public String getScript() {
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
-            cmd = String.format("pid=$$; { while ps -o pid -p $pid | grep -q $pid && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
+            cmd = String.format("pid=$$; { while ps -o pid | grep -q \"^\\s*$pid$\" && [ -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
```
dwnusbaum: To get this to work on csh/tcsh we'd need to use `\"^\\s*$pid\"\'$\'` (becomes `"^\s*$pid"'$'` after removing Java's escaping) because in those shells things like `echo "$"` throw an error.

Do we care about csh/tcsh?

svanoort: I don't *think* so - @jglick might have other thoughts. 

But is *grep* guaranteed to be supported?

svanoort: This is making me twitch a bit.  Which docker images did we test with?

dwnusbaum: No idea about whether it's _guaranteed_ to be supported on all platforms, but it is at least included in BusyBox. I'll look around for a definitive answer.

dwnusbaum: > Which docker images did we test with?

jenkinsci/slave with both alpine (3.19-1-alpine, before the procops fix) and non-alpine (3.20-1) tags.

dwnusbaum: > is grep guaranteed to be supported

Someone could always have a custom arch distro that specifically excludes grep, but I think it is pretty unlikely. The best info I could find about where `grep` is included by default is https://unix.stackexchange.com/questions/37064/which-are-the-standard-commands-available-in-every-linux-based-distribution. Based on that, and the fact that our use is compatible with BusyBox as well, I think it is pretty safe to count on grep being available, unless there are some new popular container OS's that are leaving it out entirely.

jglick:
[Oh my](https://github.com/brgl/busybox/blob/abbf17abccbf832365d9acf1c280369ba7d5f8b2/procps/ps.c#L615-L641):

> We allow (and ignore) most of the above. FIXME.

See also [this](https://stackoverflow.com/a/51711064/12916) for example.

Extend `BourneShellScriptTest` to use a Docker fixture based on `jenkinsci/slave:3.19-1-alpine` to reproduce the problem before the fix please.

In the longer term, the current implementation using shell scripting should probably be discarded in favor of a Golang wrapper that could also do a `setsid`. See discussion in [JENKINS-25503](https://issues.jenkins-ci.org/browse/JENKINS-25503). Of course this means shipping binaries for each supported platform, or shipping only an x86_64 binary and falling back to the current implementation on exotic platforms.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -220,7 +232,7 @@ private void runOnDocker(DumbSlave s) throws Exception {
         do {
             Thread.sleep(1000);
             baos = new ByteArrayOutputStream();
-            assertEquals(0, dockerLauncher.launch().cmds("ps", "-e", "-o", "pid,stat,command").stdout(new TeeOutputStream(baos, System.out)).join());
+            assertEquals(0, dockerLauncher.launch().cmds("ps", "-e", "-o", "pid,stat,comm").stdout(new TeeOutputStream(baos, System.out)).join());
```
dwnusbaum: `comm` appears to be the standardized name: http://pubs.opengroup.org/onlinepubs/009695399/utilities/ps.html, and BusyBox doesn't support `command`.

jglick:
> why not simply use the Java 8 Process tooling

Off topic here; as mentioned in JIRA, we probably cannot use Java code for that.

jglick:
BTW things like e015167 do not actually need to be committed, much less pushed. Suffices to _locally_

```bash
git checkout master -- src/main
```

develop your test until it fails in the expected way, then

```bash
git checkout HEAD -- src/main
```

and verify that it passes, then commit the new test. We trust you! Alternately, to be explicit in history:

```bash
git checkout -b tmp master
# edit src/test/ until you get the expected failure
# add @Ignore
git commit -a -m 'Reproduced failure'
git checkout support-basic-ps
git merge --no-commit tmp
# remove @Ignore
git commit -a -m 'Yup it is fixed'
git branch -d tmp
```

or the `git rebase -i` equivalent if you like that sort of thing.

jglick:
Fails I guess the way you intended:

```
java.lang.AssertionError: expected:<0> but was:<-1>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:645)
	at org.junit.Assert.assertEquals(Assert.java:631)
	at org.jenkinsci.plugins.durabletask.BourneShellScriptTest.runOnDocker(BourneShellScriptTest.java:233)
	at org.jenkinsci.plugins.durabletask.BourneShellScriptTest.runOnAlpineDocker(BourneShellScriptTest.java:214)
…
[workspace] Running shell script
ps: unrecognized option: p
BusyBox v1.27.2 (2017-12-12 10:41:50 GMT) multi-call binary.

Usage: ps [-o COL1,COL2=HEADER]

Show list of processes

	-o COL1,COL2=HEADER	Select columns for display
wrapper script does not seem to be touching the log file in /home/test/workspace@tmp/durable-e62978d8
…
```

So, time to restore fix?

dwnusbaum:
> and verify that it passes, then commit the new test. We trust you! 

@jglick Ack, in the past I have been asked to show the failing test on the public CI history, but if everyone here is ok with it then it's certainly easier to just confirm locally. In any case I will squash-merge into master when ready here to get rid of the intermediate commits. The time between the commits was due to me running the tests locally to make sure everything passed.

jglick:
> in the past I have been asked to show the failing test on the public CI history

At least for me it is fine if the PR author adds, say, a line comment to the line of the test where an error is thrown in the expected way with the main patch reverted, and shows the error output.

The CI history is actually _less_ useful since the branch project will normally be deleted once the PR is merged, and with it goes any record of the test output. (To be solved at some point by external test storage!)

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -203,12 +205,22 @@ public void smokeTest() throws Exception {
         runOnDocker(new DumbSlave("docker", "/home/test", new SSHLauncher(container.ipBound(22), container.port(22), "test", "test", "", "")));
     }
 
+    @Test public void runOnAlpineDocker() throws Exception {
+        AlpineFixture container = dockerAlpine.get();
+        runOnDocker(new DumbSlave("docker", "/home/test", new SSHLauncher(container.ipBound(22), container.port(22), "test", "test", "", "")), 45);
```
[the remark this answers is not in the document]
dwnusbaum: If it's lower than ~30 seconds or so, the exit status is 0 and everything works fine. 45 seconds seemed to be long enough to consistently hit [this code](https://github.com/jenkinsci/durable-task-plugin/blob/e1ef73b3ab090c8f4bc2b7834b363e0830b6f917/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java#L235) which triggers the failure.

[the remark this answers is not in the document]
jglick: Probably `HEARTBEAT_CHECK_INTERVAL * 2` or something. OK.

PR #79, opened 2018-08-08
[JENKINS-52881] Avoid use of ps

jglick:
[JENKINS-52881](https://issues.jenkins-ci.org/browse/JENKINS-52881)

Amends #75 + #77, which do not work when the system `ps` fails to support even the most basic standard options (apparently there are a lot of nonstandard forks out there), or does not exist at all. Use procfs when that is available (which should also be a bit more efficient); if not; the JENKINS-50892 fix is just skipped. (I guess—we never managed to write a test case for that bug.)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -140,7 +140,7 @@ public final class BourneShellScript extends FileMonitoringTask {
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
         if (capturingOutput) {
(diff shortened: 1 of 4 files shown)


jglick:
> add a test for that behavior if possible

@lvotypko’s https://github.com/lvotypko/durable-task-plugin/commit/a21ed041f67431a18d106760e244e0e8b5c94cc3 purports to do something like that (would have been in #75); I can try to polish that up as a follow-up.

svanoort:
Manual testing is enough to satisfy me at a trivial level -- since this is a SEV-1 issue, we need to move as fast as possible.  If we have any further issues around this though I'm going to insist we do a full revert and add more robust testing -- and still think that would be advantageous.

PR #82, opened 2018-10-01
[JENKINS-50902] Do not rely on shebang for sh invocation

olivergondza:
https://issues.jenkins-ci.org/browse/JENKINS-50902

Current implementation has several limitations:

- Does not support executable paths relative to `$PATH` variable as Shell step does, therefore making it hard for pipeline and freestyle to coexist.
- Reduces portability as the shebang expects absolute path making it hard for windows and unix nodes to coexist.
- Shebang does support interpreters with whitespace in its name, though that is a trivial one.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -111,16 +111,16 @@ public final class BourneShellScript extends FileMonitoringTask {
 
         FilePath shf = c.getScriptFile(ws);
 
-        String s = script, scriptPath;
-        final Jenkins jenkins = Jenkins.getInstance();
-        if (!s.startsWith("#!") && jenkins != null) {
-            String defaultShell = jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel());
-            s = "#!"+defaultShell+" -xe\n" + s;
-        }
-        shf.write(s, "UTF-8");
+        shf.write(script, "UTF-8");
         shf.chmod(0755);
 
-        scriptPath = shf.getRemote();
+        final Jenkins jenkins = Jenkins.getInstance();
+        String interpreter = !script.startsWith("#!") && jenkins != null
+                ? "'" + jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel()) + "' -xe "
+                : ""
+        ;
+
+        String scriptPath = shf.getRemote();
         List<String> args = new ArrayList<>();
 
         OsType os = ws.act(new getOsType());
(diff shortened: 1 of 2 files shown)


olivergondza:
CCing @jenkinsci/code-reviewers to verify we have not overlooked something. Based on internal testing, this does resolve the problem.

dwnusbaum:
@olivergondza Sorry for not responding earlier. It's on my radar and I will try to take a look soon.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -111,16 +111,16 @@ public String getScript() {
...
 
-        String s = script, scriptPath;
-        final Jenkins jenkins = Jenkins.getInstance();
-        if (!s.startsWith("#!") && jenkins != null) {
-            String defaultShell = jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel());
-            s = "#!"+defaultShell+" -xe\n" + s;
-        }
-        shf.write(s, "UTF-8");
+        shf.write(script, "UTF-8");
         shf.chmod(0755);
```
jglick: This should then be done conditionally: only need `a+x` if we are using a shebang.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -111,16 +111,16 @@ public String getScript() {
...
 
-        scriptPath = shf.getRemote();
+        final Jenkins jenkins = Jenkins.getInstance();
+        String interpreter = !script.startsWith("#!") && jenkins != null
```
jglick: You can assume that `jenkins != null` here.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -111,16 +111,16 @@ public String getScript() {
...
+                ? "'" + jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel()) + "' -xe "
```
jglick: `getDescriptorOrDie` is what you want IIRC

olivergondza:
> ... setting the default shell is a terrible idea which IMO should be deleted from Jenkins core

I see your point though that fact shebang accept absolute path only is limiting when creating portable pipelines.

olivergondza:
Any other thoughts to incorporate?

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -245,37 +235,90 @@ public void smokeTest() throws Exception {
...
         assertThat(baos.toString(), containsString("Hello, world!"));
         c.cleanup(ws);
     }
 
+    @Issue("JENKINS-50902")
+    @Test public void configuredInterpreter() throws Exception {
+        // Run script inside the container as this is system dependent
+        DumbSlave node = createDockerSlave(dockerSlim.get());
+        j.jenkins.addNode(node);
+        node.toComputer().waitUntilOnline();
```
dwnusbaum: nit: I think `JenkinsRule#waitOnline(Slave)` would return better error messages in the case of failure, but the behavior should be the same, so not important.

dwnusbaum:
CI picked up the close event, but not the reopen event. I ran the tests locally and they passed for me.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
olivergondza: Good point, addressed.

olivergondza:
@dwnusbaum, test passing. Can you please have another look?

dwnusbaum:
Thanks for the PR @olivergondza, sorry for the slow review cycles!

PR #83, opened 2018-10-18
Removing noise from start of step

jglick:
There is no compelling reason to print these messages for every `sh` etc. step. The working directory is generally obvious from the script, and there is no other new information since the overall log or Blue Ocean or whatever indicates the nature of the step being run. I am not really sure why I ever added this message to begin with.


svanoort:
Yeah, okay, seems reasonable to me

PR #84, opened 2018-10-19
Use the standard Timer rather than our own ScheduledExecutorService

jglick:
Analogous to https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/81 but this time reverting a bit of complexity in #60. Together they seem to solve the `RestartableJenkinsRule` problems in https://github.com/jenkinsci/ant-plugin/pull/32. `Timer` should work fine in agent JVMs (it does not reference Jenkins model classes); the difference is for durable tasks (in watch mode) run on the master node: `Timer` gets properly shut down when Jenkins exits, and is thus safe for in-process restart tests. Again unlikely to have any meaningful impact in production.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -31,8 +31,6 @@ import hudson.Launcher;
 import hudson.Util;
 import hudson.model.TaskListener;
 import hudson.remoting.Channel;
-import hudson.remoting.DaemonThreadFactory;
-import hudson.remoting.NamingThreadFactory;
 import hudson.remoting.RemoteOutputStream;
 import hudson.remoting.VirtualChannel;
 import hudson.slaves.WorkspaceList;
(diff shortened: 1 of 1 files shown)


jglick:
> I'd hesitate to trim back the thread pools that we have created to protect as Bulkheads to protect the (heavily used) Jenkins Timer thread pool from misbehaving Pipeline work

I think you misunderstand. This is watch-mode code, running in the agent JVM. (Assuming you are running the `node` block on an agent. If you are not, who cares about scalability?) So this is not the same `Timer` pool.

jglick:
CI flakiness…

car-roll (approved):

jglick:
BTW @car-roll this had no label, so appears without category in https://github.com/jenkinsci/durable-task-plugin/releases/tag/493.v195aefbb0ff2. Need to make sure to apply one before merging. There is probably some GH Action out there which enforces this…if you can find one which works reliably, is easy to configure, and works properly with forks (should I guess just add a failing Check if there is no label?), then please suggest an addition to the recommended workflow skeleton for JEP-229 in https://github.com/jenkinsci/.github/tree/master/workflow-templates (CC @timja).

PR #73, opened 2018-06-14
powershell on linux use pwsh instead of powershell

scorpionlion:

PR #88, opened 2018-11-27
Added powershell/pwsh detection [JENKINS-48803]

orthoxerox:
Fixes [JENKINS-48803](https://issues.jenkins-ci.org/browse/JENKINS-48803) by detecting whether the system has `pwsh` or `powershell` at runtime. Prefers the former to the latter. Expects the system that presents itself as Unix-compatible to have `which`.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -83,8 +83,7 @@ public final class PowershellScript extends FileMonitoringTask {
                 quote(c.getResultFile(ws)));
         }
       
-        // Note: PowerShell core is now named pwsh. Workaround this issue on *nix systems by creating a symlink that maps 'powershell' to 'pwsh'.
-        String powershellBinary = "powershell";
+        String powershellBinary = findPowershellBinary(launcher);
         String powershellArgs;
         if (launcher.isUnix()) {
             powershellArgs = "-NoProfile -NonInteractive";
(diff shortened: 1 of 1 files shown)


darxriggs:
The pull-requests #55 and #73 also cover this topic.

gabloe:
@darxriggs This gives preference to pwsh, and in Windows systems it is possible to have both Open PowerShell (pwsh) and Windows PowerShell (powershell) side by side.  What would be great is for the preference to be given to the user as a pipeline step parameter.  E.g. something like:

powershell script: '''
   Write-Host 'Foobar'
''', binaryPreference: 'pwsh'

Would require some changes in https://github.com/jenkinsci/workflow-durable-task-step-plugin and since this is common across powershell, windows batch, and bourne shell it would need to be generic.  Just my two cents :)

orthoxerox:
Hm, maybe it would be a better idea to have two separate steps, pwsh and powershell?

gabloe:
> Hm, maybe it would be a better idea to have two separate steps, pwsh and powershell?

That's also a good solution, and less complicated.

jakauppila:
Separate steps definitely makes it a lot clearer for the end user, especially on Windows; that would be my preference.

jakauppila:
Any movement on this? It'd be great to start adopting Powershell Core to ease the transition of our Jenkins servers to Linux.

jakauppila:
I would argue that a choice is required, there are some behavioral differences between Powershell and Pwsh that should require a deliberate decision by the user rather than the simple existence of it on the server.

If two separate steps is too much to maintain than the `binaryPreference` option would probably be the next best option.

casz:
Trying out a fix to both durable-task and workflow-durable-task-step, hopefully it is simple to maintain 😓

casz:
#111 is my attempt at the same thing with a docker fixture test even.
I'll go ahead and submit a downstream PR to workflow-durable-task-step

PR #89, opened 2019-01-08
always force an interpreter

witokondoria:
Pipeline jobs including an sh step are including a shebang on the underlying script. Such shebang caused the script to be mode changed to executable. These "executables", when written on a mounted filesystem with the noexec flag, become non-executable (causing a not obvious failure).

Replacing the bare executable script with a call to the runner defined at the shebang, avoids this situation

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -119,7 +119,7 @@ public final class BourneShellScript extends FileMonitoringTask {
             String shell = jenkins.getDescriptorByType(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel());
             interpreter = "'" + shell + "' -xe ";
         } else {
-            shf.chmod(0755);
+            interpreter = script.split("[\\r\\n]")[0].replaceAll("#!", '');
         }
 
         String scriptPath = shf.getRemote();


dwnusbaum:
Thanks for the PR @witokondoria!

We try to track nontrivial changes through Jira issues, so would you be able to open an issue on https://issues.jenkins-ci.org/secure/Dashboard.jspa with component `durable-task`, and more or less copy the PR description to the ticket to explain the problem and your solution?

Is the scenario you mention something that can be reproduced in a test using a Docker container? For example, we previously had errors on Alpine, and so added some Alpine-specific tests (e.g. [runOnAlpineDocker](https://github.com/jenkinsci/durable-task-plugin/blob/d6258dd63384de5ec261e7374028fd6f4a106dd0/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java#L309) uses [this](https://github.com/jenkinsci/durable-task-plugin/blob/d6258dd63384de5ec261e7374028fd6f4a106dd0/src/test/resources/org/jenkinsci/plugins/durabletask/AlpineFixture/Dockerfile) Dockerfile so we don't accidentally regress the behavior). Given that seemingly minor changes in this plugin can have surprising effects on some platforms, it is best to add tests whenever possible to avoid regressions.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -119,7 +119,7 @@ public String getScript() {
             String shell = jenkins.getDescriptorByType(Shell.DescriptorImpl.class).getShellOrDefault(ws.getChannel());
             interpreter = "'" + shell + "' -xe ";
         } else {
-            shf.chmod(0755);
+            interpreter = script.split("[\\r\\n]")[0].replaceAll("#!", '');
```
dwnusbaum: The empty character literal is causing a compilation error.
```suggestion
            interpreter = script.split("[\\r\\n]")[0].replaceAll("#!", "");
```

dwnusbaum:
FWIW I was a little concerned that shebang lines like `#!/usr/bin/env python3` might break with this, but I tested it and they work fine before and after.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Fixed by 75bdfb1

jglick:
> Perhaps just a flaky test.

Maybe, but should be investigated before merging.

jglick:
> I was a little concerned that shebang lines like `#!/usr/bin/env python3` might break with this, but I tested it and they work

Sounds like that should be made into a test case, since it is not immediately obvious from reading the code.

witokondoria:
> > I was a little concerned that shebang lines like `#!/usr/bin/env python3` might break with this, but I tested it and they work
> 
> Sounds like that should be made into a test case, since it is not immediately obvious from reading the code.

Will try to include such tests

dwnusbaum:
@witokondoria Do you think you will have any time to look into adding some tests?

PR #91, opened 2019-05-07
Added additional testing inside docker-based unit tests

car-roll:
Because durable-task is platform-sensitive, it may be beneficial to have additional testing in the different platform tests (ie runOnUbuntu, runOnCentOS, etc) as opposed to just local system tests. In this particular case, the stop, output, clean workspace, and watch tests were added (smoke tests was already present).

cc @dwnusbaum 
cc @jglick

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
@@ -97,31 +97,40 @@ public class BourneShellScriptTest {
 
     @Test
     public void smokeTest() throws Exception {
-        Controller c = new BourneShellScript("echo hello world").launch(new EnvVars(), ws, launcher, listener);
-        awaitCompletion(c);
+        smokeTest(ws, launcher, 0);
+    }
+
+    public void smokeTest(FilePath testWs, Launcher testLauncher, int sleepSeconds) throws Exception {
+        String script = String.format("echo hello world; sleep %s", sleepSeconds);
+        Controller c = new BourneShellScript(script).launch(new EnvVars(), testWs, testLauncher, listener);
+        awaitCompletion(c, testWs, testLauncher);
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        c.writeLog(ws,baos);
-        assertEquals(0, c.exitStatus(ws, launcher, listener).intValue());
+        c.writeLog(testWs,baos);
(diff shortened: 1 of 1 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -328,18 +353,21 @@ private void runOnDocker(DumbSlave s) throws Exception {
...
-            Thread.sleep(100);
-        }
-        ByteArrayOutputStream baos = new ByteArrayOutputStream();
-        c.writeLog(dockerWS, baos);
-        assertEquals(0, c.exitStatus(dockerWS, dockerLauncher, listener).intValue());
-        assertTrue(baos.toString().contains("hello world"));
-        c.cleanup(dockerWS);
+
+        System.out.println("\n--smoke 10--");
+        smokeTest(dockerWs, dockerLauncher, sleepSeconds);
```
dwnusbaum: Rather than running these all back-to-back in the same test, can we use [parameterized tests](https://github.com/junit-team/junit4/wiki/Parameterized-tests) so they run independently or at least [`ErrorCollector`](https://junit.org/junit4/javadoc/4.12/org/junit/rules/ErrorCollector.html) so that the test would run to completion and report all failures.

What I had in mind for parameterized tests would be to use something like an enum type to hold each platform (e.g. `NATIVE, ALPINE, CENTOS, UBUNTU, ...`), and then create a method `prepareAgentForPlatform(Platform p)` that just sets up an agent and adds it to Jenkins using the relevant Docker container (or `j.createOnlineSlave()` for `NATIVE`), returns some value containing the agent and things like the workspace and launcher (unless they can be computed using just the agent), and then the tests would call that method with the parameterized value and we can delete the `runOnFoo` tests and `runOnDocker`.

car-roll:
@jglick As I understand it, the `runWithCommandLauncher` and `runWIthTiniCommandLauncher` tests were really more to show that we are reaping zombies even without using tini?
It looks like we are always passing --init to docker run, so maybe we can just omit these tests now?
I wasn't sure how to incorporate it into this parameterized version, so I just added those two environments as platforms. I also included the zombie check after the smoketest.

car-roll:
Discovered that the tests introduced here will actually fail with the current version of duruable-task.  Apparently the reboot test shows that it is leaving zombies in the non-init version of the docker simple command launcher tests.

Will move these unit tests to [PR-92](https://github.com/jenkinsci/durable-task-plugin/pull/92) as I have been using theses tests for that effort

PR #92, opened 2019-05-08
Ported script launching and file toucher process to Golang binary

car-roll:
This is an attempt to harden durable-task. The changes are only for unix-based architectures.  Other architectures will be business as usual.
The most notable difference is that the shell script will is now launched via a golang binary. The binary allows the script to have its own session id.  This plus some additional signal catching should increase the survivability of long-running scripts when the jenkins terminates unexpectedly.  The use of golang also removes the need to poll `proc` for the executing shell (for systems where `proc` is available). Building this plugin now requires the golang docker image.

https://issues.jenkins-ci.org/browse/JENKINS-25503
https://issues.jenkins-ci.org/browse/JENKINS-55308

In this version, durable-task will send a copy of a statically compiled Golang binary (approx 2.5mb) from its jar to the slave agent.  ***The binary consists of 3 goroutines:***

1. launcher: The first goroutine is responsible for launching the script as a new process with a new session id. It is also responsible for writing out the exit code to the result file and signaling the heartbeat goroutine that the script has finished.
2. heartbeat: The second goroutine is responsible for updating the modified timestamp of the log file periodically. It will only exit when it has received notification that the script has completed from the launcher goroutine.
3. sigcatcher: The third goroutine is a signal catcher, preventing the binary from being terminated by jenkins until the launched script itself is terminated.


src/main/resources/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,147 @@
...
+		// capturing output
+		scriptCmd.Stderr = logFile
+	} else {
+		// Note: pointing to os.Stdout will not capture all err logs and fail unit tests
+		scriptCmd.Stderr = scriptCmd.Stdout
+	}
+	// Prevents script from being terminated if program gets terminated
+	scriptCmd.SysProcAttr = &syscall.SysProcAttr{Setsid: true}
+	// Allows child processes of the script to be killed if kill signal sent to script's process group id
+	scriptCmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
```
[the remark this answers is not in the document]
car-roll: Good catch, yeah, i'll have to do a separate check on windows when building it. Oddly enough, the script would survive when i killed the main program still, so maybe that's because of setpgid behavior.

src/main/resources/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,147 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
```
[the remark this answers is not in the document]
car-roll: So that's what i thought at first, but i came across something in a google groups mentioning that syscall is actually being frozen and now new changes are being done to it and that's it's poor documentation.  But i'll actually go look for the comment to see if it's an officially unofficial thing.  For what it's worth, in the code, the syscall functions do pointo the to x/sys/ARCH code.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,6 +181,36 @@ public String getScript() {
...
 
+    private String generateLaunchCmd(Jenkins jenkins, ShellController c, FilePath ws, OsType os, String interpreter, String scriptPath, String cookieValue, String cookieVariable) throws IOException, InterruptedException {
+        String logFile = c.getLogFile(ws).toURI().getRawPath();
+        String resultFile = c.getResultFile(ws).toURI().getRawPath();
+        FilePath controlDir = c.controlDir(ws);
+        String outputFile = c.getOutputFile(ws).toURI().getRawPath();
+
+        String launcher_binary = LAUNCHER_PREFIX + os.toString();
+
+        InputStream launcherStream = this.getClass().getResourceAsStream(launcher_binary);
```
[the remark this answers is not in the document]
car-roll: i'll go ahead and do the docker route, generating the binaries at build time.  Right now, it looks like i'll need to do 3 versions (assuming i can get the windows version to work). probably take up just under 10 megs total for all 3.

src/main/resources/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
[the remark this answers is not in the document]
car-roll: fyi, apparently setting both pgid and setsid cause a panic

car-roll:
Checked this against the kubernetes-plugin unit tests locally. Did not see any failures, but again, those were local unit tests.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,6 +188,68 @@ public String getScript() {
...
 
+    private String binaryLauncherCmd(ShellController c, FilePath ws, OsType os, String interpreter, String scriptPath, String cookieValue, String cookieVariable) throws IOException, InterruptedException {
+        String logFile = c.getLogFile(ws).toURI().getRawPath();
+        String resultFile = c.getResultFile(ws).toURI().getRawPath();
+        FilePath controlDir = c.controlDir(ws);
+        String outputFile = c.getOutputFile(ws).toURI().getRawPath();
+
+        String launcher_binary = LAUNCHER_PREFIX + os.toString();
+
+        InputStream launcherStream = this.getClass().getResourceAsStream(launcher_binary);
```
jglick: Tests failing with

```
java.lang.NullPointerException
	at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1792)
	at org.apache.commons.io.IOUtils.copyLarge(IOUtils.java:1769)
	at org.apache.commons.io.IOUtils.copy(IOUtils.java:1744)
	at hudson.FilePath.copyFrom(FilePath.java:913)
	at org.jenkinsci.plugins.durabletask.BourneShellScript.binaryLauncherCmd(BourneShellScript.java:201)
	at org.jenkinsci.plugins.durabletask.BourneShellScript.launchWithCookie(BourneShellScript.java:161)
	at org.jenkinsci.plugins.durabletask.FileMonitoringTask.launch(FileMonitoringTask.java:99)
	at org.jenkinsci.plugins.durabletask.BourneShellScriptTest.smokeTest(BourneShellScriptTest.java:99)
```

```suggestion
        InputStream launcherStream = this.getClass().getResourceAsStream(launcher_binary);
        if (launcherStream == null) {
            throw new IOException("Could not find a launcher for this OS: " + launcher_binary);
        }
```

Though better would be to fall back to a simplified version of the original shell script wrapper, as requested in https://github.com/jenkinsci/durable-task-plugin/pull/92#discussion_r282082078, meaning we do this null check instead of the current

```java
if ((os == OsType.DARWIN) || (os == OsType.UNIX)) {
```

jglick: As a matter of style (consider a subclass in another package):

```suggestion
        InputStream launcherStream = BourneShellScript.class.getResourceAsStream(launcher_binary);
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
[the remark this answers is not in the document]
jglick: You can use [`golang`](https://hub.docker.com/_/golang) for this purpose (a specific tag please)! The problem is that then the `Jenkinsfile` needs to specify

```groovy
platforms: ['docker']
```

which means that we would no longer be running _tests_ on Windows build nodes, because we do not yet have Docker support on them ([INFRA-1400](https://issues.jenkins-ci.org/browse/INFRA-1400)). I guess we could have an exotic `Jenkinsfile` like

```groovy
buildPlugin(platforms: ['docker']) // usual stuff, but only on Linux
node('windows') {
    unarchive '…' // get the *.hpi
    bat 'unzip …' // extract Win32 launcher from plugin build somewhere into target/classes/…
    bat 'mvn test'
}
```

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -286,7 +285,7 @@ public void smokeTest() throws Exception {
     }
 
     private void setGlobalInterpreter(String interpreter) {
-        j.jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).setShell(interpreter);
+        ExtensionList.lookup(Shell.DescriptorImpl.class).get(0).setShell(interpreter);
```
jglick:
```suggestion
        ExtensionList.lookupSingleton(Shell.DescriptorImpl.class).setShell(interpreter);
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,6 +188,68 @@ public String getScript() {
         return c;
     }
 
+    private String binaryLauncherCmd(ShellController c, FilePath ws, OsType os, String interpreter, String scriptPath, String cookieValue, String cookieVariable) throws IOException, InterruptedException {
+        String logFile = c.getLogFile(ws).toURI().getRawPath();
+        String resultFile = c.getResultFile(ws).toURI().getRawPath();
```
jglick: Nope, this will blow up for directories with unusual characters due to URI escaping.

```suggestion
        String resultFile = c.getResultFile(ws).getRemote();
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,6 +188,68 @@ public String getScript() {
...
+        FilePath controlDir = c.controlDir(ws);
+        String outputFile = c.getOutputFile(ws).toURI().getRawPath();
```
jglick:
```suggestion
        String outputFile = c.getOutputFile(ws).getRemote();
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -144,47 +148,24 @@ public String getScript() {
...
-                interpreter,
-                scriptPath,
-                c.getOutputFile(ws),
-                logFile,
-                resultFile, resultFile, resultFile);
+        envVars.put(cookieVariable, "please-do-not-kill-me");
+
+        String launcherCmd = null;
+        if ((os == OsType.DARWIN) || (os == OsType.UNIX)) {
+            String catchSig = "trap ':' INT TERM EXIT; ";
```
jglick: Wait, what? The idea of using a binary launcher is that this command handles everything, including signals. There should be no `sh` process in that case. I see that there _is_ a signal catcher in the binary, so why is this here in the non-legacy branch?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -144,47 +148,17 @@ public String getScript() {
...
         }
 
-        envVars.put(cookieVariable, "please-do-not-kill-me");
         // The temporary variable is to ensure JENKINS_SERVER_COOKIE=durable-… does not appear even in argv[], lest it be confused with the environment.
-        String cmd;
-        FilePath logFile = c.getLogFile(ws);
-        FilePath resultFile = c.getResultFile(ws);
-        FilePath controlDir = c.controlDir(ws);
-        if (capturingOutput) {
-            cmd = String.format("pid=$$; { while [ \\( -d /proc/$pid -o \\! -d /proc/$$ \\) -a -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc %s '%s' > '%s' 2> '%s'; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
```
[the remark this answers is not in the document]
jglick: I think this was since resolved? If so, use the **Resolve conversation** button.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -207,6 +188,68 @@ public String getScript() {
...
+                    logFile,
+                    cookieValue,
+                    cookieVariable,
+                    interpreter,
+                    scriptPath,
+                    c.getOutputFile(ws),
+                    logFile,
+                    resultFile, resultFile, resultFile);
+        } else {
+            cmd = String.format("pid=$$; { while [ \\( -d /proc/$pid -o \\! -d /proc/$$ \\) -a -d '%s' -a \\! -f '%s' ]; do touch '%s'; sleep 3; done } & jsc=%s; %s=$jsc %s '%s' > '%s' 2>&1; echo $? > '%s.tmp'; mv '%s.tmp' '%s'; wait",
```
jglick: As I think I mentioned previously, we expect almost all users to be running the binary launcher so we can strip out some of the hard-to-test exotica in this case. Probably drop `nohup` (already did that?); and the `/proc` heartbeat trick (`/proc` will only work on Linux anyway!).

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,165 @@
...
+}
+
+func launcher(wg *sync.WaitGroup, pidChan chan int,
+	scriptString string, logPath string, resultPath string, outputPath string) {
+
+	defer wg.Done()
+	recordExit := fmt.Sprintf("status=\"$?\"; echo \"$status\" > %v.tmp; mv %v.tmp %v; wait; exit \"$status\"",
+		resultPath, resultPath, resultPath)
+	scriptWithExit := trapSig + "; " + scriptString + "; " + recordExit
+	scriptCmd := exec.Command("/bin/sh", "-c", scriptWithExit)
```
jglick: Huh?? The whole point of using a Golang binary is to reimplement this functionality without shell scripts.

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,165 @@
...
+		defer outputFile.Close()
+		scriptCmd.Stdout = outputFile
+	} else {
+		scriptCmd.Stdout = logFile
+	}
+	if outputPath != "" {
+		// capturing output
+		scriptCmd.Stderr = logFile
+	} else {
+		// Note: pointing to os.Stdout will not capture all err logs and fail unit tests
```
jglick: Do we have some test coverage verifying that something like

```groovy
sh 'echo problems here >&2'
```

indeed prints `problems here` to the build log?

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,165 @@
...
+		fmt.Fprintf(os.Stderr, "%v\n", err.Error())
+		return
+	}
+	_, err = os.Stat(resultPath)
+	if !os.IsNotExist(err) {
+		fmt.Printf("Result file already exists, stopping heartbeat.\n%v\n", resultPath)
+		return
+	}
+	// create the heartbeat script
+	heartbeat := fmt.Sprintf("#! /bin/sh\n%v; pid=\"$$\"; echo \"(heartbeat) pid: \"$pid\"\"; while true ; do kill -0 %v; status=\"$?\"; if [ \"$status\" -ne 0 ]; then break; fi; echo \"\"$pid\"\"; touch %v; sleep 3; done; echo \"(\\\"$pid\\\") exiting\"; exit",
```
jglick: As above, this should be reimplemented as a goroutine or something. Part of the problem with the current implementation is that having the heartbeat done in a separate process is unreliable!

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,165 @@
...
+	err = os.Chmod(heartbeatPath, 0755)
+	checkHeartbeatErr(err)
+	heartbeatScript.WriteString(heartbeat)
+	heartbeatScript.Close()
+
+	heartbeatCmd := exec.Command(heartbeatPath)
+	/************************************
+	// Warning: DO NOT set cmd.Stdout/StdErr is set to os.Stdout/Stderr
+	// If you do, the heartbeat thread will not survive jenkins termination
+	///////////// DO NOT DO THIS ///////////////
```
jglick: Well, better write a test which fails if you do!

jglick:
> Checked this against the [kubernetes-plugin ](https://github.com/jenkinsci/kubernetes-plugin)unit tests locally.

Did you make sure it was actually running (not skipping) the live tests against a K8s cluster? Something along the lines of

```sh
sudo snap install --classic microk8s
microk8s.config > ~/.kube/config
sudo microk8s.enable dns ingress storage
mvn test -Pmicrok8s
```

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
car-roll: So I originally had everything as goroutines, but then discovered that my main golang program was crashing whenever i terminated the Jenkins instance. This then prevented the ability to write out the exit status.  I think ended up converting everything to scripts to survive Jenkins instance terminations.

However, I did discover that a lot of my crashes were caused because I was piping to stdout, so when the Jenkins instance got terminated, it also crashed the binary.  I can most likely revisit this and convert back to all goroutines again.

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
car-roll: I made a previous comment. tldr version: 1) Was always crashing as a goroutine 2)figured a workaround 3) discovered why it was crashing recently 4) didn't think about going back to goroutines 5) will now go back to giving goroutines a try

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
jglick: Yes, the hope was for the agent JVM to launch-and-forget a Golang binary, which in turn would directly launch the user process (typically `sh` but might be something else depending on `#!` and `Shell.DescriptorImpl.shell`) without any other helper processes and using only well-supported Golang system APIs.

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
car-roll: We have at least one in configuredinterpreter where you pass a bad interpreter and let it error out.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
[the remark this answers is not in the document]
car-roll: decided to use maven build profiles to build the binaries only on the docker slave.  seemed simpler (at least in my head)

car-roll:
Looks like the last success actually had an error in the deploy stage
`[2019-06-07T07:21:37.860Z] Invalid archive retrieved from Jenkins, perhaps the plugin is not properly incrementalized?`
I'm guessing that's what is causing the subsequent pipeline failures

pom.xml
```diff
@@ -56,8 +58,109 @@
...
+            <activation>
+                <os>
+                    <family>unix</family>
+                </os>
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>exec-maven-plugin</artifactId>
```
jglick: BTW there is also a Maven plugin to run Docker (or probably a bunch of them), which may or may not be helpful.

pom.xml
```diff
@@ -56,8 +58,109 @@
...
+                                <goals>
+                                    <goal>exec</goal>
+                                </goals>
+                                <configuration>
+                                    <arguments>
+                                        <argument>run</argument>
+                                        <argument>-i</argument>
+                                        <argument>--rm</argument>
+                                        <argument>--mount</argument>
+                                        <argument>type=bind,src=${basedir}/${binary.src},dst=/org/jenkinsci/plugins/durabletask</argument>
```
jglick: Better to take all the `docker` commands and put them in a separate shell script or something, so that the `maven-exec-plugin` configuration is short.

pom.xml
```diff
@@ -56,8 +58,109 @@
...
+                                <goals>
+                                    <goal>copy-resources</goal>
+                                </goals>
+                                <configuration>
+                                    <outputDirectory>src/main/resources/org/jenkinsci/plugins/durabletask</outputDirectory>
+                                    <resources>
+                                        <resource>
+                                            <directory>src/main/go/org/jenkinsci/plugins/durabletask</directory>
+                                            <includes>
+                                                <include>heartbeat-launcher-*</include>
```
jglick: This should not have been built inside the `src/` tree to begin with. Have the external script copy what it needs to `target/` and do all further processing there.

pom.xml
```diff
@@ -56,8 +58,109 @@
...
+            <version>1.17</version>
             <scope>test</scope>
         </dependency>
     </dependencies>
+    <profiles>
+        <profile>
+            <id>docker-generate-binaries</id>
+            <activation>
+                <os>
+                    <family>unix</family>
```
jglick: This seems unwise. It means that any random developer using Windows who runs `mvn package` will get a `durable-task.hpi` which is in fact unusable, with no warning. (Conversely, people with Windows subsystem for Linux will not be able to do builds, perhaps.) That is one reason I suggested the two-stage `Jenkinsfile` approach.

pom.xml
car-roll: yeah,  i was debating whether to use them or not. I think because this is less common case where I'm building an image but not pushing it, I'll put all the commands in a script so someone can better see what's going on.

pom.xml
car-roll: will do

pom.xml
car-roll: will include that in the script i am adding

pom.xml
car-roll: removed. will work on implementing two-stage jenkinsfile approach

car-roll:
added a two-staged approach. The plugin is first built and tested on the docker slave. The windows slave will then take the compiled classes from the archived jar file and expand them in  target/classes.  It  is then a matter of  running mvn resources:testResources, mvn compiler:testCompile, and finally  mvn surefire:test

The jar binary was used to extract the packaged  files as I could not get the unzip step working (I think the utility plugin is not installed). Also, I had trouble calling the infra functions, so I ended up somewhat duplicating that code.

car-roll:
slaves are taking forever to run.  Windows tests initially showed success, but now show failures on another test PR I have running.

car-roll:
Looks like it's just a single error in the `PowershellScriptTest` under `implicitError`.  Says it cannot find the phrase `Bogus error`, but it looks like in the logs it's right there, albeit a bit formatting-mangled.  I tried a regex version to find the phrase, but still no dice.  I don't have a working windows environment to test this on, unfortunately.

car-roll:
@jglick confirmed that tests in kubernetes plugin passed using microk8s (`mvn test -Pmicrok8s`)

car-roll:
And now the tests pass again...It looks like it retries multiple times for timeouts?
https://ci.jenkins.io/job/Plugins/job/durable-task-plugin/view/change-requests/job/PR-92/35/

dwnusbaum:
@car-roll When you have a chance, it would be good to open a PR to [workflow-durable-task-step](https://github.com/jenkinsci/workflow-durable-task-step-plugin/pulls) that just updates its dependency on durable-task to the latest incremental version of this PR (currently [1.30-rc497.f19493358d38](https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/durable-task/1.30-rc497.f19493358d38/)) to show that the tests there still pass with this change.

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,235 @@
...
+		// capturing output
+		scriptCmd.Stderr = scriptLogger.Writer()
+	} else {
+		// Note: pointing to os.Stdout will not capture all err logs
+		scriptCmd.Stderr = scriptCmd.Stdout
+	}
+	// Create new session
+	scriptCmd.SysProcAttr = &unix.SysProcAttr{Setsid: true}
+	for i := 0; i < len(scriptCmd.Args); i++ {
+		launchLogger.Printf("args %v: %v\n", i, scriptCmd.Args[i])
```
dwnusbaum: Do the old scripts do this? If not seems better to avoid printing args in case of secrets, although they may be masked depending on exactly what plugins are being used.

Ah ok, so after reading on, it looks like these loggers just act like `/dev/null` unless a debug flag is set?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -45,13 +48,13 @@
 import java.util.logging.Logger;
 import java.io.File;
 import javax.annotation.Nonnull;
+
```
dwnusbaum: nit: Not a big deal, but try to avoid miscellaneous whitespace-only changes unless that is all your PR is changing.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -147,50 +153,37 @@ public String getScript() {
...
-                interpreter,
-                scriptPath,
-                c.getOutputFile(ws),
-                logFile,
-                resultFile, resultFile, resultFile);
+        envVars.put(cookieVariable, "please-do-not-kill-me");
+
+        String arch = ws.act(new getArchitecture());
+        List<String> launcherCmd = null;
+        String launcher_binary = LAUNCHER_PREFIX + os.toString() + arch;
```
dwnusbaum: Java convention is camelCase
```suggestion
        String launcherBinary = LAUNCHER_PREFIX + os.toString() + arch;
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -311,6 +358,14 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
         private static final long serialVersionUID = 1L;
     }
 
+    private static final class getArchitecture extends MasterToSlaveCallable<String,RuntimeException> {
+        @Override public String call() throws RuntimeException {
+            // Note: This will only determine the architecture of the JVM.
+            return System.getProperty("sun.arch.data.model");
```
dwnusbaum: Is this specific to Oracle/OpenJDK? I would move the lookup here into the same callable as `getOsType`, change it to `getOsAndArch` or something and return both values simultaneously so we don't have to make two separate calls to the remote agent in quick succession.

If you only use the JVM arch, why not use the `os.arch` property instead? Just from the name seems more likely to be supported across JVMs. Did you do any testing to see if 32-bit Go binaries work on 64-bit platforms when a 32-bit JVM is in use?

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -83,25 +92,132 @@
...
+        String script = String.format("echo hello world; sleep %s", sleepSeconds);
+        Controller c = new BourneShellScript(script).launch(new EnvVars(), ws, launcher, listener);
         awaitCompletion(c);
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         c.writeLog(ws,baos);
         assertEquals(0, c.exitStatus(ws, launcher, listener).intValue());
         assertTrue(baos.toString().contains("hello world"));
         c.cleanup(ws);
+        if (notNative) {
+            assertTrue("no zombies running", noZombies());
```
dwnusbaum: So the binaries leave zombie processes behind only in this test? Any idea why? Would be nice to have a comment in the test explaining the reason.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -120,27 +236,37 @@ public void smokeTest() throws Exception {
         assertTrue(log.contains("sleep 999"));
         assertTrue(log.contains("got SIG"));
         c.cleanup(ws);
+        if (simpleOrTini()) {
+            assertTrue("no zombies running", noZombies());
```
dwnusbaum: Would be good to have a comment somewhere to explain what is special about these cases that causes them to leave zombies behind.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -281,85 +415,73 @@ public void smokeTest() throws Exception {
...
-        assertThat("no zombies running", baos.toString(), not(containsString(" Z ")));
-        s.toComputer().disconnect(new OfflineCause.UserCause(null, null));
+        return !baos.toString().contains(" Z ");
+    }
+
+    // to assist with debugging
+    private String psOut() throws InterruptedException {
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        try {
+            launcher.launch().cmds("ps", "-ef").stdout(new TeeOutputStream(baos, System.out)).join();
```
dwnusbaum: How noisy is this? Maybe better to delete all of the calls in the tests but leave the function here in case someone wants to use it while investigating a test failure? Would be good to update the comment above the function to explain what it does in addition to noting that it is useful for debugging.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -281,85 +415,73 @@ public void smokeTest() throws Exception {
...
     }
 
+    private void awaitCompletion(Controller c) throws IOException, InterruptedException {
+        while (c.exitStatus(ws, launcher, listener) == null) {
+            Thread.sleep(100);
+        }
+    }
+
+    private void setGlobalInterpreter(String interpreter) {
+        ExtensionList.lookupSingleton(Shell.DescriptorImpl.class).setShell(interpreter);
```
dwnusbaum: nit: I'd delete this single-line method and inline the body everywhere it was called.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -131,7 +131,7 @@
         ByteArrayOutputStream baos = new ByteArrayOutputStream();
         c.writeLog(ws, baos);
         assertTrue(c.exitStatus(ws, launcher).intValue() != 0);
-        assertThat(baos.toString(), containsString("Bogus error"));
+        assertTrue(baos.toString().contains("Bogus error"));
```
dwnusbaum: Any reason for this change? The previous assertion with `assertThat` has the benefit of showing the expected and actual strings if the assertion failed, whereas the new assertion will just say something like "expected true, got false", which is a lot less useful.

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
car-roll: Yes, that is correct, it will only output (to a file) when the debug flag is given.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
car-roll: No, I was changing things up because that assertion was failing on ci.jenkins.io, but in the end it turned out to be a timeout issue with our build servers.  I'll go ahead and change it back.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Will do.  It turns out it's not very noisy for the docker tests, but for native, it is _quite_ noisy.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Will do, adding this comment as a reminder to add a bit more in this thread as well because I actually am wondering if we still need to keep these specific zombie tests.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: It was mostly just mimicking how the tests were running in the previous iteration, but I wonder if  we need to explicitly do this.  Will expend on this thought.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: So this property is apparently also available on IBM jvm as well, but I did not specifically check.  The thing about using os.arch is that it is meant only to be for human eyes.  There are actually a good amount of values that could be populated in there.
see:
https://stackoverflow.com/questions/2062020/how-can-i-tell-if-im-running-in-64-bit-jvm-or-32-bit-jvm-from-within-a-program#comment25544753_2062045

Jenkinsfile
```diff
@@ -1 +1,71 @@
...
+        stage ('Test (windows-8)') {
+
+            dir('durable-task') {
+                deleteDir()
+                checkout scm
+                bat 'mkdir target\\classes'
+                bat 'mkdir target\\hpi'
+                // Need compiled java jar. Since we don't know the specific name, and because
+                // multiple jars are archived, get the hpi that contains the compiled jar
+                unarchive mapping: ['**/*.hpi': 'hpi']
```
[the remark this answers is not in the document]
car-roll: That's definitely a lot simpler and, like you said, would address @jglick 's concerns about  accidentally building/releasing without the go binaries.  
Another +1 in that it would allow us to also test Linux and WIndows in parallel again.

pom.xml
```diff
@@ -56,8 +59,41 @@
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
             <artifactId>ssh-slaves</artifactId>
-            <version>1.25</version>
+            <version>1.27</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jenkins-ci.plugins</groupId>
+            <artifactId>structs</artifactId>
```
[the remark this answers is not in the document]
car-roll: Will look into it

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -83,25 +92,132 @@
...
+        System.out.println("My platform: " + platform);
+        this.platform = platform;
+        this.listener = StreamTaskListener.fromStdout();
     }
 
-    @Test
-    public void smokeTest() throws Exception {
-        Controller c = new BourneShellScript("echo hello world").launch(new EnvVars(), ws, launcher, listener);
+    @Before public void prepareAgentForPlatform() throws Exception {
+        System.out.println("prepare platform: " + platform);
```
dwnusbaum: Didn't see this earlier, but it should probably be deleted.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -39,39 +40,47 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.Parameterized;
+import org.junit.runners.Parameterized.Parameters;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
 import org.jvnet.hudson.test.LoggerRule;
 import org.jvnet.hudson.test.SimpleCommandLauncher;
 
+enum TestPlatform {
+    NATIVE, ALPINE, CENTOS, UBUNTU, SLIM, SIMPLE, TINI
```
dwnusbaum: It would be good to have a platform like `NATIVE_NON_BINARY` (or `UBUNTU_NON_BINARY` to avoid platform-specific issues) to just have some kind of test for the old script.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
dwnusbaum: My question was more about why this specific test is looking at `nonNative` while the others are all looking at `simpleOrTini()`, since it looks like before this change there was no code to check for zombies at all anywhere, so it's not obvious to me what is special about the native platform for this test vs the other tests.

No strong opinion on whether we want the zombie tests or not. If they give use useful information on the other platforms, then seems fine to leave them in to keep us from accidentally making a change that regresses the behavior. If they are pointless because all of the docker fixtures pass `--init` and so it doesn't matter what our script does, then it seems better to remove them.

car-roll: Yeah, to be honest I'm not sure how my tests evolved that way to not checking for zombies in native form. I may have been emulating how the original tests only did  a zombie check for smoke tests when running on a docker platform. 

My plan is to remove the zombie checks completely, and then see if i can add `UBUNTU_NO_BINARY` platform.

jglick: > the docker agent now runs with `--init`

I changed the documentation to _recommend_ that it does. Except in Kubernetes packagings, we cannot control for sure how users run their agents. There was still a bug in this plugin, which I fixed and wrote a test for, so I recommend the tests continue to verify that no zombies are left behind.

car-roll: @jglick  thanks for the input, I'll go ahead and keep the zombie tests.  However, I think I can remove the `tini` container since we know it is reaping zombies, we don't need to test for that.  I'll also just go ahead and test for zombies only when running the `simple` container

@dwnusbaum Just fyi, I think I do remember now why I removed the zombie test for native. It was because the way that zombies were being checked works for a relatively clean docker container, but in a process-noisy native environment, the test always seems to catch phantom zombies.

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	err := scriptCmd.Start()
+	if checkScriptErr(err) {
+		exitLauncher(-2, resultPath)
+		return
+	}
+	pid := scriptCmd.Process.Pid
+	launchLogger.Printf("launched %v\n", pid)
+	err = scriptCmd.Wait()
```
jglick: It is good that you are calling `Setsid` but I think you also need to fork here somewhere. Otherwise the Golang wrapper as launched by the JVM will not exit until the user process exits, whereas we would like the JVM process handle (reflected in Jenkins APIs as a `Proc`) to exit as soon as the user process has been launched, typically ~100ms after the `sh` step starts. See #98 in which I am enforcing this behavior in tests.

car-roll: It was my understanding that the idea forking, at least in the traditional sense, is not really done in golang. The closest thing you can get to calling a traditional fork in golang would be an `execve`.  Instead, I'm using the `exec` package to spin up a new golang thread.

In this case, I want the golang wrapper to _not_ exit because it is in charge of performing the heartbeat while the  launched script is still running.  I think I might be missing something here?

jglick: > forking, at least in the traditional sense, is not really done in golang

True (they want you to use external tooling to daemonize) but it seems [it is possible anyway](https://github.com/sevlyar/go-daemon).

> I really want the golang wrapper to _not_ exit because it is in charge of performing the heartbeat

_Some_ Go process must remain running in order to update the heartbeat, redirect stdio, and capture the user process’ exit code. Just not the one that the JVM originally launched.

car-roll: I've been trying to wrestle with a solution in the code as well as in my head. There seem to be two goals here at conflict with one another, and I wished I realized this sooner so I could ask this question in PR-98. 

One goal is being able to run with no zombies in a docker container that does not run `tini`, ie, no zombie reaping for processes with `PPID` 1.  Another goal is the requirement of launching the wrapper process in the background (PR-98).  Any `Proc` that then launches a child into the background, and then exits, will cause that child process to have a `PPID` of 1. And if we are running in a docker container with no `init` process, we have now created a zombie.

I can verify with the parameterized tests in this PR that the changes made in PR-98 will create a zombie in the `simple` docker container. Specifically, the `justSlow`, `reboot`, and `watch` unit test. In other words, anytime a shell script runs long enough, a zombie will be created.

If this assumption is correct, it would seem that the problem that PR-98 is solving is more likely to occur than running out of PIDs? 🤞 And that we just recommend users to start their containers with `--tini`?

jglick: Bear with me here, as I am not deeply familiar with this area, nor with all the changes you have been making in this PR.

> I can verify with the parameterized tests in this PR that the changes made in PR-98 will create a zombie in the `simple` docker container.

Do you mean to say that #98 introduced a regression? And if so, why would `BourneShellScriptTest.runWithCommandLauncher` in `master` not have caught it?

> we just recommend users to start their containers with `--tini`?

I do not think we can do that without at least some sort of diagnostic warning. Specifically in Kubernetes [there is no such option](https://stackoverflow.com/a/51172104/12916), though it is unclear to me what the practical impact would be in various implementations. Jenkins `sh` steps are sufficiently weird that standard guidelines do not apply.

By the way, when testing in conjunction with the `kubernetes` plugin, it is critical to test both inside and outside the `container` step, as these do entirely different things: outside, the agent JVM is starting the user script using `java.lang.Process` just like a traditional “static” agent; inside, the agent JVM is only used to tail the log file and the like, and the actual process launch is done via the equivalent of `kubectl exec`.

(Quite by coincidence, I just now stumbled across [this](https://ahmet.im/blog/minimal-init-process-for-containers/).)

car-roll: So I do believe #98 introduced a regression, but the reason that `runWithCommandLauncher` did not catch it was because the script was not running long enough.  If you extend the sleep to `60` seconds, for example, the test should fail.  The only reason I caught it was because in my parameterized tests, I run the zombie check on all the tests, so it caught the zombie on the longer running tests. 
Yeah, as long as we have some sort of way to let developers know.  We need some sort of PPID 1 process reaping if we are going to be launching anything in the background without waiting for it to complete.

car-roll: @jglick would it be reasonable then to remove zombie checks from the simple command launcher (since we know that backgrounding a process and then exiting will leave one). And then just adding the diagnostic warning that you mentioned saying that some `init` process will be required if you are running in containers?

jglick: I guess. I am still trying to figure out what the implications would be, especially for Kubernetes users.

Jenkinsfile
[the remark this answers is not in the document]
car-roll: @dwnusbaum I was going through the [`buildPlugin` code](https://github.com/jenkins-infra/pipeline-library/blob/66a14842e350bd5364a1a70ca4287535cacc0390/vars/buildPlugin.groovy) and I don't think it allows properties to be directly passed. I think it only takes in a defined list of flags.

pom.xml
```diff
@@ -56,8 +59,79 @@
...
+                                    <goal>exec</goal>
+                                </goals>
+                            </execution>
+                        </executions>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+    </profiles>
+<!--    <build>-->
```
car-roll: whoops, will remove this

Jenkinsfile
```diff
@@ -1 +1,71 @@
...
+
+node('windows') {
+    timeout(60) {
+        stage ('Test (windows-8)') {
+
+            dir('durable-task') {
+                deleteDir()
+                checkout scm
+                bat 'mkdir target\\classes'
+                bat 'mkdir target\\hpi'
```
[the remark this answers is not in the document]
car-roll: Removed the bat steps. Using a simplified `Jenkinsfile` now

Jenkinsfile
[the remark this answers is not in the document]
car-roll: was able to use `withEnv` to  pass environment variable to maven.  `Jenkinsfile` can now specifically turn off binary generation for  the `windows` environment.  The `pom` will, by default, always attempt to generate the  binaries.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Tests have been moved to a separate PR.  please see #103

src/main/go/org/jenkinsci/plugins/durabletask/heartbeat-launcher.go
car-roll: Tests have been moved to #103 
Zombie tests are now ignored in the `SIMPLE`/`NO_INIT` platform.
See [JENKINS-58656](https://issues.jenkins-ci.org/browse/JENKINS-58656) for tracking on this issue

car-roll:
Due to the large number of commits and merges, I am closing this PR. The work will be broken into two separate PRs. For the main golang port of the shell wrapper process, see #106. For the unit test refactoring, see #103.

PR #95, opened 2019-06-04
Shell.DescriptorImpl.getShellOrDefault is not reliable in the presence of LauncherDecorator

jglick:
Just because the container running the Jenkins agent has `/bin/sh` does not mean the container running the actual process will. For that matter, the remote callable does not even check—it just blindly returns `/bin/sh` on non-Windows platforms. Useless.

May conflict with #92, though that retains the shell launcher as a fallback.

See https://github.com/jenkinsci/kubernetes-plugin/pull/490 for integration test.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/SlimFixture.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/SlimFixture.java
@@ -29,5 +29,8 @@ import org.jenkinsci.test.acceptance.docker.DockerFixture;
 
 /** Analog of {@link JavaContainer} but using Debian Slim. */
 @DockerFixture(id = "slim", ports = 22)
-public class SlimFixture extends DockerContainer {}
+public class SlimFixture extends DockerContainer {
 
+    public static final String SLIM_JAVA_LOCATION = "/usr/local/openjdk-8/bin/java";
+
+}
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -136,7 +136,11 @@ public final class BourneShellScript extends FileMonitoringTask {
         final Jenkins jenkins = Jenkins.getInstance();
         String interpreter = "";
         if (!script.startsWith("#!")) {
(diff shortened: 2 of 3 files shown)


jglick:
Hmm, some sort of CI problem…

src/test/java/org/jenkinsci/plugins/durabletask/SlimFixture.java
```diff
@@ -29,5 +29,8 @@
 
 /** Analog of {@link JavaContainer} but using Debian Slim. */
 @DockerFixture(id = "slim", ports = 22)
-public class SlimFixture extends DockerContainer {}
+public class SlimFixture extends DockerContainer {
 
+    public static final String SLIM_JAVA_LOCATION = "/usr/local/openjdk-8/bin/java";
```
olamy: for me it's `  /Library/Java/JavaVirtualMachines/jdk1.8.0_201.jdk/Contents/Home/ ` :)
TBH I haven't been trough all the code to be able to really understand this so maybe I'm wrong with my comment but I was a bit `doh`  :)

jglick: This is about [location](https://github.com/jenkinsci/durable-task-plugin/blob/87d6e9fccebbf5308a88582860902e9468fd8541/src/test/resources/org/jenkinsci/plugins/durabletask/SlimFixture/Dockerfile#L1) inside [`openjdk:8-jre-slim`](https://hub.docker.com/_/openjdk/), which changed recently with https://github.com/docker-library/openjdk/pull/322 and was not in any of the [expected locations](https://github.com/jenkinsci/ssh-slaves-plugin/blob/95f528730fc1e01b25983459927b7516ead3ee00/src/main/java/hudson/plugins/sshslaves/DefaultJavaProvider.java#L78-L83) apparently.

(`docker-fixtures` rebuilds images from `Dockerfile` on demand, and our base images are not immutable tags, so we are in the unfortunate position that CI builds sometimes fail for reasons unrelated to the patch at hand, as here.)

jglick:
> rebase off of these changes

`merge` is better IMO.

src/test/java/org/jenkinsci/plugins/durabletask/SlimFixture.java
olamy: @jglick thanks for the explanation

PR #97, opened 2019-06-27
If Proc is AutoCloseable, close it after launch

jglick:
Seems to solve the problem described in https://github.com/jenkinsci/kubernetes-plugin/pull/522, though perhaps not the best way.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -28,7 +28,6 @@ import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.LauncherDecorator;
 import hudson.Platform;
 import hudson.Util;
 import hudson.model.TaskListener;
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -191,23 +184,24 @@ public String getScript() {
...
+            Timer.get().schedule(new Runnable() {
+                @Override public void run() {
+                    try {
+                        LOGGER.log(Level.FINE, "cleaning up {0} in {1}", new Object[] {proc, controlDir});
+                        ((AutoCloseable) proc).close();
+                    } catch (Exception x) {
+                        LOGGER.log(Level.WARNING, "failed to clean up " + proc + " in " + controlDir, x);
+                    }
+                }
+            }, 1, TimeUnit.MINUTES);
```
Vlatombe: I'm not sure I understand how long the process is supposed to live.

jglick: Hours perhaps. `ContainerExecProc.close` merely stops listening to it, because it should have no stdio and an exit status of 0—this is the wrapper process. The stdio and exit status of the user process are captured in files, so this `Proc` handle is worthless (note that we throw it away before returning from the method).

jglick:
Superseded by #98.

PR #98, opened 2019-07-01
[JENKINS-58290] Run wrapper process in the background

jglick:
[JENKINS-58290](https://issues.jenkins-ci.org/browse/JENKINS-58290)

~Alternative to #97: regardless of the `Proc` implementation~ ~Call `Proc.join` after launch~ Make sure we use `&` to launch the wrapper process as a background process. I would hope (but have not yet checked) that #92 also passes the same test (CC @car-roll).

~To my surprise, this does _not_ solve https://github.com/jenkinsci/kubernetes-plugin/pull/522. While `ContainerExecProc.join` calls `close` at the end, `join` never completes, despite the attempt at backgrounding, and so the WebSocket remains open. So there is something fishy going on which I have not identified.~ Seems that https://github.com/moby/moby/issues/33039#issuecomment-353250651 is key.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -28,7 +28,6 @@ import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.LauncherDecorator;
 import hudson.Platform;
 import hudson.Util;
 import hudson.model.TaskListener;
@@ -38,8 +37,6 @@ import java.nio.charset.Charset;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Map;
-import java.util.WeakHashMap;
 import java.util.concurrent.TimeUnit;
 import java.util.logging.Level;
 import java.util.logging.Logger;
@@ -51,8 +48,9 @@ import hudson.remoting.VirtualChannel;
 import org.kohsuke.stapler.DataBoundConstructor;
 import javax.annotation.CheckForNull;
 import jenkins.MasterToSlaveFileCallable;
-import org.apache.commons.io.FileUtils;
 import com.google.common.io.Files;
+import hudson.Proc;
+import jenkins.util.Timer;
 
 
 
@@ -67,10 +65,16 @@ public final class BourneShellScript extends FileMonitoringTask {
 
     private static final String SYSTEM_DEFAULT_CHARSET = "SYSTEM_DEFAULT";
 
-    /** Number of times we will show launch diagnostics in a newly encountered workspace before going mute to save resources. */
+    /**
+     * Whether to stream stdio from the wrapper script, which should normally not print any.
+     * Copying output from the controller process consumes a Java thread, so we want to avoid it generally.
+     * If requested, we can do this to assist in diagnosis.
+     * (For example, if we are unable to write to a workspace due to permissions,
+     * we would want to see that error message.)
+     */
     @SuppressWarnings("FieldMayBeFinal")
     // TODO use SystemProperties if and when unrestricted
-    private static int NOVEL_WORKSPACE_DIAGNOSTICS_COUNT = Integer.getInteger(BourneShellScript.class.getName() + ".NOVEL_WORKSPACE_DIAGNOSTICS_COUNT", 10);
+    private static boolean LAUNCH_DIAGNOSTICS = Boolean.getBoolean(BourneShellScript.class.getName() + ".LAUNCH_DIAGNOSTICS");
 
     /**
      * Seconds between heartbeat checks, where we check to see if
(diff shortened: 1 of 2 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -188,26 +178,33 @@ public String getScript() {
         }
         cmd = cmd.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
 
-        args.addAll(Arrays.asList("sh", "-c", cmd));
+        if (LAUNCH_DIAGNOSTICS) {
+            args.addAll(Arrays.asList("sh", "-c", cmd));
+        } else {
+            // setsid --fork also works but not in, say, busybox; or even Ubuntu prior to Cosmic
+            args.addAll(Arrays.asList("sh", "-c", "(" + cmd + ") &"));
```
jglick: [Discussion](https://superuser.com/a/172476/129406) and [more](https://unix.stackexchange.com/a/352923/26736).

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -67,10 +61,16 @@
 
     private static final String SYSTEM_DEFAULT_CHARSET = "SYSTEM_DEFAULT";
 
-    /** Number of times we will show launch diagnostics in a newly encountered workspace before going mute to save resources. */
+    /**
+     * Whether to stream stdio from the wrapper script, which should normally not print any.
```
dwnusbaum: Typo or maybe truncated?
```suggestion
     * Whether to stream stdio from the wrapper script, which should normally not print anything.
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -67,10 +61,16 @@
...
+     * Copying output from the controller process consumes a Java thread, so we want to avoid it generally.
+     * If requested, we can do this to assist in diagnosis.
+     * (For example, if we are unable to write to a workspace due to permissions,
+     * we would want to see that error message.)
+     */
     @SuppressWarnings("FieldMayBeFinal")
     // TODO use SystemProperties if and when unrestricted
-    private static int NOVEL_WORKSPACE_DIAGNOSTICS_COUNT = Integer.getInteger(BourneShellScript.class.getName() + ".NOVEL_WORKSPACE_DIAGNOSTICS_COUNT", 10);
+    private static boolean LAUNCH_DIAGNOSTICS = Boolean.getBoolean(BourneShellScript.class.getName() + ".LAUNCH_DIAGNOSTICS");
```
dwnusbaum: Any idea of whether having this show up the first 10 times was useful in practice? I can't think of any cases where I saw output from the wrapper in the past, but maybe I just didn't recognize it when it happened.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
jglick: Not a typo.

> …normally not print any [stdio]

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
jglick: There are some cases. For example, if you messed up

```groovy
withEnv("PATH=${tool 'whatever'}/bin") { // meant to use PATH+WHATEVER or explicitly append/prepend
  sh 'oops'
}
```

then the wrapper script (pending #92) would not start, yet you would not get an explanation of why in the build log. Another case is that you messed something up with `withDockerContainer` so that the workspace volume lacked write access by the effective user ID, so the attempted redirect to the output file would fail. But in the normal case that the wrapper runs (even if you have fatal errors in the user process), piping stdio from the wrapper process merely wasted a thread.

The `encounteredPaths` / `novel` trick being deleted here was an attempt to compromise: when running the launcher the first few times on a particular workspace that might be set up poorly, print diagnostics, but subsequently skip it for efficiency. Unfortunately that trick does not work well in a `Cloud` world—you can expect _every_ workspace to be novel.

Anyway now with the backgrounding it is really impossible to track wrapper process output, so I have kept the diagnostics available only as a global option. With the diagnostic flag off by default, this means that a misconfigured path/workspace/etc. will result in the -2 exit code and the message

> process apparently never started in …

but no further details. Ideally there would be an easy way to enable launch diagnostics (and now, disable backgrounding) on the fly, or at least an option to `BourneShellScript` + `ShellStep` to enable it from the `Jenkinsfile`.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Yeah, that doesn't pass a grammaticality judgement for me, but I'll take your word that it does for you. I'd expect `print to stdio` (or just stdout) or `print any output`.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Ok, thanks for elaborating. The current behavior seems fine to me. Maybe we could have the "process apparently never started" message also mention something like "consider temporarily enabling diagnostic info through the script console by setting ...".

Vlatombe:
As this is expected to fix scaling issues with the kubernetes plugin, it'd be great to be able to cut a release of this plugin with this PR content. Thanks!

dwnusbaum:
@Vlatombe Ok, I plan to release this today, didn't have a chance to get to it earlier this week.

dwnusbaum:
Just noting that @car-roll filed [JENKINS-58656](https://issues.jenkins-ci.org/browse/JENKINS-58656) to track the fact that this change causes zombies to be left behind in environments with no init process. I am not sure if that is something to be concerned about in practice or not.

PR #103, opened 2019-07-25
Refactor UNIX unit tests to be parameterized

car-roll:
Due to #92 growing ~a bit~ unwieldy, the refactoring of the unit tests in that PR have been moved into this PR.

Due to the need of durable task to perform stably across multiple environments, the unit tests in `BourneShellScriptTest` will now be performed across multiple docker platforms.

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
@@ -73,7 +73,7 @@ import org.jvnet.hudson.test.LoggerRule;
 import org.jvnet.hudson.test.SimpleCommandLauncher;
 
 enum TestPlatform {
-    SIMPLE, ALPINE, NATIVE, CENTOS, UBUNTU, SLIM
+    NATIVE, SIMPLE, ALPINE, CENTOS, UBUNTU, SLIM
 }
 
 @RunWith(Parameterized.class)
@@ -192,8 +192,6 @@ public class BourneShellScriptTest {
         int sleepSeconds = -1;
         switch (platform) {
             case NATIVE:
-                sleepSeconds = 0;
-                break;
             case CENTOS:
             case UBUNTU:
             case SIMPLE:
--- a/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
@@ -450,6 +450,7 @@ public class BourneShellScriptTest {
             }
             Thread.sleep(100);
         }
+        System.out.println("output: " + baos);
         assertEquals(0, c.exitStatus(ws, launcher, listener).intValue());
         assertTrue(baos.toString().contains("hello world"));
         c.cleanup(ws);
--- a/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
@@ -117,7 +117,6 @@ public class BourneShellScriptTest {
     private FilePath ws;
     private Launcher launcher;
     private static int counter = 0; // used to prevent docker container name-smashing
-    private boolean testExecuted;
 
     public BourneShellScriptTest(TestPlatform platform) throws Exception {
         this.platform = platform;
@@ -125,10 +124,8 @@ public class BourneShellScriptTest {
     }
 
     @Before public void prepareAgentForPlatform() throws Exception {
-        testExecuted = false;
         switch (platform) {
             case NATIVE:
-                testExecuted = true;
                 s = j.createOnlineSlave();
                 break;
             case SLIM:
@@ -137,7 +134,6 @@ public class BourneShellScriptTest {
             case UBUNTU:
             case NO_INIT:
                 assumeDocker();
-                testExecuted = true;
                 s = prepareAgentDocker();
                 j.jenkins.addNode(s);
                 j.waitOnline(s);
@@ -195,7 +191,7 @@ public class BourneShellScriptTest {
     }
 
     @After public void agentCleanup() throws IOException, InterruptedException {
-        if (testExecuted) {
+        if (s != null) {
             s.toComputer().disconnect(new OfflineCause.UserCause(null, null));
             j.jenkins.removeNode(s);
         }
--- a/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
@@ -25,8 +25,10 @@
 package org.jenkinsci.plugins.durabletask;
 
 import hudson.EnvVars;
+import hudson.ExtensionList;
 import hudson.FilePath;
 import hudson.Launcher;
+import hudson.Platform;
 import hudson.Proc;
 import hudson.model.Slave;
 import hudson.plugins.sshslaves.SSHLauncher;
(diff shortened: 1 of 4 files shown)


car-roll:
The only test that appears to be failing is the `NATIVE` version for `backgroundLaunch`. This is just running the test locally on the agent.
https://ci.jenkins.io/job/Plugins/job/durable-task-plugin/job/PR-103/3/testReport/junit/org.jenkinsci.plugins.durabletask/BourneShellScriptTest/linux_8___Archive__linux_8____backgroundLaunch_2__NATIVE_/
I'm not sure if this is because of the agent or not, but since all the other tests pass locally for the agent, I'm inclined to just disable `backgroundLaunch` for the `NATIVE` platform

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -285,53 +394,44 @@ public void smokeTest() throws Exception {
...
+        awaitCompletion(c);
+        assertThat(baos.toString(), isEmptyString());
+        c.cleanup(ws);
     }
 
-    private void awaitCompletion(Controller c) throws IOException, InterruptedException {
-        while (c.exitStatus(ws, launcher, listener) == null) {
-            Thread.sleep(100);
+    @Issue("JENKINS-58290")
+    @Test public void backgroundLaunch() throws IOException, InterruptedException {
```
car-roll: The only test that appears to be failing is the `NATIVE` version for `backgroundLaunch`. This is just running the test locally on the agent.
https://ci.jenkins.io/job/Plugins/job/durable-task-plugin/job/PR-103/4/testReport/junit/org.jenkinsci.plugins.durabletask/BourneShellScriptTest/linux_8___Archive__linux_8____backgroundLaunch_2__NATIVE_/
I'm not sure if this is because of the agent or not, but since all the other tests pass locally for the agent, I'm inclined to just disable `backgroundLaunch` for the `NATIVE` platform

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -94,25 +112,133 @@ static void assumeDocker() throws Exception {
...
+
+    @After public void agentCleanup() throws IOException, InterruptedException {
+        if (testExecuted) {
+            s.toComputer().disconnect(new OfflineCause.UserCause(null, null));
+            j.jenkins.removeNode(s);
+        }
+    }
+
+    @Test public void smokeTest() throws Exception {
+        int sleepSeconds = -1;
```
jglick:
```suggestion
        int sleepSeconds;
```

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -94,25 +112,133 @@ static void assumeDocker() throws Exception {
...
+            case SIMPLE:
+                sleepSeconds = 10;
+                break;
+            case ALPINE:
+            case SLIM:
+                sleepSeconds = 45;
+                break;
+            default:
+                fail("Unknown enum value: " + platform);
+                break;
```
jglick: or just

```java
throw new AssertionError(platform);
```

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -45,40 +47,56 @@
...
 
+enum TestPlatform {
+    NATIVE, SIMPLE, ALPINE, CENTOS, UBUNTU, SLIM
+}
+
+@RunWith(Parameterized.class)
 public class BourneShellScriptTest {
+    @Parameters(name = "{index}: {0}")
+    public static Iterable<? extends Object> data() {
+        return EnumSet.allOf(TestPlatform.class);
```
jglick: or just

```java
public static Object[] data() {
    return TestPlatform.values();
}
```

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -94,25 +112,133 @@ static void assumeDocker() throws Exception {
...
+        // still shutting down but the new test container is being spun up. Seems more ideal than adding a wait.
+        String name = "docker";
+        String agent = "agent-" + counter++;
+        String remoteFs = "/home/jenkins/" + agent;
+        String dockerRunSimple = String.format("docker run -i --rm --name %s jenkinsci/slave:3.7-1 java -jar /usr/share/jenkins/slave.jar", agent);
+        return new DumbSlave(name, remoteFs, new SimpleCommandLauncher(dockerRunSimple));
+    }
+
+    @After public void agentCleanup() throws IOException, InterruptedException {
+        if (testExecuted) {
```
jglick:
```suggestion
        if (s != null) {
```

and then you can delete the `testExecuted` flag.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -94,25 +112,133 @@ static void assumeDocker() throws Exception {
...
-        launcher = j.jenkins.createLauncher(listener);
+    private Slave prepareAgentDocker() throws Exception {
+        switch(platform) {
+            case SLIM:
+            case ALPINE:
+            case CENTOS:
+            case UBUNTU:
+                return prepareDockerPlatforms();
+            case SIMPLE:
+                return prepareAgentCommandLauncher();
```
jglick: Not clear to me what `SIMPLE` means, could use Javadoc. Running `jenkins/agent` without `--init`? What about running it _with_ `--init`?

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -94,25 +112,133 @@ static void assumeDocker() throws Exception {
...
+        }
+        SystemCredentialsProvider.getInstance().setDomainCredentialsMap(Collections.singletonMap(Domain.global(), Collections.<Credentials>singletonList(new UsernamePasswordCredentialsImpl(CredentialsScope.GLOBAL, "test", null, "test", "test"))));
+        SSHLauncher sshLauncher = new SSHLauncher(container.ipBound(22), container.port(22), "test");
+        sshLauncher.setJavaPath(customJavaPath);
+        return new DumbSlave("docker", "/home/test", sshLauncher);
+    }
+
+    private Slave prepareAgentCommandLauncher() throws Exception{
+        // counter used to prevent name smashing when a docker container from the previous test is
+        // still shutting down but the new test container is being spun up. Seems more ideal than adding a wait.
```
jglick: Or just omit `--name` and let Docker generate a name.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -94,25 +112,133 @@ static void assumeDocker() throws Exception {
...
+        String name = "docker";
+        String agent = "agent-" + counter++;
+        String remoteFs = "/home/jenkins/" + agent;
+        String dockerRunSimple = String.format("docker run -i --rm --name %s jenkinsci/slave:3.7-1 java -jar /usr/share/jenkins/slave.jar", agent);
```
jglick: check https://hub.docker.com/r/jenkins/slave/tags for updates

```suggestion
        String dockerRunSimple = String.format("docker run -i --rm --name %s jenkins/slave:3.29-2 java -jar /usr/share/jenkins/slave.jar", agent);
```

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -295,57 +416,42 @@ public void smokeTest() throws Exception {
         c.cleanup(ws);
     }
 
-    private void setGlobalInterpreter(String interpreter) {
-        j.jenkins.getInjector().getInstance(Shell.DescriptorImpl.class).setShell(interpreter);
+    /**
+     * Checks if the golang binary outputs to stdout under normal shell execution.
```
jglick: Comment copied from #92. Does this test make sense outside of that?

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Yes, I originally  had two platforms, `SIMPLE` (without --init) and `TINI` (with  `--init`'). I removed `TINI` because the other docker containers have `init` processes, so I felt the coverage was already there.  I'll rename `SIMPLE` to something like`NO_INIT` at the very least.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Whoops, you are correct. This test has no reason being here. Well, I take that back. One reasoning could be to make sure that the shell wrapper process is not outputting to stdout so that the docker does not wait for it?  Though we know that it is done for certain since ` >&- 2>&-` is applied. Thoughts?

jglick: > make sure that the shell wrapper process is not outputting to stdout so that the docker does not wait for it

Not sure I understand the concern, but if you think it is important, keep it.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
jglick: Ah, OK, that would be clearer.

PR #106, opened 2019-08-02
[JENKINS-25503] Refactor shell wrapper process to use a binary wrapper

car-roll:
Due to the number growing number of commits and merges that were being added to #92, some cleanup was needed.  The unit test refactoring was removed and placed in #103, while the golang port of the shell wrapper process has been moved to this commit.

Summary from #92:
This is an attempt to harden durable-task. The changes are only for unix-based architectures. Other architectures will be business as usual.
The most notable difference is that the shell script will is now launched via a golang binary. The binary allows the script to have its own session id. This plus some additional signal catching should increase the survivability of long-running scripts when the jenkins terminates unexpectedly. The use of golang also removes the need to poll proc for the executing shell (for systems where proc is available). Building this plugin now requires the golang docker image.

https://issues.jenkins-ci.org/browse/JENKINS-25503
https://issues.jenkins-ci.org/browse/JENKINS-55308

In this version, durable-task will send a copy of a statically compiled Golang binary (approx 2.5mb) from its jar to the slave agent. The binary consists of 3 goroutines:

1. launcher: The first goroutine is responsible for launching the script as a new process with a new session id. It is also responsible for writing out the exit code to the result file and signaling the heartbeat goroutine that the script has finished.
2. heartbeat: The second goroutine is responsible for updating the modified timestamp of the log file periodically. It will only exit when it has received notification that the script has completed from the launcher goroutine.
3. sigcatcher: The third goroutine is a signal catcher, preventing the binary from being terminated by jenkins until the launched script itself is terminated.

Changelog:
- The script will now be launched under a new session, making it more likely to survive things like unexpected agent terminations.
- This does not change the survivability of the script when the underlying environment decides to terminate processes due to things like low memory
- Just like how the shell wrapper executes in the background (since 1.30), the script launcher is a daemonized process. The means that there is an expectation of orphaned-child cleanup (i.e. zombie-reaping) within the underlying environment
- For *NIX systems only, the shell wrapper has been replaced with a binary script launcher. The binary itself is ~2.5MB while its memory footprint is ~800KB heavier than the shell wrapper. Here is a general breakdown:
  - Original version:
    - two wrapper shell  processes that are spawned, the first  process each process between 610-640KB
    - a single 548KB sleep process used to poll the log output file
  - Go binary  version:
    - a single golang binary ~2560KB


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -137,74 +146,120 @@ public String getScript() {
         }
 
         String scriptPath = shf.getRemote();
-        List<String> args = new ArrayList<>();
-
-        if (os != OsType.DARWIN) { // JENKINS-25848
-            args.add("nohup");
```
dwnusbaum: Do we need to preserve this in some form for the fallback method? It doesn't look like it was moved somewhere else.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -137,74 +146,120 @@ public String getScript() {
...
+                                            scriptPath,
+                                            cookieValue,
+                                            cookieVariable);
         } else {
-            // JENKINS-58290: launch in the background. Also close stdout/err so docker-exec and the like do not wait.
-            args.addAll(Arrays.asList("sh", "-c", "(" + cmd + ") >&- 2>&- &"));
+            String scriptString = scriptLauncherCmd(c, ws, interpreter, scriptPath, cookieValue, cookieVariable);
+            scriptString = scriptString.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
+            launcherCmd = new ArrayList<>();
+            if (LAUNCH_DIAGNOSTICS) {
```
dwnusbaum: nit: I would put `scriptString = scriptString.replace("$", "$$")` and this if/else into `scriptLauncherCmd` since `binaryLauncherCmd` contains the equivalent.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -137,74 +146,120 @@ public String getScript() {
...
-        if (LAUNCH_DIAGNOSTICS) {
-            args.addAll(Arrays.asList("sh", "-c", cmd));
+        String arch = ws.act(new getArchitecture());
+        List<String> launcherCmd = null;
+        String launcherBinary = LAUNCHER_PREFIX + os.toString() + arch;
+        InputStream launcherStream = DurableTask.class.getResourceAsStream(launcherBinary);
+        if ((launcherStream != null) && !FORCE_SHELL_WRAPPER) {
+            FilePath controlDir = c.controlDir(ws);
+            FilePath launcherAgent = controlDir.child(launcherBinary);
+            launcherAgent.copyFrom(launcherStream);
```
dwnusbaum: Any thoughts on whether caching the binary on agents would be worth it?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -137,74 +146,120 @@ public String getScript() {
...
-        LOGGER.log(Level.FINE, "launching {0}", args);
-        Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(true);
+
+        LOGGER.log(Level.FINE, "launching {0}", launcherCmd);
+        Launcher.ProcStarter ps = launcher.launch().cmds(launcherCmd).envs(escape(envVars)).pwd(ws).quiet(true);
         if (LAUNCH_DIAGNOSTICS) {
             ps.stdout(listener);
         } else {
             ps.readStdout().readStderr(); // TODO RemoteLauncher.launch fails to check ps.stdout == NULL_OUTPUT_STREAM, so it creates a useless thread even if you never called stdout(…)
+            LOGGER.log(Level.WARNING,"Durable Task launching wrapper process in the background. An init process must be used to reap the orphaned wrapper process.");
```
dwnusbaum: IIUC, this will log a warning to the Jenkins system log for every shell step in every Pipeline, even if there is no problem. Since that will blow up the logs and there's nothing the admin can do to "fix" the warnings, I think it would make more sense to document this in the `sh` step over in workflow-durable-task and remove this warning entirely.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -188,6 +190,9 @@ private Slave prepareDockerPlatforms() throws Exception {
         if (s != null) {
             j.jenkins.removeNode(s);
         }
+        if (BourneShellScript.FORCE_SHELL_WRAPPER) {
+            BourneShellScript.FORCE_SHELL_WRAPPER = false;
```
dwnusbaum: nit: could remove the `if` and just always set `BourneShellScript.FORCE_SHELL_WRAPPER = false;`

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -395,13 +401,30 @@ private Slave prepareDockerPlatforms() throws Exception {
         c.cleanup(ws);
     }
 
+    /**
+     * Checks if the golang binary outputs to stdout under normal shell execution.
+     * The binary must NOT output to stdout or else it will crash when Jenkins is terminated
```
dwnusbaum: Is this still relevant now that the binary runs in the background?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -117,16 +125,17 @@ public String getScript() {
...
-        
+
         ShellController c = new ShellController(ws,(os == OsType.ZOS));
         FilePath shf = c.getScriptFile(ws);
 
         shf.write(script, scriptEncodingCharset);
 
         final Jenkins jenkins = Jenkins.getInstance();
+        String shell = null;
         String interpreter = "";
```
dwnusbaum: nit: Maybe a bit clearer to remove the `interpreter` variable here and instead pass `shell` to `scriptLauncherCmd` and compute `interpreter` inside of that method, since that is what `binaryLauncherCmd` does. (Just thinking of how to make the code path as similar as possible for both cases.)

Jenkinsfile
```diff
@@ -1 +1,8 @@
-buildPlugin()
+def builds = [:]
+builds['docker'] = { buildPlugin(platforms: ['docker']) }
+builds['windows'] = {
+    withEnv(['SKIP_BINARY_GENERATION=true']) {
+        buildPlugin(platforms: ['windows'])
+    }
+}
+parallel builds
```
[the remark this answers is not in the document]
car-roll: Yeah, it does look fine when you start it up, you only see two branches. However, when it finishes, you always see 4 branches.

The only reason I had it this way was because I thought I would have to also check for the OS type of the platform as well, since both platforms would get the environment variable.  I was not sure if there was a clean way to do that in the pom

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: good catch :)

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: That would be useful since the shell step is called numerous times. I initially left it out because I was worried about stale versions of the binary. I would probably need to add some versioning to the binary name or something I assume.  Not sure if there are better ideas than that.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: You are probably right, however if you choose not to run it in `daemon` mode (ie. `LAUNCH_DIAGNOSTIC = true`, then I guess it would still be a problem? But I'm not certain what the normal use cases of `LAUNCH_DIAGNOSTIC` is.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: Will look into this for a later commit

car-roll:
TODO: removed the logger warning here. Must move to `workflow-durable-task`

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -77,7 +78,7 @@
 import org.jvnet.hudson.test.SimpleCommandLauncher;
 
 enum TestPlatform {
-    NATIVE, ALPINE, CENTOS, UBUNTU, SLIM, NO_INIT
+    NATIVE, ALPINE, UBUNTU_NO_BINARY, CENTOS, UBUNTU, SLIM, NO_INIT//, UBUNTU_NO_BINARY
```
car-roll: Will revert this debug line

src/main/go/org/jenkinsci/plugins/durabletask/durable_task_monitor.go
```diff
@@ -0,0 +1,254 @@
...
+			// heartbeat
+			hbLogger.Println("touch log")
+			err = os.Chtimes(logPath, time.Now(), time.Now())
+			loggerIfErr(hbLogger, err)
+			time.Sleep(time.Second * 3)
+		}
+	}
+}
+
+func main() {
```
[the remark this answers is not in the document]
dwnusbaum: Added in ef70f1891bb1ee6388.

src/main/go/org/jenkinsci/plugins/durabletask/rebuild.sh
```diff
@@ -0,0 +1,11 @@
+#! /bin/sh
+# Convenience script to aid in quick rebuilding of golang binaries
+set -x
+NAME="durable_task_monitor"
+RESFOLDER="${HOME}/code/github/durable-task-plugin/src/main/resources/org/jenkinsci/plugins/durabletask"
```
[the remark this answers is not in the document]
dwnusbaum: Done in ef70f1891.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -55,6 +58,9 @@
  */
 public final class BourneShellScript extends FileMonitoringTask {
 
+    @Restricted(NoExternalUse.class)
+    public static boolean FORCE_SHELL_WRAPPER = false;
```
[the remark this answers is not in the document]
dwnusbaum: Done in ef70f1891.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in ef70f1891.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in ef70f1891.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
dwnusbaum: Done in ef70f18.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in ef70f18.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -117,94 +125,148 @@ public String getScript() {
...
+    }
+
+    private List<String> scriptLauncherCmd(ShellController c, FilePath ws, String shell, OsType os, String scriptPath, String cookieValue, String cookieVariable) throws IOException, InterruptedException {
+        String cmdString;
         FilePath logFile = c.getLogFile(ws);
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
+        String interpreter = "";
+
+        if (!script.startsWith("#!")) {
```
dwnusbaum: It's not totally obvious that `shell` is guaranteed non-null in this case. Might be good to add nullability annotations to all of the parameters of this method and `binaryLauncherCmd` so Spotbugs can help make sure we don't introduce any NPEs for untested combinations for untested combinations of values later on.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in 4118e1a.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -137,74 +146,120 @@ public String getScript() {
...
-        }
-        cmd = cmd.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
+        envVars.put(cookieVariable, "please-do-not-kill-me");
 
-        if (LAUNCH_DIAGNOSTICS) {
-            args.addAll(Arrays.asList("sh", "-c", cmd));
+        String arch = ws.act(new getArchitecture());
+        List<String> launcherCmd = null;
+        String launcherBinary = LAUNCHER_PREFIX + os.toString() + arch;
+        InputStream launcherStream = DurableTask.class.getResourceAsStream(launcherBinary);
```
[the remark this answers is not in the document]
dwnusbaum: Done in 4118e1a.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -117,94 +125,148 @@ public String getScript() {
...
+
+    private List<String> scriptLauncherCmd(ShellController c, FilePath ws, String shell, OsType os, String scriptPath, String cookieValue, String cookieVariable) throws IOException, InterruptedException {
+        String cmdString;
         FilePath logFile = c.getLogFile(ws);
         FilePath resultFile = c.getResultFile(ws);
         FilePath controlDir = c.controlDir(ws);
+        String interpreter = "";
+
+        if (!script.startsWith("#!")) {
+            interpreter = "'" + shell + "' -xe ";
```
dwnusbaum: It's not totally obvious that `shell` is guaranteed non-null in this case. Might be good to add nullability annotations to all of the parameters of this method and `binaryLauncherCmd` so Spotbugs can help make sure we don't introduce any NPEs for untested combinations for untested combinations of values later on.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -117,94 +125,148 @@ public String getScript() {
...
 
         String scriptPath = shf.getRemote();
-        List<String> args = new ArrayList<>();
 
-        if (os != OsType.DARWIN) { // JENKINS-25848
-            args.add("nohup");
+        // The temporary variable is to ensure JENKINS_SERVER_COOKIE=durable-… does not appear even in argv[], lest it be confused with the environment.
+        envVars.put(cookieVariable, "please-do-not-kill-me");
+
+        String arch = ws.act(new getArchitecture());
```
dwnusbaum: Would be good to have a single `OsTypeAndArch getOsAndArchitecture` method so that we only have to make one remote call using `ws.act`, not two.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -117,94 +125,148 @@ public String getScript() {
...
+        List<String> launcherCmd;
+        String launcherBinary = LAUNCHER_PREFIX + os.toString().toLowerCase() + arch;
```
dwnusbaum: Turkish users might run into problems looking for `unıx` rather than `unix`, you should pass a `Locale` to avoid issues:

```suggestion
        String launcherBinary = LAUNCHER_PREFIX + os.toString().toLowerCase(Locale.ENGLISH) + arch;
```

I would also add a comment above the `OsType` enum with something to the effect of `changes to the names of these constants must also be mirrored in compile-binaries.sh, ...`. Might be better to instead add a `getNameForBinary` method to the enum that returns a fixed string for each value (and has the desired case from the start) that would never change and call that here instead of `toString().toLowerCase(...)`.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -286,18 +348,26 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
...
-              return OsType.UNIX; // Default Value
+                return OsType.UNIX; // Default Value
             }
         }
         private static final long serialVersionUID = 1L;
     }
 
+    private static final class getArchitecture extends MasterToSlaveCallable<String,RuntimeException> {
+        @Override public String call() throws RuntimeException {
+            // Note: This will only determine the architecture of the JVM.
```
dwnusbaum: nit: I would update the comment to mention that the value will be either `32` or `64`.
```suggestion
            // Note: This will only determine the architecture of the JVM. The result will be "32" or "64".
```
Maybe worth parsing the property into an enum.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
dwnusbaum: That's fair. I think `LAUNCH_DIAGNOSTICS` is essentially unused outside of troubleshooting, because it has to be set manually by an admin.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -132,7 +140,21 @@ public String getScript() {
...
+
+        final Jenkins jenkins = Jenkins.get();
+
+        Manifest manifest;
+        try (InputStream manifestStream = DurableTask.class.getResourceAsStream("/META-INF/MANIFEST.MF")) {
+            manifest = new Manifest(manifestStream);
+        }
+
+        Node wsNode = jenkins.getNode(FilePathUtils.getNodeName(ws));
+        FilePath nodeRoot = wsNode.getRootPath();
```
car-roll: So I noticed that workspace is always two levels below root of the node, is this always the case?
i.e., do I even need to bother getting the node to get the root? Instead, I can  just  call  `getParent` on the workspace twice. I do that  already in the  unit  tests

car-roll:
`Error response from daemon: mkdir /var/lib/docker/tmp/docker-builder566756771: no space left on device` 😡  will try again tomorrow

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: I'm not sure. I would add a new agent to Jenkins, open all of the advanced config and read the help text, but what you have seems safer. I also don't know if there is a preferred place to cache things on agents. There is a passing mention of agent caches [here](https://github.com/jenkinsci/jenkins/blob/a5a92af4bebc73298d2061e8e95227d4e416fc74/core/src/main/resources/hudson/model/Slave/help-remoteFS.html), but that's all I found at a glance.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -283,21 +388,105 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
...
+            // Note: This will only determine the architecture of the JVM. The result will be "32" or "64".
+            ArchType arch;
+            String bits = System.getProperty("sun.arch.data.model");
+            if (bits.equals("64")) {
+                arch = ArchType._64;
+            } else {
+                arch = ArchType._32; // Default Value
+            }
+
+            String launcherBinary = LAUNCHER_PREFIX + launcherVersion + "_" + os.getNameForBinary() + arch;
```
dwnusbaum: nit: I would put the binaries in `$agent_root/cache/durable-task/` or something like that for clarity.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -283,21 +388,105 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
...
+            FilePath launcherPath = nodeRoot.child(launcherBinary);
+            AgentInfo agentInfo = new AgentInfo(os, arch, launcherPath);
+            try {
+                agentInfo.setLauncherAvailability(launcherPath.exists());
+            } catch (IOException | InterruptedException e) {
+                e.printStackTrace();
```
dwnusbaum: I think you should throw the exception so that the master sees the failure. If you do that, it probably makes sense to change the signature of the class to `MasterToSlaveCallable<AgentInfo,Exception>` so that you don't have to wrap the exception.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -283,21 +388,105 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
...
+            }
+            return agentInfo;
+        }
+        private static final long serialVersionUID = 1L;
+    }
     private static final class getOsType extends MasterToSlaveCallable<OsType,RuntimeException> {
```
dwnusbaum: Should this be deleted now?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -283,21 +388,105 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
...
+                return OsType.ZOS;
             } else {
-              return OsType.UNIX; // Default Value
+                return OsType.UNIX; // Default Value
             }
         }
         private static final long serialVersionUID = 1L;
     }
 
+    private static final class getArchitecture extends MasterToSlaveCallable<String,RuntimeException> {
```
dwnusbaum: ditto

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -107,7 +140,22 @@ public String getScript() {
...
+            manifest = new Manifest(manifestStream);
+        }
+
+        Node wsNode = jenkins.getNode(FilePathUtils.getNodeName(ws));
+        FilePath nodeRoot = wsNode.getRootPath();
+
+        if (!Main.isUnitTest) {
+            PLUGIN_VERSION = manifest.getMainAttributes().getValue("Plugin-Version");
+        }
+        AgentInfo agentInfo = nodeRoot.act(new GetAgentInfo(nodeRoot, PLUGIN_VERSION));
```
dwnusbaum: I don't think it makes sense to pass a `FilePath` to a `MasterToSlaveCallable` as an argument. You are already using `FilePath.act`, so if you change `GetAgentInfo` from a `MasterToSlaveCallable<AgentInfo,RuntimeException>` to a `FileCallable<AgentInfo>`, then instead of overriding `call()` you would override `invoke(File nodeRoot, VirtualChannel unused)`, and `invoke` allows you to throw `IOException` and `InterruptedException` so you don't have to handle that directly anymore.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: You added `@Nonnull` to shell, but IIUC it _is_ null if `!script.startsWith("#!")`, since it gets initialized to `null` in `launchWithCookie` and is only set if `script.startsWith("#!")`. Is that not the case?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -107,7 +140,22 @@ public String getScript() {
         if (script.isEmpty()) {
             listener.getLogger().println("Warning: was asked to run an empty script");
         }
-        OsType os = ws.act(new getOsType());
+
+        final Jenkins jenkins = Jenkins.get();
+
+        Manifest manifest;
+        try (InputStream manifestStream = DurableTask.class.getResourceAsStream("/META-INF/MANIFEST.MF")) {
```
dwnusbaum: Rather than working with `Manifest` directly I would maybe do something like `Jenkins.get().getPluginManger().getPlugin("durable-task").getVersion()`.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -432,50 +462,100 @@ private Slave prepareDockerPlatforms() throws Exception {
...
             case SLIM:
-            // (See JENKINS-58656) Running in a container with no init process is guaranteed to leave a zombie. Just let this test pass.
+                // Debian slim does not have ps
             case NO_INIT:
-            // Run only on docker platforms as NATIVE `ps` output cannot guarantee it will detect processes correctly. Just let this test pass.
-            case NATIVE:
-                return "";
+                // (See JENKINS-58656) Running in a container with no init process is guaranteed to leave a zombie.
+                // let test pass, otherwise it will mark entire test as skipped
+                assumeTrue(true);
```
dwnusbaum: Looks like a debug line or something?
```suggestion
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Also, what is `$agent_root` if you are running on `master`? Is it just `$JENKINS_HOME`?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: ugh, you're totally right. I got too paste happy. I even remembering telling myself not to do that :(

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: yes, need to delete

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: ahh, i see. Yeah,  that part bothered me using the filepath inside it. I just told myself that if i use filepath.act and it lives in the node, it would kind of be okay.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: had no idea about that! yes, that helps immensely!

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: but even if we were running on master, there is some file io penalty because the binary would still need to be copied from the jar into the workspace, right? Or do you think it's too small to even bother with?

dwnusbaum: Hmm, yeah that's fair. As long as things are stored in `$JENKINS_HOME/caches/durable_task` on master (since some things tool installations are cached in `$JENKINS_HOME/caches/tool_name`) and not somewhere else then I think what you have is fine.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: actually, that's not a debug line.  What I wanted to have the tests skipped for `SLIM` and `NO_INIT`, but the problem is that `assertZombies` is used in other tests, like `smoke`. If i just skip this test with something like `assumeTrue(!platform.equals(NO_INIT))` then the entire `smoke` test will show as skipped.

dwnusbaum: What I mean though is that `assumeTrue(true)` does nothing, 100% of the time, so `assumeTrue(true); return;` is completely equivalent to `return;`, right?

car-roll: Yeah, you're completely correct. I don't know what my brain was thinking  :p

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: Added a new agent and  dug around, but only found the same thing  you found.

car-roll:
Forever infra issues...

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
dwnusbaum: Done in bea4658.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in 5f7d04e.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in 5f7d04e.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in 5f7d04e.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in 5f7d04e.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Done in f2ac93b.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: nit: Kinda done in 5f7d04ed5, but I think it would be better to use `@Nullable` or `@CheckForNull` rather than just not annotating the things that aren't `@Nonnull`. You'd have to change conditions like the `if (!script.startsWith("#!")) { ... some code that assumes shell is nonnull ... }` to `if (!script.startsWith("#!") && shell != null) { ... }` to make SpotBugs happy, but that's probably clearer anyway.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -432,50 +458,95 @@ private Slave prepareDockerPlatforms() throws Exception {
...
+        assertEquals(0, c.exitStatus(ws, launcher, listener).intValue());
+        Long timeCheck2 = binaryPath.lastModified();
+        assertEquals(timeCheck1, timeCheck2);
+
+        binaryPath.delete();
+        binaryPath.touch(Instant.now().toEpochMilli());
+        try {
+            c = script.launch(envVars, ws, launcher, listener);
+        } catch (Exception e) {
+            assertThat(e, instanceOf(IOException.class));
```
dwnusbaum: To make sure that the try always throws an exception and hits this case, you should call `Assert.fail` at the end of the try. Right now, this test will pass if `script.launch` succeeds (it shouldn't since you corrupted the cache) _or_ if it throws the specific kind of exception you are looking for.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: Hmm, I just realized that `FilePathUtils` comes from workflow-api, so you had to add a new dependency to get this to work. I don't think we want to add a dependency on a Pipeline-specific plugin here. Can you use `ws.toComputer().getNode()` instead? I think you'd need some null checks that fall back to not using caching given the limitations of `FilePath.toComputer`, mentioned in the Javadoc for `FilePathUtils.getNodeNameOrNull`.

Maybe time to go ahead and submit a PR to core to move `FilePathUtils` there for future use.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Ah you're right. If I had done caching incorrectly and the binary copied over, it would've passed.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: No problem, will do

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: that makes way more sense

car-roll: oddly enough, SpotBugs didn't complain when I did not add the `shell != null` for `@CheckForNull`. I still kept it in there though. For `binaryLauncherCmd` i just had it as `@Nullable`

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
dwnusbaum: `FilePath.toComputer` is `@CheckForNull`. Why isn't SpotBugs giving you an error on `FilePath.toComputer().getNode()`? Maybe try running `mvn spotbugs:check` locally to see if that gives you any warnings?

dwnusbaum: Ah, the problem is that the POM makes FindBugs warnings non-fatal: https://github.com/jenkinsci/durable-task-plugin/blob/93b86c92fc0545cb4df79abe0ef2825a2c149827/pom.xml#L27

Locally, I would check for SpotBugs issues and fix all the warnings in code you modified. If that covers everything, I would delete that line from the configuration. If not, I guess I would leave it there for now, unless there are just one or two easy-to-fix warnings.

dwnusbaum: Ok, and like I mentioned yesterday, I was concerned that throwing an `Exception` if any part of `FilePath.toComputer().getNode()` was null was too strict, but as far as I understand in this case they are either non-null or the script will fail to start later on when we use the `Launcher` anyway, so it shouldn't matter.

car-roll:
Kicking off CI again

pyssling:
@car-roll This commit supposedly broke a lot of jenkins deployments: https://issues.jenkins-ci.org/browse/JENKINS-59903 . I guess it should be rolled back.

pyssling:
It's maybe also worth mentioning that jenkins is run on more architectures than x86. I previously ran slaves on ppc without problems. I think the approach taken by this commit isn't the way to go for this work, you may be better off with coding this in java, or possibly python if you feel this functionality is needed.

pyssling:
Had a look at the code for the execution wrapper, I think you could accomplish all this using standard a bash script. This would be better for portability. If you need to communicate between the shell script and java in a smooth way you can use a named pipe, i.e. a fifo or alternatively netcat and a unix socket.

car-roll:
@pyssling thanks for bringing  this to my attention.  will  address  comments  in  the jira  ticket

whittlec:
We're using this on a wide variety of platforms:

- Windows (x64)
- Linux (CentOS 7, AMD64)
- Solaris (SPARC64)
- AIX (PPC64)
- IBM i (PPC64)

IBM i behaves as Unix, so would be caught up in the change, but is much trickier to produce suitable binaries for.

If this feature could be coded up in Java rather than relying on native binaries that would maintain the platform support currently in place.

PR #107, opened 2019-09-25
Drop AtomicReference on FileMonitoringTask.FileMonitoringController#writeLogCs

slonopotamus:
Because it doesn't add anything.


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -174,7 +173,7 @@ String getCharset() {
          * {@link #transcodingCharset} on the remote side when using {@link #writeLog}.
          * May be a wrapper for null; initialized on demand.
```
dwnusbaum:
```suggestion
          * May be null; initialized on demand.
```

slonopotamus: Applied.

PR #114, opened 2019-10-28
[JENKINS-59903]  [JENKINS-59907]  caching with docker and uncommon architectures issues

car-roll:
This PR is attempting to resolve the issues outlined in JENKINS-59903 and JENKINS-59907.

In 59903, the issue occurs when a docker-based agent does not have access to the jenkins caching directory. This prevents the  binary from  ever being copied over to the agent. To fix this, whenever there is no access to the  cache directory, caching will be disabled for the agent and the binary will be copied over to the agent's control directory for every shell execution.

In 59907, the binary is attempting to launch on non-x86 architectures. This  behavior  is currently blocked as the binary has  only been tested on  x86 machines. For  non-x86, non-*NIX architectures, the shell wrapper will be used.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -84,6 +84,7 @@ public final class BourneShellScript extends FileMonitoringTask {
         DARWIN("darwin"),
         UNIX("unix"),
         WINDOWS("windows"),
+        FREEBSD("freebsd"),
         ZOS("zos");
 
         private final String binaryName;
@@ -195,7 +196,7 @@ public final class BourneShellScript extends FileMonitoringTask {
         envVars.put(cookieVariable, "please-do-not-kill-me");
 
         List<String> launcherCmd = null;
-        if (agentInfo.isX86() && (os == OsType.UNIX || os == OsType.DARWIN)) {
+        if (agentInfo.isBinaryCompatible()) {
             FilePath controlDir = c.controlDir(ws);
             FilePath binary;
             if (agentInfo.isCachingAvailable()) {
@@ -204,7 +205,7 @@ public final class BourneShellScript extends FileMonitoringTask {
                 binary = controlDir.child(agentInfo.getBinaryPath());
             }
             try (InputStream binaryStream = DurableTask.class.getResourceAsStream(binary.getName())) {
-                if (FORCE_BINARY_WRAPPER && (agentInfo.isX86()) && (binaryStream != null)) {
+                if (FORCE_BINARY_WRAPPER && (binaryStream != null)) {
                     if (!agentInfo.isCachingAvailable() || !agentInfo.isBinaryCached()) {
                         binary.copyFrom(binaryStream);
                         binary.chmod(0755);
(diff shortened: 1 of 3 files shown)


arharvey918:
This PR still seems like it will continue to try and use the UNIX binary on FreeBSD 11.1 (64-bit), which doesn't currently execute:
```
$ ./durable_task_monitor_1.31_unix_64
ELF binary type "0" not known. 
bash: ./durable_task_monitor_1.31_unix_64: cannot execute binary file: Exec format error
```

Could you add another OS type to the OSType enum for FreeBSD, similar to how you detect Z/OS?
````
println(Platform.current()); // returns UNIX
println(System.getProperty("os.name")); // returns "FreeBSD"
```

car-roll:
@arharvey918 you're right that will still be an issue. I will address that in another PR. In the meantime I want to address some current issues right now  with this PR

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -200,12 +196,17 @@ public String getScript() {
...
+        FilePath binary;
+        FilePath controlDir = c.controlDir(ws);
+        if (agentInfo.isCachingAvailable()) {
+            binary = nodeRoot.child(agentInfo.getBinaryPath());
+        } else {
+            binary = controlDir.child(agentInfo.getBinaryPath());
+        }
         List<String> launcherCmd;
-        FilePath binary = nodeRoot.child(agentInfo.getBinaryPath());
         try (InputStream binaryStream = DurableTask.class.getResourceAsStream(binary.getName())) {
```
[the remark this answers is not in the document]
car-roll: yeah, ideally i didn't want to do that,  but I didn't like doing all the nested conditionals so I kept it this way. I can try to rethink it again.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -454,25 +468,48 @@ public AgentInfo invoke(File nodeRoot, VirtualChannel virtualChannel) throws IOE
                 os = OsType.WINDOWS;
             } else if(Platform.current() == Platform.UNIX && System.getProperty("os.name").equals("z/OS")) {
                 os = OsType.ZOS;
+            } else if(Platform.current() == Platform.UNIX && System.getProperty("os.name").equals("FreeBSD")) {
+                os = OsType.FREEBSD;
```
dwnusbaum: Not that we need to fix it here, but do you know what's wrong with the Go binary on FreeBSD? Is it a Go problem, or do we need to cross-compile for it separately or something?

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -198,25 +195,33 @@ public String getScript() {
...
-                    binary.chmod(0755);
-                }
-                launcherCmd = binaryLauncherCmd(c, ws, shell,
-                                                controlDir.getRemote(),
-                                                binary.getRemote(),
-                                                scriptPath,
-                                                cookieValue,
-                                                cookieVariable);
+        List<String> launcherCmd = null;
+        if (agentInfo.isBinaryCompatible()) {
```
dwnusbaum: I would move the `FORCE_BINARY_WRAPPER` check up here and delete it from the other spot so we do less work in the default case:

 ```suggestion
         if (FORCE_BINARY_WRAPPER && agentInfo.isBinaryCompatible()) {
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: So I just confirmed, i forgot i could have go cross-compile for FreeBSD.  
The OS does have a linux compatability mode, but yeah, it does have some caveats. Probably best to do the freebsd cross-compile if we ever do want to do Freebsd compatibility.

PR #115, opened 2019-10-28
[JENKINS-59903] [JENKINS-59907] Disable golang binary as default

car-roll:
Disabling binary wrapper that was introduced in v1.31 by default. 

Added system property `org.jenkinsci.plugins.durabletask.BourneShellScript.FORCE_BINARY_WRAPPER` to enable the binary. 

#114 will be used to address the bugs.


dwnusbaum:
Ok, I missed that this isn't enough to fix the cache-related issues, since that happens down in `GetAgentInfo.invoke`, before the code that is guarded by the condition changed in this PR. Maybe just add a try/catch around the cache-related stuff and set the binary availability to false if anything happens?

PR #121, opened 2020-02-17
make generate binaries script return error for failures

car-roll:
Currently, when the `generate-binaries` script fails, it does not return a failure code and the entire build process will continue. This is especially important for the unit tests that are assuming the binary is present.

PCT will need to set `SKIP_BINARY_GENERATION=true` in the environment variables. This will skip the docker build, ignore the `binaryCaching` unit test as well as the `UBUNTU_NO_BINARY` test platform

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -81,6 +81,26 @@
             <build>
                 <plugins>
                     <plugin>
+                        <groupId>org.codehaus.mojo</groupId>
+                        <artifactId>properties-maven-plugin</artifactId>
+                        <version>1.0.0</version>
+                        <executions>
+                            <execution>
+                                <goals>
+                                    <goal>set-system-properties</goal>
+                                </goals>
+                                <configuration>
+                                    <properties>
+                                        <property>
+                                            <name>org.jenkinsci.plugins.durabletask.BourneShellScriptTest.TEST_BINARY</name>
+                                            <value>true</value>
+                                        </property>
+                                    </properties>
+                                </configuration>
+                            </execution>
+                        </executions>
+                    </plugin>
+                    <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-jar-plugin</artifactId>
                         <version>3.1.2</version>
--- a/pom.xml
+++ b/pom.xml
@@ -74,32 +74,12 @@
             <id>generate-binaries-profile</id>
             <activation>
                 <property>
-                    <name>env.SKIP_BINARY_GENERATION</name>
+                    <name>env.SKIP_DURABLE_TASK_BINARY_GENERATION</name>
                     <value>!true</value>
                 </property>
             </activation>
             <build>
                 <plugins>
-                    <plugin>
-                        <groupId>org.codehaus.mojo</groupId>
-                        <artifactId>properties-maven-plugin</artifactId>
-                        <version>1.0.0</version>
-                        <executions>
-                            <execution>
-                                <goals>
-                                    <goal>set-system-properties</goal>
-                                </goals>
-                                <configuration>
-                                    <properties>
-                                        <property>
-                                            <name>org.jenkinsci.plugins.durabletask.BourneShellScriptTest.TEST_BINARY</name>
-                                            <value>true</value>
-                                        </property>
-                                    </properties>
-                                </configuration>
-                            </execution>
-                        </executions>
-                    </plugin>
                     <plugin>
                         <groupId>org.apache.maven.plugins</groupId>
                         <artifactId>maven-jar-plugin</artifactId>
--- a/src/main/go/org/jenkinsci/plugins/durabletask/generate-binaries.sh
+++ b/src/main/go/org/jenkinsci/plugins/durabletask/generate-binaries.sh
@@ -1,5 +1,5 @@
 #! /bin/sh
-set -x
+set -ex
 # maven plugin version
 VER=$1
 # path to the golang source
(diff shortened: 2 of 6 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -88,7 +89,11 @@
 public class BourneShellScriptTest {
     @Parameters(name = "{index}: {0}")
     public static Object[] data() {
-        return TestPlatform.values();
+        if (Objects.equals(System.getenv().get("SKIP_BINARY_GENERATION"), "true")) {
```
dwnusbaum: Should we use system properties instead of environment variables? Or is there no easy way for the PCT to specify system properties?

I would use `BourneShellScript.class.GetName() + ". SKIP_BINARY_GENERATION"` as the property/env var name for better traceability wherever it needs to be used.

src/main/go/org/jenkinsci/plugins/durabletask/generate-binaries.sh
```diff
@@ -12,6 +12,11 @@ docker build --build-arg PLUGIN_VER=${VER} -t ${IMG_NAME}:${VER} .
 docker run -i --rm \
     --mount type=bind,src=${SRC},dst=/org/jenkinsci/plugins/durabletask \
     ${IMG_NAME}:${VER}
+docker rmi ${IMG_NAME}:${VER}
 mkdir -p ${DST}
 mv ${BIN_NAME}_* ${DST}/
-docker rmi ${IMG_NAME}:${VER}
\ No newline at end of file
+if [ $? -ne 0 ]
```
dwnusbaum: Like we discussed, I think `set -e` might be better than manually checking errors.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: Would the system property be able to work in the pom  as well? The pom uses that  variable to skip the  binary generation.

src/main/go/org/jenkinsci/plugins/durabletask/generate-binaries.sh
car-roll: addressed in 1e06b6e

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: I added a `TEST_BINARY` system property that gets added in the `generate-binaries-profile` profile of the pom. I was able to confirm that the system property gets applied properly [in a test PR here](https://ci.jenkins.io/job/Plugins/job/durable-task-plugin/job/PR-122/1/testReport/org.jenkinsci.plugins.durabletask/BourneShellScriptTest/). Normally there are 7 platforms to test, but when `TEST_BINARY` is false, the `UBUNTU_NO_BINARY` platform gets removed, and the `binaryCaching` unit tests get skipped.

src/test/java/org/jenkinsci/plugins/durabletask/EncodingTest.java
```diff
@@ -67,23 +68,21 @@
 @RunWith(Parameterized.class)
 public class EncodingTest {
 
-    @ClassRule public static JenkinsRule r = new JenkinsRule();
+    @Rule public LoggerRule logging = new LoggerRule().recordPackage(BourneShellScript.class, Level.FINE);
+    @Rule public JenkinsRule r = new JenkinsRule();
+    @Rule public DockerClassRule<JavaContainer> dockerUbuntu = new DockerClassRule<>(JavaContainer.class);
```
dwnusbaum: I think it would also work if you kept everything as `@ClassRule` and `static`, but combined `unixAndDocker` and `setUp`, which might make the tests a little quicker. My best guess as to the problem is that `setUp` was actually running first (AIUI there is no guaranteed ordering between multiple `@Before(Class)` methods), when we want `unixAndDocker` to run first.

pom.xml
```diff
@@ -80,6 +80,26 @@
...
+                            <execution>
+                                <goals>
+                                    <goal>set-system-properties</goal>
+                                </goals>
+                                <configuration>
+                                    <properties>
+                                        <property>
+                                            <name>org.jenkinsci.plugins.durabletask.BourneShellScriptTest.TEST_BINARY</name>
+                                            <value>true</value>
+                                        </property>
```
dwnusbaum: Ok, looking at this in more detail, I think that the way you were using the environment variable in the test previously made more sense. I don't really see a compelling reason why we'd want to generate binaries but not test them, so I think makes sense for them to be controlled by the same setting.

It looks like we have to use an environment variable to activate `generate-binaries-profile` because there is no way to pass arbitrary Java options to `buildPlugin`, so you are using `withEnv`. If that were not the case, I would say that the profile should be activated by system property and then we'd just use that property in the test.

So in summary I think it would be best to drop `BourneShellScriptTest.TEST_BINARY` here and just look at the environment variable in `BourneShellScriptTest` like you had before. Maybe change the name from `SKIP_BINARY_GENERATION` to `SKIP_DURABLE_TASK_BINARY_GENERATION` or something for clarity if the variable is set in a different context like the PCT.

src/test/java/org/jenkinsci/plugins/durabletask/EncodingTest.java
car-roll: That's actually what I tried as well, but then I was still failing. It wasn't until I realized there was an '@AfterClass' method that I managed to resolve the issues.  I'll try going back to `@ClassRule` and tweaking the `@AfterClass`

PR #127, opened 2020-09-02
Update changelog for 1.35

dwnusbaum:
I am planning on releasing the plugin today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## Changelog
 
+### Version 1.35
+
+Release date: 2020-09-02
+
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -47,19 +58,20 @@ Oct 22, 2019
 -   Allow setting pwsh as Powershell executable
     ([PR \#112](https://github.com/jenkinsci/durable-task-plugin/pull/111))
 -   Bugfix: Use setsid instead of nohup ([JENKINS-25503](https://issues.jenkins-ci.org/browse/JENKINS-25503))
-    - For *NIX systems only, the shell wrapper has been replaced with a pre-compiled golang binary.
+    - For \*NIX systems only, the shell wrapper has been replaced with a pre-compiled golang binary.
```
dwnusbaum, line 61: GitHub's Markdown processor was ok with this as-is, but I think the behavior of this kind of thing is unspecified, and my editor's syntax highlighter did not like it, so it seems best to be explicit. Same for the tildes below.

PR #129, opened 2020-10-16
Store output persistently

PhilippHomann:
I was looking for an option to store the output of some long running commands with massive logging (cmake for example) into a file and also to Jenkins' console output.
Tools like warnings-ng need files to work with.
Build failure Analyzer only works with console output to scan for failure reason.

On linux you can use tee to achive this, but on windows there is no command like this.
I tried [TeeStep](https://www.jenkins.io/doc/pipeline/steps/pipeline-utility-steps/#tee-tee-output-to-file) but it causes really high load on master with massive log output.
That caused odd problems like "Cannot contact node: java.lang.InterruptedException" logs and looping log output and jobs were running much longer because logging was too expensive.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -138,6 +138,11 @@ public final class BourneShellScript extends FileMonitoringTask {
         capturingOutput = true;
     }
 
+    @Override
+    public void storeOutput(String outputFile) {
+        this.outputFile = outputFile;
+    }
+
     @Override protected FileMonitoringController launchWithCookie(FilePath ws, Launcher launcher, TaskListener listener, EnvVars envVars, String cookieVariable, String cookieValue) throws IOException, InterruptedException {
         if (script.isEmpty()) {
             listener.getLogger().println("Warning: was asked to run an empty script");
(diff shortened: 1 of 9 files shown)


PhilippHomann:
@reviewbybees

dwnusbaum:
@PhilippHomann Thanks for the PR! It looks like you accidentally rebased some commits that were already in `master`, can you please clean up your branch's history?

PhilippHomann:
I've already prepared a fix for [workflow-durable-task-step-plugin](https://github.com/PhilippHomann/workflow-durable-task-step-plugin/tree/storeOutput).
What's the best practice to reference the results of this PR in the pom to make builds successful?

dwnusbaum:
@PhilippHomann If you look under "All checks have passed" at the bottom of the PR in GitHub, look for "continuous-integration/jenkins/incrementals", then click details, it takes you to https://repo.jenkins-ci.org/incrementals/org/jenkins-ci/plugins/durable-task/1.36-rc473.4c3740c63ec4/, which contains a build of the plugin based on this PR. You can file a PR to `workflow-durable-task-step` that updates the `durable-task` dependency to `1.36-rc473.4c3740c63ec4` so that CI builds will be able to validate the changes in both PRs. See https://github.com/jenkinsci/jep/blob/master/jep/305/README.adoc for more details on this system.

jglick:
Surely there is _some_ way to do this on Windows? Really do not want to reinvent `tee`.

PhilippHomann:
Haven't found a way to do this with windows internals.
You can do similar things like `tee` with powershell on win10, but nothing for batch and/or win7.

I think the current implementation is also something like `tee`, because you are writing to a file and reading from it to console output.
The only new thing would be the possibility to set the path of the outputfile which is written anyway to a path outside of a temporary directory.

jglick:
I do not think it is wise to add new options to this already complex subsystem of Pipeline merely to support obsolete versions of Windows, when there is a `tee` step which works (if not efficiently) and there are [other native options](https://stackoverflow.com/a/62771772/12916).

PR #130, opened 2021-02-09
Add an option to load the PowerShell profile

ofosos:
Hey Team,
I would like to be able to load the PowerShell profile on the agent. This change is a step in that direction.

Please be kind. I ran the unit tests, but I'm not sure this will work, any ideas on how to test this properly?

Cheers, Mark

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -48,6 +48,7 @@ import org.kohsuke.stapler.DataBoundSetter;
 public final class PowershellScript extends FileMonitoringTask {
     private final String script;
     private String powershellBinary = "powershell";
+    private boolean loadProfile;
     private boolean capturingOutput;
     
     @DataBoundConstructor public PowershellScript(String script) {
@@ -61,6 +62,15 @@ public final class PowershellScript extends FileMonitoringTask {
     @DataBoundSetter
     public void setPowershellBinary(String powershellBinary) {
         this.powershellBinary = powershellBinary;
+    }
+
+    public boolean isLoadProfile() {
+        return loadProfile;
+    }
+
+    @DataBoundSetter
+    public void setLoadProfile(boolean loadProfile) {
+        this.loadProfile = loadProfile;
     }
 
     public String getScript() {
@@ -94,9 +104,12 @@ public final class PowershellScript extends FileMonitoringTask {
 
         String powershellArgs;
         if (launcher.isUnix()) {
-            powershellArgs = "-NoProfile -NonInteractive";
+            powershellArgs = "-NonInteractive";
         } else {
-            powershellArgs = "-NoProfile -NonInteractive -ExecutionPolicy Bypass";
+            powershellArgs = "-NonInteractive -ExecutionPolicy Bypass";
+        }
+        if (!loadProfile) {
+            powershellArgs = "-NoProfile " + powershellArgs;
         }
         args.add(powershellBinary);
         args.addAll(Arrays.asList(powershellArgs.split(" ")));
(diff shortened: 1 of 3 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -48,6 +48,7 @@
 public final class PowershellScript extends FileMonitoringTask {
     private final String script;
     private String powershellBinary = "powershell";
+    private boolean loadProfile = false;
```
timja, line 51: false is the default
```suggestion
    private boolean loadProfile;
```

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -63,6 +64,11 @@ public void setPowershellBinary(String powershellBinary) {
         this.powershellBinary = powershellBinary;
     }
 
+    public boolean isLoadProfile() { return loadProfile; }
```
timja, line 67: use line breaks, to match the code style in the file
```suggestion
    public boolean isLoadProfile() {
        return loadProfile; 
    }
```

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -63,6 +64,11 @@ public void setPowershellBinary(String powershellBinary) {
...
+
+    @DataBoundSetter
+    public void setLoadProfile(boolean loadProfile) { this.loadProfile = loadProfile; }
```
timja, line 70: use line breaks, to match the code style in the file
```suggestion
    public void setLoadProfile(boolean loadProfile) {
        this.loadProfile = loadProfile;
    }
```

src/main/resources/org/jenkinsci/plugins/durabletask/PowershellScript/config.jelly
```diff
@@ -7,4 +7,7 @@
     <f:entry field="powershellBinary" title="Executable">
         <f:select />
     </f:entry>
+    <f:entry field="loadProfile" title="Load Profile">
```
timja, line 10: you could also help add text by creating a file called:
help-loadProfile.html in the same directory

only if you think it's needed though

ofosos:
Hey, @timja I incorporated your changes.

src/main/resources/org/jenkinsci/plugins/durabletask/PowershellScript/config.jelly
ofosos: I think it's rather obvious, what this switch does. So I would go without a help text. But thanks for the pointer!

car-roll:
Looks fine to me. I think you want to add some profile loading tests in https://github.com/jenkinsci/durable-task-plugin/blob/master/src/test/java/org/jenkinsci/plugins/durabletask/PowerShellCoreScriptTest.java

ofosos:
Hey @car-roll, I added to test to see if the flag is present in the current process. That's not really nice, but it looks better than changing profiles on developer machines ;)

Is this sufficient for you?

car-roll:
@ofosos admittedly I don't really know about PowerShell and  profiles. Is the test somehow validating that the profile has been successfully loaded? I just see assertions for script launch success.

ofosos:
@car-roll The thing is, afaik we can't check that the profile has been loaded. Unless we modify the PowerShell profile and say include a variable. However it's kind of bad to modify every developers profile. So what we do instead is this: We spin up a powershell and the PowerShell script checks the flags of the current process (the powershell process) and returns success iff this includes the -NoProfile option. We check that the normal state is success and the return code is failure, when we pass the option to load the profiles.

ofosos:
@car-roll ping

car-roll (approved):

car-roll:
@timja looks like your comments were all addressed in this PR. Anything else you would like to add?

timja (approved):

timja:
Lgtm

PR #131, opened 2021-03-28
Fix JENKINS-59529 - PowerShell command invocation errors will now fail the pipeline step

rkeithhill:
- [ x] Make sure you are opening from a **topic/feature/bugfix branch** (right side) and not your master branch!

https://issues.jenkins.io/browse/JENKINS-59529?jql=text%20~%20%22powershell%22

To fix this issue, I wrapped the invoked command in a `try { <user-supplied-command> } catch { throw }` to ensure that script/command not found errors, parameter not found and parameter validation errors will now fail the pipeline step.  This has been particularly painful when we have folks find - via the logs - steps that "pass" yet the log indicates the script never ran because someone typo'd the script name (or a parameter name).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
@@ -49,7 +49,7 @@ public final class PowershellScript extends FileMonitoringTask {
     private final String script;
     private String powershellBinary = "powershell";
     private boolean capturingOutput;
-    
+
     @DataBoundConstructor public PowershellScript(String script) {
         this.script = script;
     }
@@ -75,10 +75,10 @@ public final class PowershellScript extends FileMonitoringTask {
     @Override protected FileMonitoringController doLaunch(FilePath ws, Launcher launcher, TaskListener listener, EnvVars envVars) throws IOException, InterruptedException {
         List<String> args = new ArrayList<String>();
         PowershellController c = new PowershellController(ws);
-        
+
         String cmd;
         if (capturingOutput) {
-            cmd = String.format(". '%s'; Execute-AndWriteOutput -MainScript '%s' -OutputFile '%s' -LogFile '%s' -ResultFile '%s' -CaptureOutput;", 
+            cmd = String.format(". '%s'; Execute-AndWriteOutput -MainScript '%s' -OutputFile '%s' -LogFile '%s' -ResultFile '%s' -CaptureOutput;",
                 quote(c.getPowerShellHelperFile(ws)),
                 quote(c.getPowerShellScriptFile(ws)),
                 quote(c.getOutputFile(ws)),
@@ -101,12 +101,12 @@ public final class PowershellScript extends FileMonitoringTask {
         args.add(powershellBinary);
         args.addAll(Arrays.asList(powershellArgs.split(" ")));
         args.addAll(Arrays.asList("-Command", cmd));
-        
+
         String scriptWrapper = String.format("[CmdletBinding()]\r\n" +
                                              "param()\r\n" +
                                              "& %s %s -Command '& ''%s''; exit $LASTEXITCODE;';\r\n" +
                                              "exit $LASTEXITCODE;", powershellBinary, powershellArgs, quote(quote(c.getPowerShellScriptFile(ws))));
-     
+
         // Add an explicit exit to the end of the script so that exit codes are propagated
         // Fix https://issues.jenkins.io/browse/JENKINS-59529?jql=text%20~%20%22powershell%22
         // Wrap script in try/catch in order to propagate PowerShell errors like: command/script not recognized,
@@ -119,14 +119,14 @@ public final class PowershellScript extends FileMonitoringTask {
         // One consequence of leaving the "exit $LASTEXITCODE" is if the last native command in a script exits with
         // a non-zero exit code, the step will fail. That may sound obvious but most PowerShell scripters are not used
         // to that. PowerShell doesn't have the equivalent of "set -e" yet. However, in the context of a build system,
-        // I believe we should err on the side of false negatives instead of false positives. If a scripter does not
+        // I believe we should err on the side of false negatives instead of false positives. If a scripter doesn't
         // want a non-zero exit code to fail the step, they can do the following (PS >= v7) to reset $LASTEXITCODE:
         // whoami -f || $($global:LASTEXITCODE = 0)
         String scriptWithExit = "try { " + script + " } catch { throw }\r\nexit $LASTEXITCODE";
-        
+
         // Copy the helper script from the resources directory into the workspace
         c.getPowerShellHelperFile(ws).copyFrom(getClass().getResource("powershellHelper.ps1"));
-        
+
         if (launcher.isUnix() || "pwsh".equals(powershellBinary)) {
             // There is no need to add a BOM with Open PowerShell / PowerShell Core
             c.getPowerShellScriptFile(ws).write(scriptWithExit, "UTF-8");
@@ -140,7 +140,7 @@ public final class PowershellScript extends FileMonitoringTask {
                 writeWithBom(c.getPowerShellWrapperFile(ws), scriptWrapper);
             }
         }
-        
+
         Launcher.ProcStarter ps = launcher.launch().cmds(args).envs(escape(envVars)).pwd(ws).quiet(true);
         ps.readStdout().readStderr();
         Proc p = ps.start();
@@ -149,7 +149,7 @@ public final class PowershellScript extends FileMonitoringTask {
 
         return c;
     }
-    
+
     private static String quote(FilePath f) {
         return quote(f.getRemote());
     }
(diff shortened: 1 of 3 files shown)


rkeithhill:
This is an example of what I'm trying to fix:

image

rkeithhill:
This is an example of what I'm trying to fix:

image

Note the green - for a "passed" step.  :-(

gabloe (approved):

car-roll (approved):
LGTM

bitwiseman (approved):
Looks reasonable.

rkeithhill:
The errors seem intermittent i.e. I've not changed any code and that last time they ran, all tests passed. 🤷‍♂️

car-roll:
@rkeithhill I was able to rerun the tests using the checks tab earlier today.  looks like we still have some instability

rkeithhill:
@car-roll I seemed to have better luck when the test ran late Sunday night.  :-)  I apparently don't have the ability to rerun tests on the checks tab (either that or I'm not seeing the UI for it).  I'll try another comment tweak late tonight to see if the tests pass then.

car-roll:
Rerun CI

rkeithhill:
That's better!

rkeithhill:
@car-roll @bitwiseman Any thoughts on when the next release will happen?

car-roll:
Maybe mid-april?

rkeithhill:
@car-roll Maybe mid-may?  :-)

car-roll:
@rkeithhill my bad, I was trying to get some things in, but looks like it didn't happen. I'll just release now.

doman18:
I just upgraded plugin to version 1.36. I re-run my testing pipeline ("somefile.txt" does not exist) but still returns positive result despite of errors:

`pipeline{
    agent{ label 'windows-optiplex'}
    stages{
        stage('command'){
            steps{
                echo env.Path
                //powershell '$ErrorActionPreference = "Stop";mv somefile.txt somefile2.txt'
                powershell 'mv somefile.txt somefile2.txt'
                echo "second attempt"
                powershell("""
                try
                    {
                        mv somefile.txt somefile2.txt
                    }
                    catch
                    {
                        write-host "Caught an exception"
                        exit 1
                    }
                """)
            }
           
        }
        
        stage('command2'){
            steps{
                powershell returnStatus: true, script: 'mv somefile.txt somefile2.txt'
            }
           
        }
        
        stage('command3'){
            steps{
                script{
                    myfunction()
                }
            }
           
        }
    }
}

def myfunction(){
    //def status = powershell(returnStdout: true, script: 'mv somefile.txt somefile2.txt')
    def status = powershell(returnStatus: true, script: 'mv somefile.txt somefile2.txt')
    if(status != 0){
        error 'Something wrong'
    }
}`

RPitt:
Note that this **breaks perfectly valid powershell** scripts and IMHO should be reverted or fixed asap.

For example:
```
[CmdletBinding()]
Param( [String ]$Param1)
Write-Host "Param1=$Param1"
```
gets turned into:
```
try { [CmdletBinding()]
Param( [String ]$Param1)

Write-Host "Param1=$Param1" } catch { throw }
exit $LASTEXITCODE
```
with the latter being invalid powershell and failing to run.

rkeithhill:
Move-Item generates non-terminating errors by default. Use parameter -ErrorAction Stop to cause a terminating error.

Sent from my Windows 10 device

rkeithhill:
I think we can fix this scenario with: try { & { script }} catch { throw }.   Will need to add to the test suite.

Sent from my Windows 10 device

car-roll:
@RPitt thanks for reporting this bug. Could you please file this bug in the jenkins.io issue tracker? (If you haven't done so already) It's better to have the discussion over there so that other people can find it.

d-panas:
> Move-Item generates non-terminating errors by default. Use parameter -ErrorAction Stop to cause a terminating error.

Hmm ... but  IMHO steps like bat, sh, powershell should consistently return errors by default in any case. And blocking those erros should be an intended action, not default behaviour. So powershell's $ErrorActionPreference = "Stop" or any equivalent solution should be applied by default. Otherwise the default behaviour is very misleading and sometimes even dangerous.

doman18:
> Move-Item generates non-terminating errors by default. Use parameter -ErrorAction Stop to cause a terminating error.

Hmm ... but  IMHO steps like bat, sh, powershell should consistently return errors by default in any case. And blocking those erros should be an intended action, not default behaviour. So powershell's $ErrorActionPreference = "Stop" or any equivalent solution should be applied by default. Otherwise the default behaviour is very misleading and sometimes even dangerous.

rkeithhill:
@doman18 I tend to agree but it might surprise some PowerShell scripters if we go that route.  The only "partial equivalent" of `set -e` in PowerShell is `$ErrorActionPreference = "Stop"` as you point out.  That only affects invocations of PowerShell built-in commands and not native applications.  There is no equivalent of set -e for native applications.  So it is incumbent on the PowerShell script to test `$LASTEXITCODE` after every native app call and throw if non-zero.  You can wrap that up in a function like we do called `Start-NativeExecution`.  

Anyway, the issue with always setting `$ErrorActionPreference = "Stop"` is that folks might be surprised that `dir idontexist.txt` will fail their PS step and pipeline - as would any other PowerShell non-terminating error.  FWIW, we always put `$ErrorActionPreference = "Stop"` near the top of our scripts (right after the param decls).  However, always adding this command has the same problem that @RPitt ran into, you can't put any script above the `[CmdletBinding()]param()` decls.  At least not in the `powershellScript.ps1` file that contains the "script" that is in the `powershell/pwsh` pipeline step.  We might be able to move that up a level. I'm investigating that.

rkeithhill:
@RPitt Yes, `[CmdletBinding()]param([Parameter()][string]$Param1)` is valid PowerShell script but it serves no purpose.  There is no way to pass in the value of `Param1` e.g.
```
steps {
    pwsh '''
[CmdletBinding()]
param(
    [Parameter()]
    [string]
    $Param1
)
"Param1 is $Param1"
'''
}
```
The only other args you can provide except for the script is: `encoding`, `label`, `returnStatus` and `returnStdout`.  You can see this a bit more clearly in the `powershellWrapper.ps1` script that invoked the script of the pwsh step:
```powershell
[CmdletBinding()]
param()
& pwsh -NoProfile -NonInteractive -Command '& ''/home/blduser/builds/cryptopp/main@tmp/durable-d343052c/powershellScript.ps1''; exit $LASTEXITCODE;';
exit $LASTEXITCODE;
```
And don't ask me why there are two `exit $LASTEXITCODE` statements. I didn't touch this wrapper script.  But you can see there are no args supplied to `powershellScript.ps1`  - which currently looks like this:
```powershell
try {
[CmdletBinding()]
param(
    [Parameter()]
    [string]
    $Param1
)
"Param1 is $Param1"
} catch { throw }
exit $LASTEXITCODE
```
Since there is no way to pass in a value to Param1, I think I could fix this with by changing the generated script to:
```powershell
try {
    & {
        [CmdletBinding()]
        param(
            [Parameter()]
            [string]
            $Param1
        )
        "Param1 is $Param1"
    }
} catch { throw }
exit $LASTEXITCODE
```
I'm also going to look into moving changes out of `powershellScript.ps1` and into `powerhellWrapper.ps1`.  

BTW I would like to point out that yes, while you can't specify `[CmdletBinding()]param()` at the moment, the powershell step will at least now fail on:
```
pwsh ".\scriptDoesNotExist.ps1"
pwsh ".\build.ps1 -ParamDoesNotExist"
pwsh ".\build.ps1 -BuildType InvalidBuildType # should fail parameter validation but doesn't
```
The above "false positives" were about to get the use of PowerShell "kicked out" of our pipeline builds.  :-(  That said, it seems that PowerShell parse errors are still reporting false positives.  I think we can fix that as well.

rkeithhill:
OK I think I have a fix that addresses @RPitt's issue **and** catches parsing errors.  I'm going to open up an issue on jenkins.io to track this and associate with the PR.  I need to add some more tests to cover these two scenarios `[CmdletBinding()]` and parse errors but I "think" the fix is to revert the `powershellScript.ps1` change and handle this in `powershellWrapper.ps1` changing from the current script:
```powershell
[CmdletBinding()]
param()
& pwsh -NoProfile -NonInteractive -Command '& ''/home/blduser/builds/cryptopp/main@tmp/durable-d343052c/powershellScript.ps1''; exit $LASTEXITCODE;';
exit $LASTEXITCODE;
```
to
```powershell
[CmdletBinding()]
param()
& pwsh -NoProfile -NonInteractive -Command '& { try { "/home/blduser/builds/cryptopp/main@tmp/durable-d343052c/powershellScript.ps1" } catch { throw }}; exit $LASTEXITCODE'
exit $LASTEXITCODE
```
This does not set `$ErrorActionPreference = 'Stop'`.  I'm a little nervous about that breaking even more folks. And it is somewhat orthogonal to the original Jenkins issue this PR fixed.

rkeithhill:
PR #134 has been created to fix this.

rkeithhill:
Also, [here ](https://ci.jenkins.io/job/Plugins/job/durable-task-plugin/job/PR-134/4/artifact/org/jenkins-ci/plugins/durable-task/1.37-rc478.bb1a2b917f1f/) is a link to a prerelease version of the plugin with this fix in place?  @RPitt (or anyone) is there any chance you could test it?

PR #132, opened 2021-04-07
use `cmd.exe` when launching the script

jtnord:
`cmd.exe` is always `cmd.exe` and never `cmd.com` or `cmd.bat` or any other valid `PATH_EXT` so use it to launch the outer process.

Whilst I probably should have fixed the other occurence (used to spawn the innder batch script) I am attempting to diagnose an opaque error and having a difference between the outer and inter script calling is helpful.  (hence the addition of the extra logs also!)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -30,14 +30,19 @@ import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
 import hudson.Proc;
+import hudson.model.Computer;
 import hudson.model.TaskListener;
 import java.io.IOException;
+import java.util.logging.Logger;
 import org.kohsuke.stapler.DataBoundConstructor;
 
 /**
  * Runs a Windows batch script.
  */
 public final class WindowsBatchScript extends FileMonitoringTask {
+
+    private final static Logger LOGGER = Logger.getLogger(WindowsBatchScript.class.getName());
+    
     private final String script;
     private boolean capturingOutput;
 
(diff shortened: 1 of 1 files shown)


alecharp (approved):

rsandell (approved):

MRamonLeon (approved):

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -76,7 +81,12 @@ public String getScript() {
         c.getBatchFile1(ws).write(cmd, "UTF-8");
         c.getBatchFile2(ws).write(script, "UTF-8");
 
-        Launcher.ProcStarter ps = launcher.launch().cmds("cmd", "/c", "\"\"" + c.getBatchFile1(ws) + "\"\"").envs(escape(envVars)).pwd(ws).quiet(true);
+        LOGGER.fine(() -> {
+            Computer computer = ws.toComputer();
+            String computerName = (computer == null) ? "unknown" : (computer.getName() + ":" + computer.getDisplayName());
+            return "Attempting to launch durable batch script on " + computerName + " with environment " + envVars.toString();
```
car-roll, line 87: I'm most worried about `envVars.toString()` since I think we store pipeline secrets in `envVars`. Wouldn't this expose them all?

jtnord: to system logs yes.   System logs are in Jenkins only readable by Admins who have access to these credentials / secrets anyway. 
If you are punting System logs to an external system for storage, or allowing read access to users other than root or the jenkins user on the system, then an argument could be made that is not Jenkins' concern. (this is FINE logging after all)
If you are interacting with Vault / AWS / K8s as your secret provider or the like then you can just enable finest logging on the HTTP client in use and that will most likely log all the responses too.

Anyway - this does not need to be merged, this is here as a reference in an attempt to fix a specific issue in a specific environment where the error that comes back is `cmd is not a valid application` when in fact it is on the path and is a valid application.

jtnord:
I think this can probably be closed.  WDYT @car-roll

PR #135, opened 2021-05-15
[JENKINS-65413] Move binary generation to new library

car-roll:
Binary generation during the build process is a somewhat complex and brittle process that is linux-only. This PR moves the binary generation and the related [`go` code](https://github.com/jenkinsci/durable-task-plugin/tree/durable-task-1.36/src/main/go/org/jenkinsci/plugins/durabletask) to [`lib-durable-task`](https://github.com/jenkinsci/lib-durable-task). 

This should also make things easier for CI to actually run the tests 😓 


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -486,6 +488,14 @@ public AgentInfo invoke(File nodeRoot, VirtualChannel virtualChannel) throws IOE
             } else {
                 binaryCompatible = false;
             }
+            String archType = "";
+            if (os == OsType.DARWIN) {
+                if (arch.contains("aarch") || arch.contains("arm")) {
```
car-roll, line 493: Without having any apple silicon machines to test on, I did my best to figure out what `os.arch` would actually contain

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -480,21 +482,37 @@ private Slave prepareDockerPlatforms() throws Exception {
     @Test public void binaryCaching() throws Exception {
         assumeTrue(!Objects.equals(System.getenv().get("SKIP_DURABLE_TASK_BINARY_GENERATION"), "true") && !platform.equals(TestPlatform.UBUNTU_NO_BINARY));
```
dwnusbaum, line 483: Should we get rid of the stuff related to the environment variable now? Is there any other code that uses it?

pom.xml
```diff
@@ -24,7 +24,7 @@
         <changelist>-SNAPSHOT</changelist>
         <jenkins.version>2.150.1</jenkins.version>
         <java.level>8</java.level>
-        <binary.src>src/main/go/org/jenkinsci/plugins/durabletask</binary.src>
+        <lib-durable-task.version>1.0</lib-durable-task.version>
```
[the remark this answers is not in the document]
jglick: True, though I would not expect a lot of changes.

Jenkinsfile
```diff
@@ -1,6 +1,4 @@
 // TODO: Come up with a way to enable binary generation on Linux but not
 // on Windows without needing to run buildPlugin multiple times to avoid
 // https://github.com/jenkinsci/durable-task-plugin/pull/126#issuecomment-681964838.
```
jglick, line 3:
```suggestion
```

Jenkinsfile
```diff
@@ -1,6 +1,4 @@
...
-withEnv(['SKIP_DURABLE_TASK_BINARY_GENERATION=true']) {
-  buildPlugin(platforms: ['docker', 'maven-windows'])
-}
+buildPlugin(platforms: ['docker', 'maven-windows'])
```
jglick, line 4:
```suggestion
buildPlugin(platforms: ['docker', 'maven-windows'])
```

(trailing newline)

pom.xml
```diff
@@ -47,6 +47,11 @@
         <tag>${scmTag}</tag>
     </scm>
     <dependencies>
+        <dependency>
+            <groupId>io.jenkins.plugins</groupId>
+            <artifactId>lib-durable-task</artifactId>
+            <version>${lib-durable-task.version}</version>
```
jglick, line 53: No reason to use a POM version unless you are referring to it at least twice.

pom.xml
```diff
@@ -69,55 +74,35 @@
...
-                        <version>1.6.0</version>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>3.1.2</version>
+                <executions>
+                    <execution>
+                        <id>extract-binaries</id>
```
jglick, line 85: Why bother with all this? You can simply declare a dep on that JAR and it will be bundled inside the plugin and its resources accessible. (Though I would then suggest making the upstream project package binaries into a reasonable package prefix.)

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
jglick: You can look it up online.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
jglick:
```suggestion
        assumeFalse(platform.equals(TestPlatform.UBUNTU_NO_BINARY));
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: Yeah, i checked online and was pretty sure its `aarch64`, just wanted it to be known that I never actually tested it 😛

pom.xml
car-roll: So I did all this to basically rename the binary file. The binary name is `durable_task_monitor_VER_ARCH_BITS`. I was replacing the VER with the `durable_task` VER. 
I believe I was thinking I wanted to people to know which version of durable-task was running the binary. That's probably not necessary since users will just look at their jenkins instance to figure out what's going on.

car-roll: Oh yes, there was also the question of how to get the version of `lib-durable-task` since it's not a plugin and doesn't really have any java code. I could parse the pom, I guess.

jglick: > how to get the version of `lib-durable-task`

Why would you need to?

```java
URL bin = BourneShellScript.class.getResource("/io/jenkins/plugins/lib-durable-task/durable_task_monitor_amd64");
```

Am I missing something?

car-roll: no you're not...i totally got lost in the weeds for this 👀

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: fixed in 388a7bf1c9a1ba7c8e00f1b367c01daaa95b1c0a

Jenkinsfile
car-roll: fixed in 388a7bf1c9a1ba7c8e00f1b367c01daaa95b1c0a

Jenkinsfile
car-roll: fixed in 388a7bf1c9a1ba7c8e00f1b367c01daaa95b1c0a

pom.xml
car-roll: and of course I just belatedly realized that I can probably use `getResourceAsStream` and rename the binary there as well.

pom.xml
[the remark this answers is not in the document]
car-roll: Moved `lib-durable-task` to automatic releases: see https://github.com/jenkinsci/lib-durable-task/pull/16

pom.xml
car-roll: Only one instance used now in c658ad81067e362c9fb0c9d51763d3232a97a6f2

pom.xml
car-roll: Resolved in c658ad81067e362c9fb0c9d51763d3232a97a6f2
Completely removed this section of the POM and use the package resource in `BourneShellScript`. No more messing with versions and filenames

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -416,22 +417,37 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
         private static final long serialVersionUID = 7599995179651071957L;
         private final OsType os;
         private final String binaryPath;
+        private String architecture;
```
jglick, line 420:
```suggestion
        private final String architecture;
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -416,22 +417,37 @@ private int recordExitStatus(FilePath workspace, int code) throws IOException, I
...
 
+        // Avoid deserialization issues with newly added fields
+        private Object readResolve() {
+            if (architecture == null || architecture.isEmpty()) {
+                String[] splits = binaryPath.split("_");
+                architecture = splits[splits.length - 1];
+            }
+            return this;
+        }
+
```
jglick, line 442:
```suggestion
```

Only relevant for objects serialized persistently, such as to `config.xml` or `program.dat`. This class is only ever sent via Remoting, so we need not worry about version skew.

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: made final in dd9324a8e1e9ed44b845c57cd9d3ddac2ea68939

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
car-roll: removed in dd9324a8e1e9ed44b845c57cd9d3ddac2ea68939

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -228,6 +222,7 @@ private Slave prepareDockerPlatforms() throws Exception {
                 throw new AssertionError(platform);
         }
 
+        String arch = System.getProperty("os.arch");
```
dwnusbaum, line 225: I think this is unused, right?

```suggestion
```

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -478,23 +473,38 @@ private Slave prepareDockerPlatforms() throws Exception {
     }
 
     @Test public void binaryCaching() throws Exception {
-        assumeTrue(!Objects.equals(System.getenv().get("SKIP_DURABLE_TASK_BINARY_GENERATION"), "true") && !platform.equals(TestPlatform.UBUNTU_NO_BINARY));
+        assumeFalse(platform.equals(TestPlatform.UBUNTU_NO_BINARY));
         String os;
+        String architecture;
+        String arch = System.getProperty("os.arch");
```
dwnusbaum, line 479: nit: It's kind of confusing having variables named both `arch` and `architecture`. IIUC, `architecture` is used for binary selection and `arch` is just used to figure out if we are on an ARM-based Mac. If that's right, I might move `String arch = System.getProperty("os.arch");` down right after `os = "darwin";` below so it's clearer that it is only used for the ARM check.

dwnusbaum (approved):
Looks ok to me, added a few minor comments.

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: you're right, removed in fcab5fac0547fd470de5427ad1b6832cc3154614

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
car-roll: hopefully un-confused in fcab5fac0547fd470de5427ad1b6832cc3154614

PR #136, opened 2021-05-20
Fix batch line endings to Windows (CRLF)

dsolovyev:
Fixes the same issue with wrong (Unix) line endings in .bat scripts as in [JENKINS-7478](https://issues.jenkins.io/browse/JENKINS-7478) for pipeline `bat` step.
The idea for the fix of `WindowsBatchScript` is taken from https://github.com/jenkinsci/jenkins/pull/1370

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
@@ -40,6 +40,7 @@ import org.junit.BeforeClass;
 import org.junit.Rule;
 import org.jvnet.hudson.test.Issue;
 import org.jvnet.hudson.test.JenkinsRule;
+import java.util.Arrays;
 
 public class WindowsBatchScriptTest {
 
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
```diff
@@ -132,4 +133,35 @@ private void testWithPath(String path) throws Exception {
...
+            ":zwei\n" +
+            "echo zwei\n" +
+            "goto :eins\n";
+        Controller c = new WindowsBatchScript(script).launch(new EnvVars(), ws, launcher, listener);
+        while (c.exitStatus(ws, launcher, listener) == null) {
+            Thread.sleep(100);
+        }
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        c.writeLog(ws, baos);
+        assertEquals(0, c.exitStatus(ws, launcher, listener).intValue());
```
dwnusbaum, line 163: What happens before the fix? Just a non-zero exit code here?

dwnusbaum (approved):
Thanks for the PR! This looks ok as far as I understand.

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
dsolovyev: It sets exit code to 1 and prints error message that cmd can't find first label (`eins`).
I believe we shouldn't check the message since it's locale-specific.
```bat
> cmd /c call test-crlf.bat
zwei
The system cannot find the batch label specified - eins

> echo %errorlevel%
1
```

car-roll (approved):

PR #12, opened 2015-12-02
[JENKINS-27419] Properly handle scripts that call exit without /b

jglick:
[JENKINS-27419](https://issues.jenkins-ci.org/browse/JENKINS-27419)

@reviewbybees esp. @jtnord

PR #137, opened 2021-05-21
Properly handle batch scripts that call exit /b after an error

dsolovyev:
Fixed case when `exit /b` is used to return from `bat` step after an error produces SUCCESS result:
```groovy
node('windows') {
    bat '''
        cmd /c exit 42
        exit /b
    '''
}
```
```bat
Started by user Solovyev, Dmitriy
Running in Durability level: MAX_SURVIVABILITY
[Pipeline] Start of Pipeline
[Pipeline] node
Running on *********** in c:\******\jenkins\workspace\test-exit-b-bat-pipeline
[Pipeline] {
[Pipeline] bat

*************@*********** c:\******\jenkins\workspace\test-exit-b-bat-pipeline>cmd /c exit 42 

*************@*********** c:\******\jenkins\workspace\test-exit-b-bat-pipeline>exit /b 
[Pipeline] }
[Pipeline] // node
[Pipeline] End of Pipeline
Finished: SUCCESS
```

Such issue in "Freestyle" jobs has been fixed quite long ago in https://github.com/jenkinsci/jenkins/commit/12a83871398ed03a677b8e7bb985786d9ffd9924.

Related to [JENKINS-27419](https://issues.jenkins.io/browse/JENKINS-27419)/https://github.com/jenkinsci/durable-task-plugin/pull/12 and [JENKINS-25678](https://issues.jenkins.io/browse/JENKINS-25678)/https://github.com/jenkinsci/durable-task-plugin/pull/18.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
@@ -62,13 +62,13 @@ public final class WindowsBatchScript extends FileMonitoringTask {
 
         String cmd;
         if (capturingOutput) {
-            cmd = String.format("@echo off \r\ncmd /c \"\"%s\"\" > \"%s\" 2> \"%s\"\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+            cmd = String.format("@echo off \r\ncmd /c call \"%s\" > \"%s\" 2> \"%s\"\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
                 quote(c.getBatchFile2(ws)),
                 quote(c.getOutputFile(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)));
         } else {
-            cmd = String.format("@echo off \r\ncmd /c \"\"%s\"\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+            cmd = String.format("@echo off \r\ncmd /c call \"%s\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
                 quote(c.getBatchFile2(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)));
--- a/src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
@@ -105,6 +105,17 @@ public class WindowsBatchScriptTest {
         c.cleanup(ws);
     }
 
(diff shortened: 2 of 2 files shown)


dsolovyev:
Hi @dwnusbaum, can you please review this PR?

dwnusbaum (commented):
@dsolovyev I am not familiar enough with Windows batch scripts to really review this directly. It looks ok to me, and from what I can tell we have regression tests for the related issues you linked so it shouldn't break old code, but if it were up to me I would probably want to spend some time reading through the related issues in more detail before merging.

dsolovyev:
OK, thanks. By the way I fixed the regression tests in my first commit.

car-roll (approved):
LGTM. I did my best to look at the comments and context of the changes and it does seem to make sense, especially since freestyle appears to be calling the scripts in the same way.

bitwiseman (approved):

car-roll:
thanks @bitwiseman !

jglick:
https://github.com/jenkinsci/credentials-binding-plugin/pull/136 demonstrates that this change causes a test regression on Windows. Whatever https://github.com/jenkinsci/credentials-binding-plugin/blob/a84e243a4d6170a5dbc8d27cba91b413bf296ab3/src/test/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest/basicsPipeline-step1.bat#L1-L10 is supposed to do, https://github.com/jenkinsci/credentials-binding-plugin/blob/a84e243a4d6170a5dbc8d27cba91b413bf296ab3/src/test/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest/basicsPipeline-Jenkinsfile#L13 used to pass and now it fails with

```
ERROR: script returned exit code 3
```

Is this change of behavior intentional? If so, how should the batch script be corrected?

car-roll:
@jglick No, this change of behavior is definitely not intentional. Looking at the batch script you linked, I would only expect a 1 or a 0 as a return code. I"ll poke around a bit since I'm able to reproduce this locally as well.

dsolovyev:
There is a bug in [basicsPipeline-step1.bat](https://github.com/jenkinsci/credentials-binding-plugin/blob/a84e243a4d6170a5dbc8d27cba91b413bf296ab3/src/test/resources/org/jenkinsci/plugins/credentialsbinding/impl/CertificateMultiBindingTest/basicsPipeline-step1.bat) - forgotten quotes around the argument of `if not exist`.

Hence the return code is 3:
```bat
>if not exist C:\Jenkins\workspace\redentials-binding-plugin_PR-136\target\tmp\j h7596393474942004483\workspace\p exit /b 1
The system cannot find the path specified.

>echo %errorlevel%
3
```
This patch just helps it to be caught.

jglick:
Glad to have caught that. I am considering just disabling the Windows CI builds, though—the agents are too flaky.

PR #139, opened 2021-06-08
[JENKINS-27617] Add support for windows binary

car-roll:
Adding support for windows binary to launch powershell and batch scripts in a new process group.
Downstream of https://github.com/jenkinsci/lib-durable-task/pull/7

The following system properties must be added to the `jenkins.xml` file in the windows installation path
`-Dorg.jenkinsci.plugins.durabletask.WindowsBatchScript.FORCE_BINARY_WRAPPER=true`
`-Dorg.jenkinsci.plugins.durabletask.PowershellScript.FORCE_BINARY_WRAPPER=true`

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
@@ -121,7 +121,7 @@ public final class AgentInfo implements Serializable {
                 }
             }
 
-            // Note: This will only determine the architecture bits of the JVM. The result will be "32" or "64".
+            // Note: This will only determine the architecture bits of the JVM.
             String bits = System.getProperty("sun.arch.data.model");
             if (bits.equals("64") || bits.equals("32")) {
                 archType += bits;
@@ -136,7 +136,12 @@ public final class AgentInfo implements Serializable {
                 binaryCompatible = false;
             }
 
-            String binaryName = BINARY_PREFIX + binaryVersion + "_" + os.getNameForBinary() + "_" + archType;
+            String extension = "";
+            if (os == OsType.WINDOWS) {
+                extension = ".exe";
+            }
+
+            String binaryName = BINARY_PREFIX + binaryVersion + "_" + os.getNameForBinary() + "_" + archType + extension;
             String binaryPath;
             boolean isCached;
             boolean cachingAvailable;
--- a/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
@@ -6,6 +6,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
+import jenkins.MasterToSlaveFileCallable;
 import org.jenkinsci.remoting.RoleChecker;
 
 import hudson.Platform;
@@ -75,7 +76,7 @@ public final class AgentInfo implements Serializable {
         return cachingAvailable;
     }
 
-    public static final class GetAgentInfo implements FileCallable<AgentInfo> {
+    public static final class GetAgentInfo extends MasterToSlaveFileCallable<AgentInfo> {
         private static final long serialVersionUID = 1L;
         private static final String BINARY_PREFIX = "durable_task_monitor_";
         private static final String CACHE_PATH = "caches/durable-task/";
@@ -159,11 +160,6 @@ public final class AgentInfo implements Serializable {
             AgentInfo agentInfo = new AgentInfo(os, archType, binaryCompatible, binaryPath, cachingAvailable);
             agentInfo.setBinaryAvailability(isCached);
             return agentInfo;
-        }
-
-        @Override
-        public void checkRoles(RoleChecker roleChecker) throws SecurityException {
-
         }
     }
 }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
@@ -1,0 +1,155 @@
+package org.jenkinsci.plugins.durabletask;
+import java.io.File;
+import java.io.IOException;
+import java.io.Serializable;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+
+import org.jenkinsci.remoting.RoleChecker;
+
+import hudson.Platform;
+import hudson.FilePath.FileCallable;
+import hudson.remoting.VirtualChannel;
+
+public final class AgentInfo implements Serializable {
+    private static final long serialVersionUID = 7599995179651071957L;
+    private final OsType os;
+    private final String binaryPath;
+    private final String architecture;
+    private boolean binaryCompatible;
+    private boolean binaryCached;
+    private boolean cachingAvailable;
+
+    public enum OsType {
+        // NOTE: changes to these binary names must be mirrored in lib-durable-task
+        DARWIN("darwin"),
+        LINUX("linux"),
+        WINDOWS("windows"),
+        FREEBSD("freebsd"),
+        ZOS("zos");
+
+        private final String binaryName;
+        OsType(final String binaryName) {
+            this.binaryName = binaryName;
+        }
+        public String getNameForBinary() {
+            return binaryName;
+        }
+    }
+
+    public AgentInfo(OsType os, String architecture, boolean binaryCompatible, String binaryPath, boolean cachingAvailable) {
+        this.os = os;
+        this.architecture = architecture;
+        this.binaryPath = binaryPath;
+        this.binaryCompatible = binaryCompatible;
+        this.binaryCached = false;
+        this.cachingAvailable = cachingAvailable;
+    }
+
+    public OsType getOs() {
+        return os;
+    }
+
+    public String getArchitecture() {
+        return architecture;
+    }
+
+    public String getBinaryPath() {
+        return binaryPath;
+    }
+
+    public void setBinaryAvailability(boolean isCached) {
+        binaryCached = isCached;
+    }
+
+    public boolean isBinaryCompatible() {
+        return binaryCompatible;
+    }
+
+    public boolean isBinaryCached() {
+        return binaryCached;
+    }
+
+    public boolean isCachingAvailable() {
+        return cachingAvailable;
+    }
+
+    public static final class GetAgentInfo implements FileCallable<AgentInfo> {
+        private static final long serialVersionUID = 1L;
+        private static final String BINARY_PREFIX = "durable_task_monitor_";
+        private static final String CACHE_PATH = "caches/durable-task/";
+        // Version makes sure we don't use an out-of-date cached binary
+        private String binaryVersion;
+
+        GetAgentInfo(String pluginVersion) {
+            this.binaryVersion = pluginVersion;
+        }
+
+        @Override
+        public AgentInfo invoke(File nodeRoot, VirtualChannel virtualChannel) throws IOException, InterruptedException {
+            OsType os;
+            if (Platform.isDarwin()) {
+                os = OsType.DARWIN;
+            } else if (Platform.current() == Platform.WINDOWS) {
+                os = OsType.WINDOWS;
+            } else if(Platform.current() == Platform.UNIX && System.getProperty("os.name").equals("z/OS")) {
+                os = OsType.ZOS;
+            } else if(Platform.current() == Platform.UNIX && System.getProperty("os.name").equals("FreeBSD")) {
+                os = OsType.FREEBSD;
+            } else {
+                os = OsType.LINUX; // Default Value
+            }
+
+            String arch = System.getProperty("os.arch");
+            boolean binaryCompatible;
+            if ((os != OsType.FREEBSD) && (arch.contains("86") || arch.contains("amd64"))) {
+                binaryCompatible = true;
+            } else {
+                binaryCompatible = false;
+            }
+            String archType = "";
+            if (os == OsType.DARWIN) {
+                if (arch.contains("aarch") || arch.contains("arm")) {
+                    archType = "arm";
+                } else {
+                    archType = "amd"; // Default Value
+                }
+            }
+
+            // Note: This will only determine the architecture bits of the JVM. The result will be "32" or "64".
+            String bits = System.getProperty("sun.arch.data.model");
+            if (bits.equals("64")) {
+                archType += "64";
+            } else {
+                archType += "32"; // Default Value
+            }
+
+            String binaryName = BINARY_PREFIX + binaryVersion + "_" + os.getNameForBinary() + "_" + archType;
+            String binaryPath;
+            boolean isCached;
+            boolean cachingAvailable;
+            try {
+                Path cachePath = Paths.get(nodeRoot.toPath().toString(), CACHE_PATH);
+                Files.createDirectories(cachePath);
+                File binaryFile = new File(cachePath.toFile(), binaryName);
+                binaryPath = binaryFile.toPath().toString();
+                isCached = binaryFile.exists();
+                cachingAvailable = true;
+            } catch (Exception e) {
+                // when the jenkins agent cache path is not accessible
+                binaryPath = binaryName;
+                isCached = false;
+                cachingAvailable = false;
+            }
+            AgentInfo agentInfo = new AgentInfo(os, archType, binaryCompatible, binaryPath, cachingAvailable);
+            agentInfo.setBinaryAvailability(isCached);
+            return agentInfo;
+        }
+
+        @Override
+        public void checkRoles(RoleChecker roleChecker) throws SecurityException {
+
+        }
+    }
+}
--- a/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
@@ -22,12 +22,12 @@ public final class AgentInfo implements Serializable {
     private boolean cachingAvailable;
 
     public enum OsType {
-        // NOTE: changes to these binary names must be mirrored in lib-durable-task
         DARWIN("darwin"),
         LINUX("linux"),
-        WINDOWS("windows"),
+        WINDOWS("win"),
         FREEBSD("freebsd"),
-        ZOS("zos");
+        ZOS("zos"),
+        UNKNOWN("unknown");
 
         private final String binaryName;
         OsType(final String binaryName) {
@@ -79,6 +79,7 @@ public final class AgentInfo implements Serializable {
         private static final long serialVersionUID = 1L;
         private static final String BINARY_PREFIX = "durable_task_monitor_";
         private static final String CACHE_PATH = "caches/durable-task/";
+        private static final String NOT_SUPPORTED = "NOT_SUPPORTED";
         // Version makes sure we don't use an out-of-date cached binary
         private String binaryVersion;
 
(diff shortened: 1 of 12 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
```diff
@@ -246,6 +225,8 @@
     }
 
     @Test public void noStdoutPollution() throws Exception {
+        // TODO? Binary wrapper does not allow redirection of non-success streams to the logger
```
car-roll, line 228: So Powershell can only re-direct the non-success streams (`2` - `6`) to a separate file (each stream must have a separate file), or all to the `success` stream (`1`). Additionally, the default behavior is that these non-error streams get merged into the `success` stream when a script is launched via `-File` or `-Command`. I don't see a clean way of separating `verbose, warning, debug, etc` from the `success` stream and putting them into the log. I could mirror anything in output to be put into the log, but I think that defeats the purpose of this test. I'm comfortable with documenting the behavior of the binary as diverging from the script wrapper for this particular case as it is the only deviation I know of.

I believe `powershellhelper.ps1` was able to get by this restriction by having special conditions, such as catching errors and then writing them to log, but I don't think it's worth that complexity.

If anyone has any suggestions, I'm more than open to hear them.

src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
```diff
@@ -0,0 +1,169 @@
+package org.jenkinsci.plugins.durabletask;
```
car-roll, line 1: `AgentInfo` was just pulled out of `BourneShellScript` so that it could be used more generically by other classes.

src/test/java/org/jenkinsci/plugins/durabletask/PowershellScriptTest.java
jakauppila: We rely pretty heavily on having a clean stdout as we have often generate a JSON payload within the script and Write-Output to pass it back up. So with that we utilize the verbose stream for logging during the execution to know what's going on.

car-roll: I think I managed a clean stdout now. Unfortunately, through `go`, the non-error streams were redirected to stdout, but if I did I piped the output to a file, I could get a clean output. Unfortunately that required some bom-less file writing tricks, but I think it's good now.

bitwiseman (approved):

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -59,34 +86,100 @@ public String getScript() {
...
+        if (durablePlugin == null) {
+            throw new IOException("Unable to find durable task plugin");
+        }
+        String pluginVersion = StringUtils.substringBefore(durablePlugin.getVersion(), "-");
+        AgentInfo agentInfo = nodeRoot.act(new AgentInfo.GetAgentInfo(pluginVersion));
         BatchController c = new BatchController(ws);
 
-        String cmd;
+        List<String> launcherCmd = null;
+        if (FORCE_BINARY_WRAPPER && agentInfo.isBinaryCompatible()) {
```
dwnusbaum, line 101: Could some of this logic be extracted into `FileMonitoringTask` and reused across the different subclasses? Or are there too many differences?

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -29,18 +29,45 @@
...
 import hudson.util.LineEndingConversion;
+import jenkins.model.Jenkins;
 
 /**
  * Runs a Windows batch script.
  */
 public final class WindowsBatchScript extends FileMonitoringTask {
+    @SuppressFBWarnings("MS_SHOULD_BE_FINAL") // Used to control usage of binary or shell wrapper
+    @Restricted(NoExternalUse.class)
+    public static boolean FORCE_BINARY_WRAPPER = Boolean.getBoolean(WindowsBatchScript.class.getName() + ".FORCE_BINARY_WRAPPER");
```
dwnusbaum, line 58: Nice, I definitely think it's a good idea to use a different property for each step.

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
```diff
@@ -167,13 +169,29 @@ private void testWithPath(String path) throws Exception {
...
+        String log = baos.toString("UTF-8");
+        assertTrue(log, log.contains("Helló, Wõrld ®"));
         c.cleanup(ws);
     }
 
+    private void awaitCompletion(Controller c) throws IOException, InterruptedException {
+        while (c.exitStatus(ws, launcher, listener) == null) {
+            Thread.sleep(100);
+        }
+        Thread.sleep(3000); // Pause allows slower systems to exit binary before cleanup is attempted
```
dwnusbaum, line 194: This is kind of a long sleep, do things flake if this is reduced to one second or less? Or could you directly check for the process to exit or attempt to delete a file used by the binary in a loop instead and exit once it succeeds?

car-roll: So I know there is a retry delete attempt somewhere in either the test package or FilePath and that was looping too fast. I tried getting the sleep down to 1.5 secs, but that's still not working. I think I will need to directly check the process or add in some other retry in cleanup

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
car-roll: Was about to move some of the logic out to `FileMonitoringTask`. A lot of code was shared, the only issue was that each shell type was using it in a slightly unique way 😛 
See c224d6ae001d22c984829428032465285f38bec1 for changes

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
car-roll: So there is[ a lot of tuning](https://github.com/jenkinsci/jenkins/blob/2f612fcac9194419e4a73bc2ee5b7ef422b9d6dd/core/src/main/java/hudson/Util.java#L1767) that you can do in `hudson.Util` to handle the `deleteRecursive` that's failing. The only problem is that they are all `@Restricted(value = NoExternalUse.class)`.

car-roll: Okay, added additional retry logic in `FileMonitoringTask.cleanup()`. Was able to remove sleep in favor of this behavior.

see d7d4517

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -25,31 +25,53 @@
...
 
+import javax.annotation.Nonnull;
+
 /**
  * Runs a Powershell script
  */
 public final class PowershellScript extends FileMonitoringTask {
+    @SuppressFBWarnings("MS_SHOULD_BE_FINAL") // Used to control usage of binary or shell wrapper
+    @Restricted(NoExternalUse.class)
+    public static boolean FORCE_BINARY_WRAPPER = Boolean.getBoolean(PowershellScript.class.getName() + ".FORCE_BINARY_WRAPPER");
```
jtnord, line 59: side note, I would have expected this to be true so we can remove the old code in a future version (and have this incase compatability is required) - esp if this is required to actually work correctly in environments.

Are we planning on changing this to default to true in the future?

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -83,9 +105,105 @@ public String getScript() {
...
+        List<String> launcherCmd = null;
+        FilePath binary;
+        // Binary does not support pwsh on linux
+        boolean pwshLinux;
+        if ((agentInfo.getOs() == AgentInfo.OsType.LINUX) && "pwsh".equals(powershellBinary)) {
+            pwshLinux = true;
+        } else {
+            pwshLinux = false;
+        }
+        if (FORCE_BINARY_WRAPPER && !pwshLinux && (binary = requestBinary(nodeRoot, agentInfo, ws, c)) != null) {
```
jtnord, line 135: NIT: its not "FORCING" if it is conditional on something else...
"FORCE" is I don't care about the environment or any other factors I am doing this regardless.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
jtnord: NIT: this is missnamed - it does not FORCE anything, it is just "preferring" something if other conditions are also met.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -59,34 +86,79 @@ public String getScript() {
...
+        String logFile = c.getLogFile(ws).getRemote();
+        String resultFile = c.getResultFile(ws).getRemote();
+        String outputFile = c.getOutputFile(ws).getRemote();
+        String controlDirPath = c.controlDir(ws).getRemote();
+
+        List<String> cmd = new ArrayList<>();
+        cmd.add(binaryPath);
+        cmd.add("-daemon");
+        cmd.add("-executable=cmd");
+        cmd.add("-args=/C call \\\"" + scriptPath + "\\\"");
```
jtnord, line 127: should this not be 2 commands?
```suggestion
        cmd.add("-args=/C");
        cmd.add("call \\\"" + scriptPath + "\\\"");
```

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -59,34 +86,79 @@ public String getScript() {
...
+        return cmd;
+    }
+
+    @Nonnull
+    private List<String> scriptLauncherCmd(BatchController c, FilePath ws) throws IOException, InterruptedException {
+
+        String cmdString;
         if (capturingOutput) {
-            cmd = String.format("@echo off \r\ncmd /c call \"%s\" > \"%s\" 2> \"%s\"\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+            cmdString = String.format("@echo off \r%ncmd /c call \"%s\" > \"%s\" 2> \"%s\"\r%necho %%ERRORLEVEL%% > \"%s\"\r%n",
```
jtnord, line 145: Not convinced here with the String format.
`%n` may be `\r\n` or `\n` depending on if the controller is running on Windows or not.

We really do not want this to be platform dependant new lines - we really do want to force it to be windows as this is batch script and they like windows line ends.  (not sure how this would cope with `\r\r\n` - it may work - it generally does in later version of windows with just `\n` but the output here would be funky and possibly harder to debug than it should be).

```suggestion
            cmdString = String.format("@echo off \r\ncmd /c call \"%s\" > \"%s\" 2> \"%s\"\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
```

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
```diff
@@ -59,34 +86,79 @@ public String getScript() {
...
                 quote(c.getBatchFile2(ws)),
                 quote(c.getOutputFile(ws)),
                 quote(c.getLogFile(ws)),
                 quote(c.getResultFile(ws)));
         } else {
-            cmd = String.format("@echo off \r\ncmd /c call \"%s\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
+            cmdString = String.format("@echo off \r%ncmd /c call \"%s\" > \"%s\" 2>&1\r%necho %%ERRORLEVEL%% > \"%s\"\r%n",
```
jtnord, line 151: ditto line formatting for windows should not be platform dependant

```suggestion
            cmdString = String.format("@echo off \r\ncmd /c call \"%s\" > \"%s\" 2>&1\r\necho %%ERRORLEVEL%% > \"%s\"\r\n",
```

jtnord (approved):
couple of NITs but not so sure about the change in String formatting for the windows Batch script (switching from `\n` to `%n`)

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
car-roll: You know, for windows I'm perfectly fine with having this become the default right now as this directly addresses a Jenkins bug.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
car-roll: Good point. I'll go with `USE_BINARY_WRAPPER` or `USE_SCRIPT_WRAPPER` depending if we make the binary the default

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
car-roll: I'm perfectly fine with keeping it `\n`. This was only changed because spotbugs reported as an error. I would just need to add a `SUPPRESS` I believe.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
car-roll: In this case, the entire string is being supplied as the argument to the `-args` flag. If separated, the parsing would be off and the binary would not get the full string needed for the `-args` flag.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
car-roll: I think I'm okay with making this the default for bash as well too. Previously, the binary was not turned on by default just to make sure the old behavior was still working correctly. I'm fairly comfortable now with making the binary default everywhere and the old behavior as a fall-back.

src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java
car-roll: hmm, interesting, i think it was just my windows machine that was failing with that spotbug error. There's clearly a suppress for this particular issue.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
dwnusbaum: FWIW, it might make sense to leave it off by default and do one release just to make sure nothing got messed up in the general refactoring in this PR and other related PRs that have been merged, advertise on the Jenkins mailing lists that the binary exists and will become the default in the future and activate the feature on some actual Jenkins servers that you use to get some real-world testing (maybe you've already done this though), and then do another release that actually turns it back on by default.

(Just thinking back to everything that happened when we tried making the binary the default behavior on linux)

car-roll: Yeah, that's true. I was thinking the pain had already been done on the linux side. But it would be good to make an announcement and then use a SNAPSHOT like you mentioned.

car-roll: using script wrapper by default. see e8a5b455f4ee91b4fd7692fbb16e427ec91ae3e1
will work on a plan to dogfood binary some more and then switch to default in a separate PR

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -371,7 +450,17 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
...
+                try {
+                    controlDir(workspace).deleteRecursive();
+                    break;
+                } catch (IOException e) {
+                    if (e.getMessage().contains("Unable to delete")) {
+                        Thread.sleep(MS_BETWEEN_RETRIES);
+                        continue;
+                    }
+                }
+            }
```
dwnusbaum, line 463: I'm a little confused about this. IIUC, you added it to avoid needing `Thread.sleep` in the Windows tests, but was the behavior seen in the test something that was happening in real usage too? I'm not sure if it makes sense to add this behavior for all platforms. Also, doesn't `deleteRecursive` already have internal retries?

If it's between this or having long calls to `Thread.sleep` in the Windows tests, I would go with the calls to `sleep`. Maybe another option could be to tweak [the properties that control the retries in `Util.deleteRecursive`](https://github.com/jenkinsci/jenkins/blob/97d919c971fedf4ea5c23172359fec959c15829a/core/src/main/java/hudson/Util.java#L1752-L1779) in the Windows tests only.

car-roll: So just to clarify, I was never seeing this in real usage. The only time I ever see this issue is in this test code. Of course my real world testing is just simple loops.

car-roll: There, was able to get the tuning to work with the testing.  See 9bd66fb8cff734d41b3959d32cda54b49f96b8a6

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
car-roll: sigh...looks like the tuning worked locally, but does not work for CI?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
car-roll: of course the solution only works on my local machine and not CI...

car-roll: Okay, I think I got something working. No amount of `FlagRule` or `System.setProperty` was working. Instead, I gave up on that and decided to do something like how we check for zombies in `BourneShellScriptTest` by using `ps`. But insted of `ps` I use `tasklist` and check for the existence of the `durable-task-monitor` binary.

see 0dcb76cc80957432b23cca459f3fa5f1052d42c7

src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java
```diff
@@ -0,0 +1,169 @@
...
+            }
+            AgentInfo agentInfo = new AgentInfo(os, archType, binaryCompatible, binaryPath, cachingAvailable);
+            agentInfo.setBinaryAvailability(isCached);
+            return agentInfo;
+        }
+
+        @Override
+        public void checkRoles(RoleChecker roleChecker) throws SecurityException {
+
+        }
```
dwnusbaum, line 167: I think this method being empty is a potential security issue (the agent would be able to execute this on the controller, which in this case might not be a big deal but better not to allow it). Instead of having the class implement `FileCallable`, switch to extending `MasterToSlaveFileCallable` and then delete this override.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
```diff
@@ -200,6 +321,10 @@ public FilePath getPowerShellWrapperFile(FilePath ws) throws IOException, Interr
             return controlDir(ws).child("powershellWrapper.ps1");
         }
 
+        public FilePath getTemporaryOutputFile(FilePath ws) throws IOException, InterruptedException {
+            return controlDir(ws).child("temporaryOutput.txt");
```
dwnusbaum, line 325: This gets cleaned up automatically when the control directory is cleaned up, right? Also, the directory name is randomized, and there would never be more than one `temporaryOutput.txt` file per step, right?

dwnusbaum (commented):
Looks fine to me in general, I added one comment about a direct use of `FileCallable` that I think should be change.

src/main/java/org/jenkinsci/plugins/durabletask/PowershellScript.java
car-roll: that is correct, the powershell script pipes the output directly to `temporaryOutput.txt` which is then converted to the final `jenkins-output.txt`. There will only be one `temporaryOutput.txt` file per script and it lives with the other files in the control directory so it gets cleaned up at the same time.

car-roll: Funny aside: a lot of this code existed before the existence of `powershellhelper.ps1`. I ended up re-implementing a lot of that old code unintentionally.

PR #141, opened 2021-07-29
Switch from MD5 to SHA-256 algorithm to hash JENKINS_SERVER_COOKIE value

imonteroperez:
### Description
There is no particular security concern here, but some analyzers flag using MD5 as problematic. So, proposing to switch to SHA-256 and trying to keep backward compatibility on stoping tasks which cookie was hashed with MD5 (`JENKINS_SERVER_COOKIE`)

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -61,6 +61,8 @@ import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Paths;
 import java.nio.file.StandardOpenOption;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.util.Collections;
 import java.util.LinkedList;
 import java.util.List;
@@ -104,7 +106,15 @@ public abstract class FileMonitoringTask extends DurableTask {
     private @CheckForNull String charset;
 
     private static String cookieFor(FilePath workspace) {
-        return "durable-" + Util.getDigestOf(workspace.getRemote());
+        return "durable-" + digest(workspace.getRemote());
+    }
+
+    private static String digest(String text) {
+        try {
+            return Util.toHexString(MessageDigest.getInstance("SHA-256").digest(text.getBytes(StandardCharsets.UTF_8)));
+        } catch (NoSuchAlgorithmException e) {
+            throw new Error(e);
+        }
     }
 
     @Override public final Controller launch(EnvVars env, FilePath workspace, Launcher launcher, TaskListener listener) throws IOException, InterruptedException {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -64,7 +64,6 @@ import java.nio.file.StandardOpenOption;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
 import java.util.Collections;
-import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
@@ -470,10 +469,8 @@ public abstract class FileMonitoringTask extends DurableTask {
         }
 
         @Override public final void stop(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
-            Map<String, String> cookiesMap = new HashMap<>();
-            cookiesMap.put(COOKIE, cookieFor(workspace));
-            cookiesMap.put(COOKIE_OLD, cookieFor(workspace, true)); // To maintain backward compatibility 
-            launcher.kill(Collections.unmodifiableMap(cookiesMap));
+            launcher.kill(Collections.singletonMap(COOKIE, cookieFor(workspace)));
+            launcher.kill(Collections.singletonMap(COOKIE_OLD, cookieFor(workspace, true))); // To maintain backward compatibility
         }
 
         @Override public void cleanup(FilePath workspace) throws IOException, InterruptedException {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -115,8 +115,7 @@ public abstract class FileMonitoringTask extends DurableTask {
      * @return the cookie value
      */
     private static String cookieFor(FilePath workspace, boolean old) {
-        String remote = workspace.getRemote();
-        return String.format("durable-%s", old ? Util.getDigestOf(remote) : digest(remote));
+        return String.format("durable-%s", old ? Util.getDigestOf(workspace.getRemote()) : digest(workspace.getRemote()));
     }
     
     private static String cookieFor(FilePath workspace) {
--- a/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
@@ -94,9 +94,10 @@ public abstract class FileMonitoringTask extends DurableTask {
 
     private static final Logger LOGGER = Logger.getLogger(FileMonitoringTask.class.getName());
 
-    private static final String COOKIE = "JENKINS_SERVER_COOKIE_2";
+    private static final String COOKIE = "JENKINS_SERVER_COOKIE_ID";
     
-    protected static final String COOKIE_OLD = "JENKINS_SERVER_COOKIE";
+    /** To maintain backward compatibility */
+    private static final String COOKIE_OLD = "JENKINS_SERVER_COOKIE";
     
     protected static final String BINARY_RESOURCE_PREFIX = "/io/jenkins/plugins/lib-durable-task/durable_task_monitor_";
 
@@ -108,10 +109,18 @@ public abstract class FileMonitoringTask extends DurableTask {
      */
     private @CheckForNull String charset;
 
+    /**
+     * Provides the cookie value for a given {@link FilePath} workspace. It also uses
+     * a flag to identify if we want this cookie hash based on MD5 (former/old format) or SHA-256 algorithm.
+     * This method is only used to maintain backward compatibility on stopping tasks that were
+     * launched before the new format was applied.
+     * @param workspace
+     * @param old
+     * @return
+     */
     private static String cookieFor(FilePath workspace, boolean old) {
-        String suffix = "durable-%s";
         String remote = workspace.getRemote();
-        return String.format(suffix, old ? Util.getDigestOf(remote) : digest(remote));
+        return String.format("durable-%s", old ? Util.getDigestOf(remote) : digest(remote));
     }
     
     private static String cookieFor(FilePath workspace) {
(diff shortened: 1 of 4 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -443,6 +470,7 @@ public Integer invoke(File f, VirtualChannel channel) throws IOException, Interr
 
         @Override public final void stop(FilePath workspace, Launcher launcher) throws IOException, InterruptedException {
             launcher.kill(Collections.singletonMap(COOKIE, cookieFor(workspace)));
+            launcher.kill(Collections.singletonMap(COOKIE_OLD, cookieFor(workspace, true))); // To maintain backward compatibility
```
raul-arabaolaza, line 473: Can calling kill in an already killed instance  have some undesirable side effects?

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -103,8 +108,30 @@
      */
     private @CheckForNull String charset;
 
+    /**
+     * Provides the cookie value for a given {@link FilePath} workspace. It also uses
+     * a flag to identify if we want this cookie hash based on MD5 (former/old format) or SHA-256 algorithm.
+     * This method is only used to maintain backward compatibility on stopping tasks that were
+     * launched before the new format was applied.
+     * @param workspace
+     * @param old
```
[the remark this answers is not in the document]
imonteroperez: Updated on https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/3c02abe8422c18e32d94e4a3d7c01047699d24ed

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
imonteroperez: AFAIK no side effects, but please would you mind elaborating what are you thinking?

raul-arabaolaza: For example, let's asume this code is launched in a job that is about to be migrated but running. Will the call to kill using the new cookie raise any exception? Will it print things that may confuse the user in the log? After that if the first call to `kill` works, may the second call in an already killed job fail somehow? Would it print any stack trace or confusing information   to the log also? That kind of things

jglick (changes requested):
Rather save `cookieValue` as a field in `FileMonitoringController`. `stop` should use that field; for compatibility with builds running across the restart, `readResolve` can set it to the old MD5 value.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: see https://github.com/jenkinsci/durable-task-plugin/pull/141#pullrequestreview-720404766

imonteroperez:
> Rather save `cookieValue` as a field in `FileMonitoringController`. `stop` should use that field; for compatibility with builds running across the restart, `readResolve` can set it to the old MD5 value.

> Rather save `cookieValue` as a field in `FileMonitoringController`. `stop` should use that field; for compatibility with builds running across the restart, `readResolve` can set it to the old MD5 value.

I guess in this case it is not really required to implement the `readResolve` method right? I mean, `cookieValue` is only going to be used on the `stop` method, so let's just evaluate it there and in case it is `null` setup it to the old MD5 value. If `cookieValue` is not null, it implies it was initialized with the SHA-256 value.

See: https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/ae6d5ea3a8af041a4aa5d48711d7f600ba35a092

jglick:
Yes, you can interpret `null` values in lieu of implementing `readResolve`, the difference would just be stylistic.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -103,8 +105,30 @@
...
     private @CheckForNull String charset;
 
+    /**
+     * Provides the cookie value for a given {@link FilePath} workspace. It also uses
+     * a flag to identify if we want this cookie hash based on MD5 (former/old format) or SHA-256 algorithm.
+     * This method is only used to maintain backward compatibility on stopping tasks that were
+     * launched before the new format was applied.
+     * @param workspace - workspace path used to setup the digest
+     * @param old - boolean to select if we want to use former/old hash algorithm to maintain backward compatibility
+     * @return
```
jglick, line 115:
```suggestion
```

or actually provide content

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -103,8 +105,30 @@
...
+     */
+    private static String cookieFor(FilePath workspace, boolean old) {
+        String remote = workspace.getRemote();
+        return String.format("durable-%s", old ? Util.getDigestOf(remote) : digest(remote));
+    }
+    
     private static String cookieFor(FilePath workspace) {
```
jglick, line 122: just inline

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -241,6 +265,9 @@ protected static FilePath requestBinary(FilePath nodeRoot, AgentInfo agentInfo,
          * Only used if {@link #writeLog(FilePath, OutputStream)} is used; not used for {@link #watch}.
          */
         private long lastLocation;
+        
+        /** Store the cookie value for JENKINS_SERVER_COOKIE */
```
jglick, line 269:
```suggestion
        /** Store the value for {@link #COOKIE} */
```

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -275,6 +302,7 @@ protected FileMonitoringController(FilePath ws) throws IOException, InterruptedE
             } else {
                 controlDir = null;
             }
+            cookieValue = cookieFor(ws);
```
jglick, line 305: This should rather be a constructor parameter, defaulting like this (to the _old_ format) only in the case that the implementation of `launchWithCookie` or `doLaunch` neglected to use the new constructor overload. (And of course deprecate the original overload and fix the callers in this repo.)

imonteroperez: IIUC you are proposing:
- to deprecate the current `FileMonitoringController` constructor and there set up `cookieValue` to the old format.
- to create a new one with `cookieValue` as constructor parameter (value that could be `null` or being provided by the callers - expected the new format)
- to modify all the callers to use the new one

It sounds to me like introducing some extra complexity, and, in addition, making it feasible in code to still have the old format on launching new tasks. Idea was to replace/switch and maintain backward compatibility just for previously launched tasks on nodes with restart capabilities.

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
imonteroperez: Updated to provide some content on https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/930b0ab95b2736951a7baadbd8fb5961dc177176

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
imonteroperez: Updated on https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/930b0ab95b2736951a7baadbd8fb5961dc177176

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
jglick: > value that could be `null`

No, non-null.

> introducing some extra complexity

Not much I think, just threading a variable through another call, rather than trying to recompute it and hoping you come up with the same result.

> still have the old format on launching new tasks

No, because the impls of `DurableTask` in this plugin (which is probably all of them in the world) are updated to call the new constructor.

imonteroperez: Got it, refactored in https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/2fb15cccaba6b6fcd4594628b065eff267b7b43d

imonteroperez:
> Still recommend [#141 (comment)](https://github.com/jenkinsci/durable-task-plugin/pull/141#discussion_r681979407)

Applied https://github.com/jenkinsci/durable-task-plugin/pull/141#discussion_r681979407 on https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/2fb15cccaba6b6fcd4594628b065eff267b7b43d

jglick (commented):
You just changed where the value was redundantly computed. It is computed first in https://github.com/jenkinsci/durable-task-plugin/blob/787d3ac9a42077476881f6233c2fee6920eaccfa/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java#L136

jglick:
So for example https://github.com/jenkinsci/durable-task-plugin/blob/787d3ac9a42077476881f6233c2fee6920eaccfa/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java#L119-L138 should be passing along the existing cookie value.

imonteroperez:
> You just changed where the value was redundantly computed. It is computed first in
> 
> https://github.com/jenkinsci/durable-task-plugin/blob/787d3ac9a42077476881f6233c2fee6920eaccfa/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java#L136

Perfect, I updated the code to avoid redundant invocations to `cookieFor` method. Updated in https://github.com/jenkinsci/durable-task-plugin/pull/141/commits/21dc2b4e6afae01025e5de2b6dcd327b914bcbb5. Thanks!

raul-arabaolaza (approved):
For administrative reasons mostly, I am far from an expert

car-roll (approved):
LGTM, I'm playing a bit of catch up, but I think all of @jglick 's comments were addressed?

olamy (approved):

alecharp:
@dwnusbaum could you have a look at this PR? Thanks.

dwnusbaum (approved):

car-roll:
@alecharp I was just waiting on a confirmation by @jglick that all his requested changes were satisfied before merging. But yes, since @olamy already approved and i'm _pretty_ sure all comments have been addressed (minus the new ones), then we can merge. Thanks for the extra review @dwnusbaum !

batmat:
Sorry Devin, we should have clarified upfront: given Jesse is out on vacation, we were solely hoping for your review (which you did, thanks a lot!) kind of as an alternative to Jesse's one to move forward before he comes back given we think we addressed his feedback.

jglick:
I am not able to do any more reviews this week so go ahead. Nothing serious anyway.

imonteroperez:
> @imonteroperez can you review unresolved comment and resolve all of them when deemed addressed? Once done, we should be able to merge & release I believe. Thanks!

@batmat All comments addressed/solved

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -264,7 +292,18 @@ void registerForCleanup(Closeable c) {
...
 
+        protected FileMonitoringController(FilePath ws, @NonNull String cookieValue) throws IOException, InterruptedException {
+            setupControlDir(ws);
+            this.cookieValue = cookieValue;
+        }
+        
+        @Deprecated
         protected FileMonitoringController(FilePath ws) throws IOException, InterruptedException {
+            setupControlDir(ws);
+            this.cookieValue = cookieFor(ws, true);
```
jglick, line 303: or just
```java
this(ws, cookieFor(ws, true));
```

jglick (commented):
Looks right from a brief glance, thanks!

PR #149, opened 2022-05-03
s)

CatherineKiiru:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,7 +57,7 @@ Release date: 2019-10-22
 > **NOTE**: To revert to previous release behavior, pass the system property
 > `org.jenkinsci.plugins.durabletask.BourneShellScript.FORCE_SHELL_WRAPPER=true` to the Java command line used to start Jenkins.
 
--   Update ssh-slaves ([PR \#100](https://github.com/jenkinsci/durable-task-plugin/pull/100))
+-   Update ssh-agents ([PR \#100](https://github.com/jenkinsci/durable-task-plugin/pull/100))
 -   Do not fail tests when run on a machine without Docker installed.
     ([PR \#101](https://github.com/jenkinsci/durable-task-plugin/pull/101))
 -   Improve watcher logging
@@ -89,7 +89,7 @@ Release date: 2019-07-05
     be used to enable zombie-reaping in docker containers (--init) or kubernetes pods (shared
     process namespaces).
 -   Bugfix: Use `sh` to run shell scripts rather than attempting to
-    use the absolute path to the default shell from the master on
+    use the absolute path to the default shell from the Controller on
     agents. ([PR \#95](https://github.com/jenkinsci/durable-task-plugin/pull/95))
 -   Bugfix: Make PowerShell exit codes propagate correctly. Fixes a
     regression from version 1.23
@@ -129,7 +129,7 @@ Release date: 2018-09-25
     processes from 15 seconds to 5 minutes
     ([JENKINS-48300](https://issues.jenkins-ci.org/browse/JENKINS-48300))
 -   Developer: Define API for pushing durable task logs from build
-    agents directly instead of having the Jenkins master pull logs from
+    agents directly instead of having the Jenkins controller pull logs from
     build agents
     ([JENKINS-52165](https://issues.jenkins-ci.org/browse/JENKINS-52165))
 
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
```diff
@@ -57,7 +57,7 @@ Release date: 2019-10-22
 > **NOTE**: To revert to previous release behavior, pass the system property
 > `org.jenkinsci.plugins.durabletask.BourneShellScript.FORCE_SHELL_WRAPPER=true` to the Java command line used to start Jenkins.
 
--   Update ssh-slaves ([PR \#100](https://github.com/jenkinsci/durable-task-plugin/pull/100))
+-   Update ssh-agents ([PR \#100](https://github.com/jenkinsci/durable-task-plugin/pull/100))
```
MarkEWaite, line 60: Slightly different for linguistic reasons.  The previous entry used the "plugin ID" which is not well known to users and is unchanging even for inclusive naming.  Let's use the plugin name rather than the identifier
```suggestion
-   Update SSH Build Agents plugin ([PR \#100](https://github.com/jenkinsci/durable-task-plugin/pull/100))
```

CHANGELOG.md
```diff
@@ -89,7 +89,7 @@ Release date: 2019-07-05
     be used to enable zombie-reaping in docker containers (--init) or kubernetes pods (shared
     process namespaces).
 -   Bugfix: Use `sh` to run shell scripts rather than attempting to
-    use the absolute path to the default shell from the master on
+    use the absolute path to the default shell from the Controller on
```
MarkEWaite, line 92: Lower case rather than upper case inside the sentence
```suggestion
    use the absolute path to the default shell from the controller on
```

CHANGELOG.md
```diff
@@ -338,7 +338,7 @@ Release date: 2015-12-03
 Release date: 2015-04-08
 
 -   Do not kill a one-shot agent merely because a flyweight task
-    happened to run on it (rather than on master as usual). Works around
+    happened to run on it (rather than on the Controller as usual). Works around
```
MarkEWaite, line 341: Lower case
```suggestion
    happened to run on it (rather than on the controller as usual). Works around
```

pom.xml
```diff
@@ -58,7 +58,7 @@
         </dependency>
         <dependency>
             <groupId>org.jenkins-ci.plugins</groupId>
-            <artifactId>ssh-slaves</artifactId>
+            <artifactId>ssh-agents</artifactId>
```
MarkEWaite, line 61: That's an identifier, needs to remain the same
```suggestion
            <artifactId>ssh-slaves</artifactId>
```

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -268,7 +268,7 @@ private List<String> scriptLauncherCmd(ShellController c, FilePath ws, @CheckFor
 
     /*package*/ static final class ShellController extends FileMonitoringController {
 
-        /** Last time we checked the timestamp, in nanoseconds on the master. */
+        /** Last time we checked the timestamp, in nanoseconds on the Controller. */
```
MarkEWaite, line 271: Lower case
```suggestion
        /** Last time we checked the timestamp, in nanoseconds on the controller. */
```

src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java
```diff
@@ -640,7 +640,7 @@ private static class Watcher implements Runnable {
                     Timer.get().schedule(this, 100, TimeUnit.MILLISECONDS);
                 }
             } catch (Exception x) {
-                // note that LOGGER here is going to the agent log, not master log
+                // note that LOGGER here is going to the agent log, not the Controller log
```
MarkEWaite, line 643: Lower case
```suggestion
                // note that LOGGER here is going to the agent log, not the controller log
```

MarkEWaite (changes requested):
Minor improvements

src/main/java/org/jenkinsci/plugins/durabletask/Handler.java
```diff
@@ -34,7 +34,7 @@
 
 /**
  * A remote handler which may be sent to an agent and handle process output and results.
- * If it needs to communicate with the master, you may use {@link VirtualChannel#export}.
+ * If it needs to communicate with the Controller, you may use {@link VirtualChannel#export}.
```
car-roll, line 37:
```suggestion
 * If it needs to communicate with the controller, you may use {@link VirtualChannel#export}.
```
I assume the same here

PR #152, opened 2022-08-08
[JENKINS-69277] relaunch an agent if the connection is broken

jtnord:
Even though this is a "Once" Retention Strategy we expect a durable task to
be able to run to completion even if the connection is broken.  So we
relaunch the agent if the agent is offline and it is not done.

We are not forcing a reconnect, as this allows any existing connect
operation (which could be the first one, to continue running).

https://issues.jenkins.io/browse/JENKINS-69277

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -76,7 +76,7 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
         // so lets not presume any asynchronous nature and instead that this could be a long blocking op.
         boolean shouldRestart = false;
         synchronized (this) {
-            if (!terminating && c.isOffline() && !disabled) {
+            if (!terminating && c.isOffline() && c.isLaunchSupported() && !disabled) {
                 shouldRestart = true;
             }
         }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -71,7 +71,11 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
                 done(c);
             }
         }
-
+        // if the agent is not done and it is offline we should trigger a relaunch
+        if (!terminating && c.isOffline()) {
+            LOGGER.log(Level.FINE, "Attempting relaunch of {0}", c.getName());
+            c.connect(false);
+        }
         // Return one because we want to check every minute if idle.
         return 1;
     }
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -48,7 +48,7 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
     private static final Logger LOGGER = Logger.getLogger(OnceRetentionStrategy.class.getName());
 
     private transient boolean terminating;
-    
+
     private int idleMinutes;
 
     /**
@@ -72,7 +72,15 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
             }
         }
         // if the agent is not done and it is offline we should trigger a relaunch
-        if (!terminating && c.isOffline()) {
+        // do not do this whilst we hold the lock, as connect is documented only as "This method may return immediately"
+        // so lets not presume any asynchronous nature and instead that this could be a long blocking op.
+        boolean shouldRestart = false;
+        synchronized (this) {
+            if (!terminating && c.isOffline() && !disabled) {
+                shouldRestart = true;
+            }
+        }
+        if (shouldRestart) {
             LOGGER.log(Level.FINE, "Attempting relaunch of {0}", c.getName());
             c.connect(false);
         }
--- a/src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
@@ -1,0 +1,227 @@
+package org.jenkinsci.plugins.durabletask.executors;
+
+import java.io.File;
+import java.io.IOException;
+import java.time.Duration;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+import org.jenkinsci.plugins.durabletask.BourneShellScript;
+import org.jenkinsci.plugins.durabletask.WindowsBatchScript;
+import org.jenkinsci.plugins.workflow.cps.CpsFlowDefinition;
+import org.jenkinsci.plugins.workflow.job.WorkflowJob;
+import org.jenkinsci.plugins.workflow.job.WorkflowRun;
+import org.junit.BeforeClass;
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.rules.TemporaryFolder;
+import org.jvnet.hudson.test.BuildWatcher;
+import org.jvnet.hudson.test.Issue;
+import org.jvnet.hudson.test.JenkinsRule;
+import org.jvnet.hudson.test.LoggerRule;
+import org.jvnet.hudson.test.SimpleCommandLauncher;
+import hudson.Functions;
+import hudson.model.Computer;
+import hudson.model.LoadStatistics;
+import hudson.model.TaskListener;
+import hudson.model.Descriptor.FormException;
+import hudson.model.labels.LabelAtom;
+import hudson.remoting.RequestAbortedException;
+import hudson.remoting.Which;
+import hudson.slaves.AbstractCloudComputer;
+import hudson.slaves.AbstractCloudSlave;
+import hudson.slaves.Cloud;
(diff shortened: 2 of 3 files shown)


Vlatombe (commented):
Looks reasonable.

jtnord:
@timja  FYI I think Azure ACI agents may have a similar bug due[ to no reconnection kickoff](https://github.com/jenkinsci/azure-container-agents-plugin/blob/920e4546d0be9eb8a14775847b8bde01bbc74365/src/main/java/com/microsoft/jenkins/containeragents/strategy/ContainerOnceRetentionStrategy.java#L46-L69)

jglick (commented):
I guess you tested this end-to-end in the GCE plugin? `KubernetesSlave` uses this class but in that case the change would be at best useless, and possibly harmful—if the agent JVM is dead it means the `jnlp` container exits and the whole pod is destroyed. Unclear if we would even reach this line in the K8s case.

jtnord:
> I guess you tested this end-to-end in the GCE plugin? `KubernetesSlave` uses this class but in that case the change would be at best useless, and possibly harmful—if the agent JVM is dead it means the `jnlp` container exits and the whole pod is destroyed. Unclear if we would even reach this line in the K8s case.

deployed on the Gaia controller - reconnection occured.   `connect` is documented as a `no-op` on inbound agents, so I would not expect any regression in k8s agents.

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -71,7 +71,19 @@ public long check(final AbstractCloudComputer c) {
                 done(c);
             }
         }
-
+        // if the agent is not done and it is offline we should trigger a relaunch
+        // do not do this whilst we hold the lock, as connect is documented only as "This method may return immediately"
+        // so lets not presume any asynchronous nature and instead that this could be a long blocking op.
+        boolean shouldRestart = false;
+        synchronized (this) {
```
jtnord, line 78: whilst this locking is not perfect (we may try to relaunch and agent just
as it was done), we have set the computer to not accept any tasks so the
agent should become idle, and be terminated (either in the current
termination block) or the next time the check runs.

jtnord:
BTW - I am open to ideas on how to unit test this.  Depending on enough things for workflow (in test scope) would likely end up causing a circular reference at some point and mock-cloud is too much mock that it is unrealistic

jglick:
> `connect` is documented as a `no-op` on inbound agents

Ah OK, so this should only affect clouds with outbound launchers.

> Depending on enough things for workflow (in test scope) would likely end up causing a circular reference

Should not be a problem. All sorts of plugins have circular references if you consider test scope.

> mock-cloud is too much mock that it is unrealistic

`mock-slaves` plugin you mean? I suspect you could actually reproducible well enough with that, or something adapted from it. Maybe something like https://github.com/jenkinsci/jenkins-test-harness/blob/9741f4968f856e6019674f7816fb96afe3b4fb31/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java#L109 depending on the scenario you observe.

jtnord:
> > `connect` is documented as a `no-op` on inbound agents
> 
> Ah OK, so this should only affect clouds with outbound launchers.

technically it is "pro-active" launching ([ref](https://javadoc.jenkins.io/hudson/model/Computer.html#connect(boolean)))

which eventually calls the `Launcher.launch`.  Looking at the k8s [implementation ](https://github.com/jenkinsci/kubernetes-plugin/blob/282da634b8f0515440a08998d431e2a270c509dd/src/main/java/org/csanchez/jenkins/plugins/kubernetes/KubernetesLauncher.java#L97)  this should be guarded by `launched` for the success case and for the failure the `node` should be terminated.

the method being synchronized will protect against multiple successful launches.

looking some more..

> > Depending on enough things for workflow (in test scope) would likely end up causing a circular reference
> 
> Should not be a problem. All sorts of plugins have circular references if you consider test scope.

👍 

> > mock-cloud is too much mock that it is unrealistic
> 
> `mock-slaves` plugin you mean? 

yup -that has a [MockCloud](https://github.com/jenkinsci/mock-slave-plugin/blob/8cf18e8ca67a48ea879eb9086bd4e46365aeea2d/src/main/java/org/jenkinci/plugins/mock_slave/MockCloud.java)

I suspect you could actually reproducible well enough with that, or something adapted from it. Maybe something like https://github.com/jenkinsci/jenkins-test-harness/blob/9741f4968f856e6019674f7816fb96afe3b4fb31/src/main/java/org/jvnet/hudson/test/InboundAgentRule.java#L109 depending on the scenario you observe.

👍

src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
```diff
@@ -71,7 +71,11 @@ public long check(final AbstractCloudComputer c) {
                 done(c);
             }
         }
-
+        // if the agent is not done and it is offline we should trigger a relaunch
+        if (!terminating && c.isOffline()) {
```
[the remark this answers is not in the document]
jtnord: fixed in https://github.com/jenkinsci/durable-task-plugin/pull/152/commits/c7520b58a2a534bb6bfc6032be1b5438fc8152d5

jtnord:
test added - can be seen standalone in #153

src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
```diff
@@ -0,0 +1,227 @@
...
+    public TemporaryFolder temp = new TemporaryFolder();
+
+    @Rule
+    public BuildWatcher bw = new BuildWatcher();
+
+    @BeforeClass
+    public static void setupStatics() {
+        LoadStatistics.CLOCK = 100; // 100ms for the LoadStatistics so we have quick provisioning
+        // deadConnectionsShouldReLaunched seems to fail always when run as part of the suite - but passes always when run individually without the binary wrapper
+        // (because the ping command and its parent cmd.exe get killed :-o )
```
jtnord, line 68: this smells like a bug in the wrapper - however this is not something introduced here.
could be easy to reproduce by changing this to false for someone to investigate - but I beleive we wanted to change the default of the binary wrappers to be enabled anyway - so this seems like it could end up in being wasted effort.

src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
```diff
@@ -0,0 +1,227 @@
...
+        WindowsBatchScript.USE_BINARY_WRAPPER=true; 
+        BourneShellScript.USE_BINARY_WRAPPER=true;
```
jtnord, line 70: no idea if the same issue happens on Linux as windows where sometimes the command get terminated or not.

src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
```diff
@@ -0,0 +1,227 @@
...
+
+        @Override
+        public Collection<PlannedNode> provision(CloudState state, int excessWorkload) {
+            LOGGER.log(Level.FINE, "SimpleCloud.provision(({0}, {1}), {2}", new Object[] {state.getLabel(), state.getAdditionalPlannedCapacity(), excessWorkload});
+            Collection<PlannedNode> retVal = new HashSet<PlannedNode>();
+            for (int i =0; i < excessWorkload - state.getAdditionalPlannedCapacity(); ++i) {
+                String agentName;
+                synchronized (this) {
+                    if (count != 0) {
+                        // sometimes we end up with 2 agents due to the lovely cloud API.
```
jtnord, line 173: not entirely sure what causes this - but this is good enough for this test.

src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
```diff
@@ -0,0 +1,227 @@
...
+        assertThat(jr.jenkins.getComputers(), arrayWithSize(1)); // the Jenkins computer itself
+    }
+
+    private void killAgent() throws IOException, InterruptedException {
+        Computer[] computers = Jenkins.get().getComputers();
+        for (Computer c : computers) {
+            if (c instanceof MySimpleCloudComputer) {
+                LOGGER.log(Level.FINE, "Asking {0} to commit suicide", c);
+                try {
+                    RemotingDiagnostics.executeGroovy("Runtime.getRuntime().halt(1)", c.getChannel());
```
jtnord, line 138: simpler than attempts to go via the OS to get the right process to kill it.

src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
car-roll: Yes, this has been a long-standing windows issue (see https://issues.jenkins.io/browse/JENKINS-47657) that was hopefully solved by the binary wrapper for windows. Using the script wrapper will definitely not work as the process will always be a child of the launching agent.
FWIW I have been running the majority of tests in powershellscript, powershellcore, etc, with the binary wrapper set as true.

src/test/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategyTest.java
car-roll: The linux version doesn't really have this problem (or if it does, it doesn't consistently rear its head like windows). However, I do prefer to test with the binary wrapper as we would eventually like to make this the default.

car-roll (approved):

car-roll:
CI hanging for some reason. Had an issue like this with another PR

PR #179, opened 2023-05-21
Refresh plugin

basil:


basil:
Used my admin powers to replay the build with the `Jenkinsfile` changes here: https://ci.jenkins.io/job/Plugins/job/durable-task-plugin/job/PR-179/2/

basil:
CC @car-roll

src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java
```diff
@@ -162,7 +162,7 @@ private Slave prepareAgentDocker() throws Exception {
             case NO_INIT:
                 return new DumbSlave("docker",
                             "/home/jenkins/agent",
-                            new SimpleCommandLauncher("docker run -i --rm jenkins/slave:3.29-2 java -jar /usr/share/jenkins/slave.jar"));
+                            new SimpleCommandLauncher("docker run -i --rm jenkins/agent:latest-jdk11 java -jar /usr/share/jenkins/agent.jar"));
```
car-roll, line 165: @basil thanks for the PR! Question, is it possible that the jenkins version of the `agent` image could get out of sync with the plugin so that it could not run? Or is that not how it works?

basil: Remoting is generally both forward and backward compatible, so this should not be a problem.

PR #201, opened 2024-01-04
[JENKINS-72488] allow apk to upgrade openssh in Alpine Dockerfile

lne3:
Allow apk to upgrade openssh in Alpine Dockerfile to fix [JENKINS-72488](https://issues.jenkins.io/browse/JENKINS-72488)

At https://gitlab.alpinelinux.org/alpine/aports/-/issues/15606 user alice describes the issue and possible solutions:

> that's because that image already has openssh-client installed, but openssh was updated (so 9.5 doesn't exist anymore), and installing openssh (9.6), then adds a dependency on openssh-client (9.6) (because openssh depends on it), but an apk add won't upgrade, so it fails (since it doesn't want to upgrade to openssh-client 9.6). apk add -u openssh would work, or apk upgrade && apk add openssh to allow upgrades as part of the addition

### Testing done

Succesfully, ran docker build with the modified Dockerfile. This PR's github checks will show whether it works in CI environment.

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/test/resources/org/jenkinsci/plugins/durabletask/AlpineFixture/Dockerfile
+++ b/src/test/resources/org/jenkinsci/plugins/durabletask/AlpineFixture/Dockerfile
@@ -1,6 +1,6 @@
 FROM jenkins/agent:alpine-jdk11
 USER root
-RUN apk add --update --no-cache openssh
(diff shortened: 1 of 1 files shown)


jglick (approved):
Thanks!

PR #204, opened 2024-02-02
[JENKINS-70874] Fix 'Text file busy' issue

ckullabosch:
https://issues.jenkins.io/browse/JENKINS-70874

### Testing done

```[tasklist]
### Submitter checklist
```

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -141,7 +141,8 @@ public final class BourneShellScript extends FileMonitoringTask {
         ShellController c = new ShellController(ws,(os == OsType.ZOS), cookieValue, jenkinsResultTxtEncoding);
         FilePath shf = c.getScriptFile(ws);
 
-        // here we need to keep the file open for a longer time to make the race to happen more likely
+        // JENKINS-70874: if a new process is forked during this call, the writeable file handle will be copied and leading to the "Text file busy" issue
+        // when executing the script.
         shf.write(script, scriptEncodingCharset);
 
         String shell = null;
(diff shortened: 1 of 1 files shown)


ckullabosch:
@jglick My understanding is that this not a JDK level problem. It's an inherent problem, when forking and writing executable files concurrently (independent of the language used). The issue in https://bugs.openjdk.org/browse/JDK-8068370 is probably the same root cause (but not fixable inside the JVM). You are right that lib-durable-task needs a fix as well. Please see here for more background information: https://issues.jenkins.io/browse/JENKINS-48258?focusedId=336050&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-336050

src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -229,27 +231,32 @@ private List<String> scriptLauncherCmd(ShellController c, FilePath ws, @CheckFor
         if (os == OsType.WINDOWS) { // JENKINS-40255
             scriptPath = scriptPath.replace("\\", "/"); // cygwin sh understands mixed path  (ie : "c:/jenkins/workspace/script.sh" )
         }
+        String scriptPathCopy = scriptPath + ".copy"; // copy file to protect against "Text file busy", see JENKINS-70874
```
jglick, line 234: Ought to be covered by https://github.com/jenkinsci/durable-task-plugin/blob/8ea90b6f96daddccc18e4622eb289675247e213c/src/main/java/org/jenkinsci/plugins/durabletask/FileMonitoringTask.java#L487 I think.

jglick (approved):
Patch itself seems reasonable. Will leave open for a bit to see if anyone else cares to comment on the approach.

jglick:
(Can I assume that the problem was reproducible locally without this patch, and corrected when running with this patch? If you are not familiar with Jenkins plugin development generally, the easiest way is to use `mvn hpi:run`.)

ckullabosch:
Yes, I did run tests for multiple hours and have not seen the issue again after applying the fix. Nevertheless it's happening very rare and is hard to reproduce.

car-roll (approved):

ckullabosch:
The git-client plugin needs also a fix for the "Text file busy" issue, please see https://github.com/jenkinsci/git-client-plugin/pull/1103 and https://issues.jenkins.io/browse/JENKINS-72667.

PR #221, opened 2024-08-05
Switch from CentOS to UBI

jglick:
Tests started failing recently in this plugin, e.g. https://github.com/jenkinsci/durable-task-plugin/pull/220/checks?check_run_id=28221800550 (#220). This seems to be due to a CentOS EOL https://github.com/docker-library/official-images/pull/17094. For purposes of these tests, we just care about general distro flavor, so using UBI is close enough.

Tested locally with:

```bash
mvnd test -Dtest='BourneShellScriptTest#*[3: CENTOS]'
```


MarkEWaite (approved):
Thanks!

raul-arabaolaza (approved):

basil (approved):

car-roll (approved):

PR #224, opened 2024-08-06
No need for an explicit dep on Hamcrest

jglick:
https://github.com/jenkinsci/durable-task-plugin/pull/222#pullrequestreview-2219707164 https://github.com/jenkinsci/durable-task-plugin/pull/223 Normally included in all plugins via `jenkins-test-harness`, and in this case also via `awaitility`.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -77,11 +77,6 @@
             <scope>test</scope>
         </dependency>
         <dependency>
-             <groupId>org.hamcrest</groupId>
-             <artifactId>hamcrest</artifactId>
-             <scope>test</scope>
-        </dependency>
-        <dependency>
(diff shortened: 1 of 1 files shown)


jtnord (commented):
But as it is used it directly it should have an explicit dep and not rely on any transitive dependency.

as such this seems incorrect to me.

jglick:
> as it is used it directly it should have an explicit dep

Perhaps, but the same could be said of 90% of Jenkins plugins, and this is the only one I am familiar with which declares the dep explicitly. (As of your #152, apparently.) Unless you propose to activate the Enforcer rule which would mandate use of explicit dependencies, this seems like a gratuitous inconsistency.

Artmorse (approved):

PR #232, opened 2024-09-13
Handle unsigned int errorcodes more gracefully in Windows Batch

StefanSpieker:
We have a job on a Windows machine which executes a Matlab project within a batch-file, and has actually a quite high error code of `3221225477`. We are using the binary wrapper, and so this quite high number gets written into the text file. The parsing fails with an `IOException` and the connection gets lost to the agent. I was first confused, but actually windows uses an unsigned int for the error codes, as documented [here](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-seterrormode). I decided to set it to max integer, which is from my point of view better than throwing the exception, but I added a warning log message about it.

### Testing done

Added a unit test which reproduces the error and verifies that it is resolved like this.

### Submitter checklist

### DIFF
--- a/src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
+++ b/src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
@@ -114,6 +114,22 @@ public class WindowsBatchScriptTest {
         c.cleanup(ws);
     }
 
+    @Test public void exitCommandUnsignedInt() throws Exception {
+        Controller c = new WindowsBatchScript("echo hello world\r\nexit 3221225477").launch(new EnvVars(), ws, launcher, listener);
+        awaitCompletion(c);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        c.writeLog(ws, baos);
+        if (enableBinary) {
+            assertEquals(Integer.valueOf(Integer.MAX_VALUE), c.exitStatus(ws, launcher, listener));
+        } else {
+            assertEquals(Integer.valueOf(-1073741819), c.exitStatus(ws, launcher, listener));
+        }
+
+        String log = baos.toString();
+        assertTrue(log, log.contains("hello world"));
+        c.cleanup(ws);
+    }
+
     @Test public void exitBCommandAfterError() throws Exception {
         Controller c = new WindowsBatchScript("cmd /c exit 42\r\nexit /b").launch(new EnvVars(), ws, launcher, listener);
         awaitCompletion(c);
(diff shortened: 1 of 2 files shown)


src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
```diff
@@ -114,6 +114,22 @@ private void testWithPath(String path) throws Exception {
...
+    @Test public void exitCommandUnsignedInt() throws Exception {
+        Controller c = new WindowsBatchScript("echo hello world\r\nexit 3221225477").launch(new EnvVars(), ws, launcher, listener);
+        awaitCompletion(c);
+        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+        c.writeLog(ws, baos);
+        if (enableBinary) {
+            assertEquals(Integer.valueOf(Integer.MAX_VALUE), c.exitStatus(ws, launcher, listener));
+        } else {
+            assertEquals(Integer.valueOf(-1073741819), c.exitStatus(ws, launcher, listener));
+        }
```
jglick, line 126: Is this the difference between https://github.com/jenkinsci/lib-durable-task/blob/c823f6cb78f5208589b41e5ba467635b5ea121c1/src/pkg/common/common.go#L151-L157 vs. https://github.com/jenkinsci/durable-task-plugin/blob/cc2bd5774993fb6d30af1c28a91178ea463cb646/src/main/java/org/jenkinsci/plugins/durabletask/WindowsBatchScript.java#L148 ? Not great there is this inconsistency, though I suppose it is unlikely to matter much in practice (either way it is a failure).

jglick (approved):
FTR https://github.com/jenkinsci/durable-task-plugin/blob/cc2bd5774993fb6d30af1c28a91178ea463cb646/src/main/java/org/jenkinsci/plugins/durabletask/Controller.java#L78-L81 would ideally be expanded to `Long` but that would require changes also to https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/b7cec8f3b95e9807f78706346641fd9e41ff4039/src/main/java/org/jenkinsci/plugins/workflow/steps/durable_task/DurableTaskStep.java#L655 etc., which feels like overkill.

src/test/java/org/jenkinsci/plugins/durabletask/WindowsBatchScriptTest.java
car-roll: Yikes, so windows is always using unsigned int to return error codes? If so, I think lib-durable needs to be updated

PR #243, opened 2024-11-22
[JENKINS-33708] Do not call `nohup` for `sh` steps running on Windows

sunnyqeen:
see https://issues.jenkins.io/browse/JENKINS-33708 discussion.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -263,7 +263,7 @@ public final class BourneShellScript extends FileMonitoringTask {
 
         cmdString = cmdString.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
         List<String> cmd = new ArrayList<>();
-        if (os != OsType.DARWIN) { // JENKINS-25848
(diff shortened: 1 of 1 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -263,7 +263,7 @@ private List<String> scriptLauncherCmd(ShellController c, FilePath ws, @CheckFor
 
         cmdString = cmdString.replace("$", "$$"); // escape against EnvVars jobEnv in LocalLauncher.launch
         List<String> cmd = new ArrayList<>();
-        if (os != OsType.DARWIN) { // JENKINS-25848
+        if (os != OsType.DARWIN && os != OsType.WINDOWS) { // JENKINS-25848  JENKINS-33708
```
jglick, line 266: :thinking: though given https://github.com/jenkinsci/durable-task-plugin/blob/09f6e98d784a71f0c5038388ec9f20a722258ae4/src/main/java/org/jenkinsci/plugins/durabletask/AgentInfo.java#L25-L31 perhaps this should just be `os == OsType.LINUX`.

sunnyqeen: freebsd could also have nohup.
or would better add pre-step to check if nohup exists

PR #277, opened 2025-07-24
Flake in `OnceRetentionStrategyTest.deadConnectionsShouldReLaunched` after `DurableTaskStep` change

jglick:
Test introduced in #152 became unstable as of https://github.com/jenkinsci/workflow-durable-task-step-plugin/pull/463. Actually I am glad to be reminded of this old PR as it showed that until a year ago even an outbound cloud agent using the _recommended_ retention strategy could suffer from a permanent build hang as a result of a broken socket, suggesting that there are likely multiple flawed retention strategies floating around.


dwnusbaum (approved):

jglick:
Before: failed the first time on my laptop after bumping `workflow-durable-task-step` to `1442.vb_a_b_f5f3da_9f9` (and BOM to `4023.va_eeb_b_4e45f07` to match). After: passed 24× in a row.

PR #288, opened 2025-09-17
Dropping `Queue` lock from `OnceRetentionStrategy.done`

jglick:
`AbstractCloudSlave._terminate` is a potentially slow process (as suggested by the `IOException` / `InterruptedException` in its signature). For example `KubernetesSlave.deleteSlavePod` waits for a response from the API server, which might be overloaded and hanging. In the meantime the Jenkins queue lock is held (as of #2) which can cause everything else in the system to hang.

I tried to follow the justification in https://issues.jenkins.io/browse/JENKINS-26380 without success. The original issue was about a deadlock when `done` directly called `terminate`. That is addressed by backgrounding in `threadPoolForRemoting`. Then there was a lot of discussion about possible fixes in different places. The most relevant explanation seems to be

> builds being scheduled and then failing on a node that does not exist... all paths identified that the nodes were being removed **while** the queue maintenance was in progress, thus you should not remove nodes concurrent with the queue being maintained, instead you need to get the Queue's lock, confirm that the node is truly idle and then remove it all while holding the Queue's lock... and even that has issues... as the job can be scheduled but not started... so actually you mark the node as no longer accepting tasks, release the lock, sleep 100ms, re-acquire the lock and then double check it's idle and only then do you remove the node

But `OnceRetentionStrategy` already called `setAcceptingTasks(false)` which should be blocking other tasks from being scheduled onto this agent. (Inside `Executor.finish1`, called before `Computer.removeExecutor` in `finish2` which acquires the queue lock and which IIUC would mark the agent as idle.) There should be no need to check idleness of the agent—it is illegal to use this agent more than once (hence the name).

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/OnceRetentionStrategy.java
@@ -134,20 +134,18 @@ public final class OnceRetentionStrategy extends CloudRetentionStrategy implemen
         }
         LOGGER.log(Level.FINE, new Throwable(), () -> "terminating " + c + " idle? " + c.isIdle() + " idleStartMilliseconds=" + new Date(c.getIdleStartMilliseconds()));
         Computer.threadPoolForRemoting.submit(() -> {
-            Queue.withLock(() -> {
-                try {
-                    AbstractCloudSlave node = c.getNode();
-                    if (node != null) {
-                        LOGGER.fine(() -> "terminating " + c + " now");
-                        node.terminate();
-                    }
-                } catch (InterruptedException | IOException e) {
-                    LOGGER.log(Level.WARNING, "Failed to terminate " + c.getName(), e);
-                    synchronized (OnceRetentionStrategy.this) {
-                        terminating = false;
-                    }
+            try {
+                AbstractCloudSlave node = c.getNode();
+                if (node != null) {
+                    LOGGER.fine(() -> "terminating " + c + " now");
+                    node.terminate();
                 }
-            });
+            } catch (InterruptedException | IOException e) {
+                LOGGER.log(Level.WARNING, "Failed to terminate " + c.getName(), e);
+                synchronized (OnceRetentionStrategy.this) {
+                    terminating = false;
+                }
+            }
         });
     }
 


jglick:
> I don't find what is matching
> 
> > called before Computer.removeExecutor in finish2

https://github.com/jenkinsci/workflow-durable-task-step-plugin/blob/83386f485ce6730e36b7d8b0569fd837bbd2d083/src/main/java/org/jenkinsci/plugins/workflow/support/steps/ExecutorStepExecution.java#L995 → https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/jenkins/model/queue/AsynchronousExecution.java#L140 → https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/model/Executor.java#L531-L533 → https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/model/Executor.java#L508-L520 → https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/model/queue/WorkUnitContext.java#L177

> finish2 is potentially asynchronously though in calling `removeExecutor`

WDYM?

> There is no contract to state that `check()` must only be called whilst holding the queue lock as far as I am aware.

https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/slaves/CloudRetentionStrategy.java#L53-L54 https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/slaves/ComputerRetentionWork.java#L70-L78 https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/slaves/RetentionStrategy.java#L104 https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/slaves/SlaveComputer.java#L971-L974 so I think the (unusual) `check` → `done` case is covered: `setAcceptingTasks` will be called with the queue lock held.

> I'm not sure how this currently works if the Computer currently has a flyweight task executing when the heavyweight one finishes

Not sure that is even possible; at least in Pipeline usage, flyweight tasks are only assigned to `MasterComputer`.

jtnord:
> > finish2 is potentially asynchronously though in calling removeExecutor

> WDYM?

https://github.com/jenkinsci/jenkins/blob/5ad08302cdf80a66701ff48d02590e0dccf58c62/core/src/main/java/hudson/model/Computer.java#L1095-L1098

`finish2` removing the executor will be IIUC async as the lock is not held when Computer.removeExecurot() is called so the removal is punted to the standard computer threadpool.

jglick:
Well `tryWithLock` _may_ succeed. Either way, yes the `removeComputer` will happen sometime later. My point is that this should be fine: `setAcceptingTasks` is called inside `finish1`, so by the time `Computer.removeExecutor` (called somehow from `finish2`) runs, whenever that is, which I think is when the queue might consider scheduling a new task on it, this computer will already be refusing new tasks. And even if there were some flaw in that reasoning and a race condition exists whereby a new task could “sneak” onto the agent between the completion of its legitimate task and before `setAcceptingTasks` is called, it has been vulnerable since this class was created (before #2) anyway—I am only touching code run _after_ that call is made.

jtnord (approved):

jtnord:
> Well `tryWithLock` _may_ succeed. Either way, yes the `removeComputer` will happen sometime later. My point is that this should be fine: `setAcceptingTasks` is called inside `finish1`, so by the time `Computer.removeExecutor` (called somehow from `finish2`) runs, whenever that is, which I think is when the queue might consider scheduling a new task on it, this computer will already be refusing new tasks. And even if there were some flaw in that reasoning and a race condition exists whereby a new task could “sneak” onto the agent between the completion of its legitimate task and before `setAcceptingTasks` is called, it has been vulnerable since this class was created (before #2) anyway—I am only touching code run _after_ that call is made.

Which appears to be a valid shortcut as `OnceRetentionStrategy` is only applicable for Nodes with exactly 1 executor.

PR #536, opened 2026-01-12
Update test fixtures to Java 21

jglick:
Needed in order to test PCT `-Djenkins.version=2.545`.

PR #537, opened 2026-01-12
Update test fixtures to Java 25

jglick:
Follows up #536; why not get ahead of the game and avoid an emergency later? Agents running Java 25 should work fine on recent-ish controllers.


gbhat618 (approved):

jglick:
> Will there be a release for it?

No need I think; dropping Java 21 is a long way out, by which time there would surely have been releases of this plugin for other reasons.

PR #541, opened 2026-02-19
`binaryInactive` assertion failure on Windows in CI

jglick:
https://github.com/jenkinsci/durable-task-plugin/pull/540#issuecomment-3927869025 Improving failure messages from assertion introduced in #139.


jglick:
No, the test tries to run `tasklist` to ensure that a process is no longer running. I can try to make the test pass when this tool is not permitted, and we just will not get this assertion in CI. (It passes for me locally on Windows 10.)

PR #561, opened 2026-05-28
Improved `LAUNCH_DIAGNOSTICS` behavior

jglick:
While working on https://github.com/jenkinsci/kubernetes-plugin/pull/2837 I rediscovered https://github.com/jenkinsci/kubernetes-plugin/pull/871#issuecomment-710049155. Simplifying the launcher behavior when this flag is set, to actually check the wrapper status and log output normally. It is possible we could just make this behavior unconditional (only calling `quiet(true)` normally) but that would require more careful testing.

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -24,20 +24,17 @@
 
 package org.jenkinsci.plugins.durabletask;
 
-import edu.umd.cs.findbugs.annotations.NonNull;
 import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
 import hudson.EnvVars;
 import hudson.Extension;
 import hudson.FilePath;
 import hudson.Launcher;
-import hudson.PluginWrapper;
 import hudson.Proc;
 import hudson.Util;
 import hudson.model.TaskListener;
 import hudson.remoting.VirtualChannel;
 import hudson.tasks.Shell;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.File;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -56,7 +53,9 @@ import org.kohsuke.accmod.restrictions.NoExternalUse;
 import org.kohsuke.stapler.DataBoundConstructor;
 import edu.umd.cs.findbugs.annotations.CheckForNull;
 import edu.umd.cs.findbugs.annotations.Nullable;
+import hudson.AbortException;
 import jenkins.MasterToSlaveFileCallable;
+import jenkins.util.SystemProperties;
 
 /**
  * Runs a Bourne shell script on a Unix node using {@code nohup}.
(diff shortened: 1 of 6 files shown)


src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
```diff
@@ -324,7 +324,7 @@ private FilePath pidFile(FilePath ws) throws IOException, InterruptedException {
                 long currentTimestamp = getLogFile(workspace).lastModified();
                 if (currentTimestamp == 0) {
                     listener.getLogger().println("process apparently never started in " + controlDir);
-                    if (!LAUNCH_DIAGNOSTICS) {
```
jtnord, line 327: why was this change from the LAUNCH_DIAGNOSTIC field?  
the change means that termination diagnostics will always reflect the latest change of the property, yet the launching diagnostics does not and requires a restart (or setting via the script console).

jglick: This is part of another class, and I did not wish to introduce a new serialized `boolean` field. Since this clause should rarely be encountered, and _usually_ the property would be set at startup, this seemed close enough.

jglick:
I need to verify whether the `join` works with the binary wrapper. I hope it forks a clone and exits the main process, akin to the Bourne shell wrapper script, but I do not recall the details.

jtnord (approved):

jglick:
At least `BourneShellScriptUbuntuTest` (which uses the binary wrapper) passes locally with

```diff
diff --git src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
index 31dcf0b..c132916 100644
--- src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
+++ src/main/java/org/jenkinsci/plugins/durabletask/BourneShellScript.java
@@ -81,7 +81,7 @@ public final class BourneShellScript extends FileMonitoringTask {
      *
      * For the binary wrapper, this enables the debug flag.
      */
-    private final boolean LAUNCH_DIAGNOSTICS = SystemProperties.getBoolean(LAUNCH_DIAGNOSTICS_PROP);
+    private final boolean LAUNCH_DIAGNOSTICS = SystemProperties.getBoolean(LAUNCH_DIAGNOSTICS_PROP, true);
 
     /**
      * Seconds between heartbeat checks, where we check to see if
```

I also needed to

```diff
@@ -208,9 +208,6 @@ public final class BourneShellScript extends FileMonitoringTask {
         }
         // JENKINS-58290: launch in the background. No need to close stdout/err, binary does not write to them.
         cmd.add("-daemon");
-        if (LAUNCH_DIAGNOSTICS) {
-            cmd.add("-debug");
-        }
         return cmd;
     }
 
```

to avoid a failure in `watch` https://github.com/jenkinsci/durable-task-plugin/blob/88c2fa56953e5de731bc9f4f3fac665b2d3173e4/src/test/java/org/jenkinsci/plugins/durabletask/BourneShellScriptTest.java#L292

Issue #560, opened 2026-05-26
Improving CauseOfBlockage during queue maintenance with cache


bennettzhu1:
### What feature do you want to see added?

Hey, I'd love to see `ContinuedTask.CauseOfBlockage` cached to speed up queue maintenance. We have seen some slowness in terminating agents after their jobs had finished, and looking at some thread dumps, found out that the queue maintenance was very slow and hangs around `GetCauseOfBlockage`.
```
"AtmostOneTaskExecutor[Periodic Jenkins queue maintenance] [#222258]" #626708 [564800] daemon prio=5 os_prio=0 cpu=148915.38ms elapsed=152.48s tid=0x000073912c0196a0 nid=564800 runnable  [0x00007388bdbfd000]
   java.lang.Thread.State: RUNNABLE
	at org.jenkinsci.plugins.durabletask.executors.ContinuedTask$Scheduler.canTake(ContinuedTask.java:65)
	at hudson.model.Queue$JobOffer.getCauseOfBlockage(Queue.java:285)
	at hudson.model.Queue.maintain(Queue.java:1774)
	at hudson.model.Queue$1.call(Queue.java:339)
	at hudson.model.Queue$1.call(Queue.java:336)
	at jenkins.util.AtmostOneTaskExecutor.lambda$maybeRun$0(AtmostOneTaskExecutor.java:107)
	at jenkins.util.AtmostOneTaskExecutor$$Lambda/0x0000000080a73d48.call(Unknown Source)
	at jenkins.security.ImpersonatingExecutorService.lambda$wrap$1(ImpersonatingExecutorService.java:75)
	at jenkins.security.ImpersonatingExecutorService$$Lambda/0x0000000080a7a000.call(Unknown Source)
	at java.util.concurrent.FutureTask.run(java.base@21.0.10/Unknown Source)
	at hudson.remoting.AtmostOneThreadExecutor$Worker.run(AtmostOneThreadExecutor.java:121)
	at java.lang.Thread.runWith(java.base@21.0.10/Unknown Source)
	at java.lang.Thread.run(java.base@21.0.10/Unknown Source)
```
The queue maintenance outer loop iterates through each buildable, and for each buildable iterates through idle executors: https://github.com/jenkinsci/jenkins/blob/master/core/src/main/java/hudson/model/Queue.java#L1768

[...]

PR #563, opened 2026-05-29
Performance: Avoid allocation if log level > finer

sghill:
This PR stops allocating `Supplier` objects in a hot path if `Level.FINER` is not loggable.

I've been having some performance issues with Jenkins controllers as the queue size and incoming agent connections grow, similar to what's described in #560. 

I profiled with JFR's `settings=profile` configuration, and found 20% of samples over a few hours were in `LOGGER.finer(Supplier)` in this code path. I know sampling isn't perfect and there are more impactful changes described in 560, but I thought this may be a nice first contribution to this plugin - it is a hot path that is allocating a `Supplier` on every iteration of the loop. For anyone who has the root logger set to `INFO`, it'd be better to avoid that allocation since it won't be  logged anyway.

Screenshot 2026-05-28 at 20 30 53

### Testing done

`mvn clean verify` and [JMH Benchmarks](https://github.com/sghill/durable-task-plugin/blob/perf/benchmark/benchmarks/pom.xml#L217-L224) that include the `org.apache.logging.log4j:log4j-jul` bridge I'm using in production. I ran my benchmarks on a Macbook Pro M1 Max with 2 forks, 5 warmup iterations and 10 benchmark iterations.

It showed an improvement of 13% with a queue size of 2000:

Screenshot 2026-05-28 at 21 23 07

### Submitter checklist
- [x] Ensure you have provided tests that demonstrate the feature works or the issue is fixed

### DIFF
--- a/src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
+++ b/src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
@@ -30,6 +30,8 @@ import hudson.model.Node;
 import hudson.model.Queue;
 import hudson.model.queue.CauseOfBlockage;
 import hudson.model.queue.QueueTaskDispatcher;
+
+import java.util.logging.Level;
 import java.util.logging.Logger;
 import org.kohsuke.accmod.Restricted;
 import org.kohsuke.accmod.restrictions.NoExternalUse;
@@ -58,24 +60,34 @@ public interface ContinuedTask extends Queue.Task {
         }
 
         @Override public CauseOfBlockage canTake(Node node, Queue.BuildableItem item) {
+            // as this method can be a hostspot, do not use a Supplier in logging statements
+            // rather guard the logging statement inside an if
+            final boolean isFiner = LOGGER.isLoggable(Level.FINER);
             if (isContinued(item.task)) {
-                LOGGER.finer(() -> item.task + " is a continued task, so we are not blocking it");
+                if (isFiner) {
+                    LOGGER.finer(item.task + " is a continued task, so we are not blocking it");
+                }
                 return null;
             }
+            final boolean isFine = LOGGER.isLoggable(Level.FINE);
             for (Queue.BuildableItem other : Queue.getInstance().getBuildableItems()) {
                 if (isContinued(other.task)) {
                     Label label = other.task.getAssignedLabel();
                     if (label == null || label.matches(node)) { // conservative; might actually go to a different node
-                        LOGGER.fine(() -> "blocking " + item.task + " in favor of " + other.task);
+                        if (isFine) {
+                            LOGGER.fine("blocking " + item.task + " in favor of " + other.task);
+                        }
                         return new HoldOnPlease(other.task);
-                    } else {
-                        LOGGER.finer(() -> other.task + "’s label " + label + " does not match " + node);
+                    } else if (isFiner) {
+                        LOGGER.finer(other.task + "’s label " + label + " does not match " + node);
                     }
-                } else {
-                    LOGGER.finer(() -> other.task + " is not continued, so it would not block " + item.task);
+                } else if (isFiner) {
+                    LOGGER.finer(other.task + " is not continued, so it would not block " + item.task);
                 }
             }
-            LOGGER.finer(() -> "no reason to block " + item.task);
+            if (isFiner) {
+                LOGGER.finer("no reason to block " + item.task);
+            }
             return null;
         }
 


src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -58,8 +60,11 @@ private static boolean isContinued(Queue.Task task) {
         }
 
         @Override public CauseOfBlockage canTake(Node node, Queue.BuildableItem item) {
+            final boolean isFiner = LOGGER.isLoggable(Level.FINER);
             if (isContinued(item.task)) {
-                LOGGER.finer(() -> item.task + " is a continued task, so we are not blocking it");
+                if (isFiner) {
+                    LOGGER.finer(() -> item.task + " is a continued task, so we are not blocking it");
```
jtnord: The reason to use a supplier is that you delay performing string concatenation which reduces costly operations in the case the logger is not enabled.
Given we know this logger level is enabled, creating a supplier to perform string concatenation is a waste of CPU when you can just do String concatenation here.  (we don;t use any asynchronous logging, and if we did things would get funky as the items passed into many suppliers are not final)

e.g.  (all suggestions untested)

```suggestion
                    LOGGER.finer(item.task + " is a continued task, so we are not blocking it");
```

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -68,14 +73,16 @@ private static boolean isContinued(Queue.Task task) {
                     if (label == null || label.matches(node)) { // conservative; might actually go to a different node
                         LOGGER.fine(() -> "blocking " + item.task + " in favor of " + other.task);
```
jtnord: is this not also important?


```suggestion

                        if (isFiner) {
                            LOGGER.fine("blocking " + item.task + " in favor of " + other.task);
                        }
```

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -68,14 +73,16 @@ private static boolean isContinued(Queue.Task task) {
...
                         return new HoldOnPlease(other.task);
-                    } else {
+                    } else if (isFiner) {
                         LOGGER.finer(() -> other.task + "’s label " + label + " does not match " + node);
```
jtnord:
```suggestion
                        LOGGER.finer(other.task + "’s label " + label + " does not match " + node);
```

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -68,14 +73,16 @@ private static boolean isContinued(Queue.Task task) {
...
                     }
-                } else {
+                } else if (isFiner) {
                     LOGGER.finer(() -> other.task + " is not continued, so it would not block " + item.task);
```
jtnord:
```suggestion
                    LOGGER.finer(other.task + " is not continued, so it would not block " + item.task);
```

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -68,14 +73,16 @@ private static boolean isContinued(Queue.Task task) {
...
                 }
             }
-            LOGGER.finer(() -> "no reason to block " + item.task);
+            if (isFiner) {
+                LOGGER.finer(() -> "no reason to block " + item.task);
```
jtnord:
```suggestion
                LOGGER.finer("no reason to block " + item.task);
```

jtnord (commented):
The continued use of a supplier is dubious.  
Given the expensive thing is string concatination and that is what the supplier guards against, using a supplier when we know we will log (not throw away) the message just adds an unnecessary code path.

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
sghill: We would want another guard for `isFine` rather than `isFiner`. Happy to add this. 

In my case it's not important as none of my tasks are Continued

jtnord: good catch, I missed the difference in levels.

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
```diff
@@ -58,8 +60,11 @@ private static boolean isContinued(Queue.Task task) {
         }
 
         @Override public CauseOfBlockage canTake(Node node, Queue.BuildableItem item) {
+            final boolean isFiner = LOGGER.isLoggable(Level.FINER);
```
jtnord, line 65: for future maintainers.

```suggestion
            // as this method can be a hostspot, do not use a Supplier in logging statements
            // rather guard the logging statement inside an if
            final boolean isFiner = LOGGER.isLoggable(Level.FINER);
```

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
sghill: Will give this a test and report back, thanks for the suggestion

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
sghill: Added another variable for `isFine`

src/main/java/org/jenkinsci/plugins/durabletask/executors/ContinuedTask.java
sghill: Incorporated all the recommended changes, thanks for the review and feedback @jtnord!

jtnord (approved):

jtnord:
amends #281 which itself attempted to reduce overhead by moving from parameterized logging to a Supplier.
//cc @jglick

sghill:
Hi @jtnord and @jglick, could we get this merged and released?

I had another controller lock up today and point to heavy CPU usage in this method. Happy to help out with maintaining this plugin if wanted

jglick (approved):

jglick:
https://github.com/jenkinsci/durable-task-plugin/compare/master...sghill:durable-task-plugin:perf/benchmark FTR

